PR #8, opened 2015-04-15
Fix pep8 errors in ipynb_exporter

JaviMerino:
I ran first pep8 and then pylint, so I missed the pep8 errors introduced
by me shutting up pylint 😇


setrofim:
:) yup, that occasionally happens: 

https://github.com/ARM-software/workload-automation/commit/086f03730bf0438c04c49e481b38a0aa97070d01

PR #9, opened 2015-04-15
Add a notify result processor

JaviMerino:
This result processor displays a desktop notification when the run
finishes.  It's useful when you are running a long agenda in WA and want
to be notified when the results are available.


JaviMerino:
pynotify is not in pypi so I don't think it would work.  We could use notify2, which is in pypi and has a very similar unpythonic interface.  Would you prefer that?

setrofim:
Yes -- it should be possible to install all (Python) dependencies for WA with 

    sudo pip install wlauto[all] 

anything else makes installation too complicated.

PR #12, opened 2015-04-17
Fix trace-cmd after 2276ae0c5b54

JaviMerino:
Commit 2276ae0c5b54 ("Fixing config processing for extensions with
non-identifier names.") broke customizing the trace-cmd instrumentation
from the agenda.  With an agenda like:

config:
  instrumentation: [trace-cmd, delay]
  trace_events: ['thermal*']
  trace_buffer_size: 28000

trace_events and trace_buffer_size get added to the RunConfiguration's
_raw_config under the trace-cmd name, but then when it's looked up in
_finalize_config_list(), the dictionary is actually looked up using
identifier(extname), i.e. 'trace_cmd'.  Fix this by adding the user's
configuration using identifier(name) as well.

### DIFF
--- a/wlauto/core/configuration.py
+++ b/wlauto/core/configuration.py
@@ -639,7 +639,7 @@ class RunConfiguration(object):
         for param, ext in ga.iteritems():
             for name in [ext.name] + [a.name for a in ext.aliases]:
                 self._load_default_config_if_necessary(name)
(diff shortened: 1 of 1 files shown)


JaviMerino:
This fixes the issue *for* *me*, but I barely understand this code, so I could have broken other stuff.

PR #15, opened 2015-04-26
Add ebizzy workload

lisatn:
Add ebizzy to allow users to run a workload resembling common
web server application workloads.


lisatn:
The ebizzy workload should run on both Linux and Android.

PR #19, opened 2015-05-24
Cpustate

chagar01:
Hi Sergei

Can you please pull this new result processor it creates load parallelism and state residency data. It puts up with dropped traces and so should be fairly robust. Been tested on a few platforms, SMP and bL, and I've even put up with pylint/pep8 pendantry and fixed thereoff

Cheers

Charles


setrofim:
Implemented in commit 

https://github.com/ARM-software/workload-automation/commit/02af02f0cbf98ea2d54660187e788ae13c49438a

PR #20, opened 2015-05-24
Work around lacking sudo

chagar01:
Hi Sergei

You might wanna consider this one. Not sure it's the right thing to do, but may be worth considering. Not fussed if you drop it either.

Cheers

Charles


setrofim:
This is already resolved by 

https://github.com/ARM-software/workload-automation/commit/bf12d1845703098935849cb81a994b3f10100f4c

PR #25, opened 2015-06-04
get options in config_example.py reachable

zhizhouzh:
Some options in config_example.py are not reachable. It makes users
confused. So make the options in the file as gobal_alias.

Signed-off-by: zhizhou.zhang <zhizhou.zhang@spreadtrum.com>


setrofim:
Thanks a lot of this!

PR #31, opened 2015-06-28
Lmbench

bobbyb-arm:
Hi Sergei,

This is not a pull request - I plan to coalesce commits when the patches are ready.

However, wondered have cleaned up based on your initial comments, and run through pylint/pep8. 

Any comments appreciated.

-Bobby


setrofim:
Hi Bobby,

A few of minor points:

- The binary directory should be renamed to just "arm64" instead of
  "arm64-v8a", otherwise the binaries won't get picked up.
- The validate() method in this case is not needed. allowed_values will be
  automatically enforced.
- In the description, I would move the first sentence of the second paragraph
  up into the first paragraph, so something like

  "Run a subtest of lmbench,  a suite of simple, portable ANSI/C
   microbenchmarks for UNIX/POSIX"

- It would be good to have a short description of the currently supported tests
  ('lat_mem_rd' and 'bw_mem') in the workload description.
- Rather than including the entirety of GPLv2, Why not just link to it at the
  top of what is currently COPYING-2 file?

Otherwise, looks good.

Cheers,
Sergei

________________________________________

bobbyb-arm:
Closing as obsolete. pull/34 supersedes.

PR #34, opened 2015-06-29
Initial commit of lmbench workload

bobbyb-arm:
Hi Sergei,

Thanks for your review. If I've got it right, this is the squashed branch for merging.

I updated per your comments, except I think we should keep COPYING out of (perhaps conservative) convention.

-Bobby


setrofim:
Looks good. Thanks for contributing this workload!

PR #37, opened 2015-07-06
servo board instrument

freedomtan:
A simple implementation of using servo board as an instrument to measure big and LITTLE cluster power consumption. Tested on 8173 oak-rev1 with the energy_model instrument


derkling:
@setrofim what is the status of this PR?
This is a really cool feature to have possibly also (if not even before) in devlib ;-)

setrofim:
I guess, you could say the status is "on ice". IIRC, Vas ran into some issues while testing this, and he had an alternative implementation pending before he left.

I currently don't have the setup to test this. If you can try this out and confirm that it works without issues, I'd be happy to merge it.

setrofim:
Obsoleted by https://github.com/ARM-software/workload-automation/pull/144

PR #38, opened 2015-07-07
Add stream workload

lisatn:
Initial commit of stream workload.


wlauto/workloads/stream/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+
+        self.timeout = 50
+
+    def run(self, context):
+        if self.threads:
+            export_command = 'export OMP_NUM_THREADS={}'.format(self.threads)
+            self.device.execute(export_command)
+            self.output = self.device.execute(self.optional_command, timeout=self.timeout)
+        else:
+            self.output = self.device.execute(self.default_command, timeout=self.timeout)
```
setrofim: Rather than create two commands and select between them here, why not move the if statement to the setup and only create one command? Additionally, you shouldn't assume that the subsequent execute() calls will happen in the same shell (this not the case for Android), so you can't rely on export. So something lik

    if self.threads:
         self.command = 'OMP_NUM_THREADS={} {} > {}'.format(self.threads, self.stream_optional, self.results)
    else:
         self.command = '{} > {}'.format(self.stream_default, self.results)

in the setup(), followed by just

         self.output = self.device.execute(self.command, timeout=self.timeout)

wlauto/workloads/stream/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+        Parameter('threads', kind=int, default=0,
+                  description='The number of threads to execute if OpenMP is enabled')
+    ]
+
+    def setup(self, context):
+        self.stream_noomp_binary = context.resolver.get(Executable(self, self.device.abi, 'stream_noomp'))
+        self.stream_omp_binary = context.resolver.get(Executable(self, self.device.abi, 'stream_omp'))
+
+        self.stream_default = self.device.install(self.stream_noomp_binary)
+        self.stream_optional = self.device.install(self.stream_omp_binary)
```
setrofim: This is better done by setting class variables in initialize(). That way, binaries will only be installed once per run, rather than each time this workload is executed (see rt-app workload implementation for an example)

wlauto/workloads/stream/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+                        context.result.add_metric(label, float(match.group(1)), units)
+
+                for label, regex in regex_map2.iteritems():
+                    match = regex.search(line)
+                    if match:
+                        context.result.add_metric(label, float(match.group(1)), match.group(2))
+
+    def teardown(self, context):
+        self.device.uninstall_executable(self.stream_default)
+        self.device.uninstall_executable(self.stream_optional)
```
setrofim: if installation is moved to initialize(), as per the comment above, then thi should, correspondingly, be moved to finalize() (again, see rt-app workload for an example)

lisatn:
Thanks for the feedback, Sergei. Will make changes.

setrofim:
Thanks for contributing this workload.

PR #39, opened 2015-07-09
ipynb_exporter convert to html

JaviMerino:
Add a couple of options to the `ipynb_exporter` so that it can create static html versions of the notebook.


JaviMerino:
I've tested it with the latest ipython as well as with the one by default in Ubuntu.  pep8 and pylint are happy for the two files modified in these changes.

setrofim:
Looks good. Thanks for contributing this enhancement.

PR #43, opened 2015-08-10
Add iozone workload

lisatn:
Initial commit of the iozone workload. For now, the default timeout is set to 4 hours, which seems extreme. For some Linux devices, they take over 2 hours to run all 13 tests using the default iozone -a command, depending on number of CPUs, file stride size, processor cache size, and so forth. This workload has not been tested on Android devices yet.


wlauto/workloads/iozone/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+        self.results = os.path.join(self.device.working_directory,
+                                    iozone_results_txt)
+        self.command = self._build_command()
+
+        if self.threads and self.auto_mode:
+            raise ConfigError("You cannot set the number of threads and enable"
+                              " auto mode at the same time.")
+
+    def _build_command(self):
+        iozone_command = '{}'.format(self.device_binary)
```
setrofim: iozone creates a temporary file in in the current working directory, which may not be writable. This should therefore be:

     iozone_command = 'cd {} && {}'.format(self.device.working_directory, self.device_binary)

wlauto/workloads/iozone/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+
+        if self.auto_mode:
+            self.auto_option = ' -a'
+            iozone_command += self.auto_option
```
setrofim: Is there a reason to save `self.auto_option`?  could be just

     if self.auto_mode:
        iozone_command += ' -a'

The same goes for all the other options below.

wlauto/workloads/iozone/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+                if cache_size_match:
+                    context.result.add_metric("processor_cache_size",
+                                              float(cache_size_match.group(1)),
+                                              cache_size_match.group(2))
+
+                cache_line_match = cache_line_regex.search(line)
+                if cache_line_match:
+                    context.result.add_metric("processor_cache_line_size",
+                                              float(cache_line_match.group(1)),
+                                              cache_line_match.group(2))
```
setrofim: I may be misunderstanding what iozone reports, but these seem to be the parameters under which the test was run, rather than the results of the test (so if anything, they should be added as classifiers), and the actual metrics aren't being extracted?

lisatn: Yes, you may be correct. The values for time resolution, processor cache size, and processor cache line size are displayed every time a user runs an iozone command (with the possibility that the values may vary), so I was looking for consistent items to add to the results.csv file for each run. Otherwise I wouldn't know what to extract.

setrofim: The set of metrics does not need to be completely consistent for all runs -- it's entirely OK for a workload to report different metrics depending on the parameter values specified when configuring it.

The values that should be extracted are the things that you want to compare between different executions or use for further analysis, etc.  Why are you running this in the first place? What is it measuring? Why would I want to run this, what would it give me? etc...

Looking at the raw output produced, and without knowing how you use this, it seems that the entire table of reported results should be extracted. With the first two columns (kb and reclen) being additional classifiers--attributes giving context to metrics--and the remaining columns being the metrics (which would result in 1625 values reported for each iteration for a a full -a run).

lisatn: What you said made sense, and I would have gone along with your suggestion. Except I'm running into the problem where if users execute the iozone workload with the -t option enabled (to run a specific number of threads), then the results would be different:


    Iozone: Performance Test of File I/O
            Version $Revision: 3.430 $
        Compiled for 32 bit mode.
        Build: linux-arm

    Contributors:William Norcott, Don Capps, Isom Crawford, Kirby Collins
                 Al Slater, Scott Rhine, Mike Wisner, Ken Goss
                 Steve Landherr, Brad Smith, Mark Kelly, Dr. Alain CYR,
                 Randy Dunlap, Mark Montague, Dan Million, Gavin Brebner,
                 Jean-Marc Zucconi, Jeff Blomberg, Benny Halevy, Dave Boone,
                 Erik Habbinga, Kris Strecker, Walter Wong, Joshua Root,
                 Fabrice Bacchella, Zhenghua Xue, Qin Li, Darren Sawyer,
                 Vangel Bojaxhi, Ben England, Vikentsi Lapa.

    Run began: Fri Aug 14 04:07:55 2015

    Command line used: /usr/local/bin/iozone -t 2
    Output is in kBytes/sec
    Time Resolution = 0.000001 seconds.
    Processor cache size set to 1024 kBytes.
    Processor cache line size set to 32 bytes.
    File stride size set to 17 * record size.
    Throughput test with 2 processes
    Each process writes a 512 kByte file in 4 kByte records

    Children see throughput for  2 initial writers  =  213579.16 kB/sec
    Parent sees throughput for  2 initial writers   =   16403.29 kB/sec
    Min throughput per process          =   95499.95 kB/sec
    Max throughput per process          =  118079.20 kB/sec
    Avg throughput per process          =  106789.58 kB/sec
    Min xfer                    =     412.00 kB

    Children see throughput for  2 rewriters    =  398145.08 kB/sec
    Parent sees throughput for  2 rewriters     =   20088.98 kB/sec
    Min throughput per process          =  172895.72 kB/sec
    Max throughput per process          =  225249.36 kB/sec
    Avg throughput per process          =  199072.54 kB/sec

Thoughts?

setrofim: Then you should handle both types of output. The command line used is reported as part of the results, and it's trival to check for the presenece of "-t", diverging into a different sub-parser based on that.  I'd also recomment always running with "-R" option that generates "Excel" output which would be a lot easlier to parse in auto test case.

lisatn:
I submitted a couple more changes. Let me know if you can't see my latest commits.

wlauto/workloads/iozone/__init__.py
```diff
@@ -0,0 +1,228 @@
...
+    By default, iozone will run all tests in auto mode. To run
+    specific tests, they must be written in the form of:
+
+    ['0', '1', '5', '6']
+
+    The official website for iozone is at www.iozone.org.
+    """
+
+    parameters = [
+        Parameter('tests', kind=list_of_strs, default=[],
```
setrofim: - Why `list_of_strs`? Seems that `list_of_ints` would be more appropriate.
- Should set `allowed_values=range(13)` as only 0-12 indicate valid tests
- `default` can be omitted

wlauto/workloads/iozone/__init__.py
```diff
@@ -0,0 +1,228 @@
...
+                  description='List of performance tests to run.'),
```
setrofim: Please document what test numbers mean in the description.

wlauto/workloads/iozone/__init__.py
```diff
@@ -0,0 +1,228 @@
...
+        Parameter('auto_mode', kind=bool, default=True,
+                  description='Run tests in auto mode.'),
+        Parameter('timeout', kind=int, default=14400,
+                  description='Timeout for the workload.'),
+        Parameter('file_size', kind=int, default=0,
```
setrofim: `default` can be omitted

wlauto/workloads/iozone/__init__.py
```diff
@@ -0,0 +1,228 @@
...
+                  description='Fixed file size.'),
+        Parameter('record_length', kind=int, default=0,
```
setrofim: `default` can be omitted

wlauto/workloads/iozone/__init__.py
```diff
@@ -0,0 +1,228 @@
...
+                  description='Fixed record length.'),
+        Parameter('threads', kind=int, default=0,
```
setrofim: `default` can be omitted

setrofim:
The metrics generated by the workload do not make sense. In the results.csv for
a test run I see

	id,workload,iteration,metric,header,value,units
	1,iozone,1,reclen,"['Writer', 'report']",64,kb
	1,iozone,1,bytes,"['Writer', 'report']",81645,kb
	1,iozone,1,bytes,"['Writer', 'report']",87786,kb
	1,iozone,1,bytes,"['Writer', 'report']",91310,kb
	1,iozone,1,bytes,"['Writer', 'report']",91936,kb
	1,iozone,1,bytes,"['Writer', 'report']",92347,kb
	[ ... ]

- "reclen" should be extracted as a classifier, not reported as metric
- The rest of the metrics are all called "bytes" and their entries are
  identical apart from the actual values. It's not clear how they
  should be interpreted.
- The could "header" is confusing. It's not clear what "header" means
  in the context of an I/O test. The values appear to be serialized lists
  which would be difficult to process, and the word "report" appears to be
  redundant and confusing.
- untins should be KB/s, as iozone reports data transfer rates.

From the raw output reported by iozone, I would expect to see something like


	id,workload,iteration,metric,file_size,record_length,value,units
	1,iozone,1,write,64,4,81645,kb/s
	1,iozone,1,write,64,8,87786,kb/s
	1,iozone,1,write,64,16,91310,kb/s
	1,iozone,1,write,64,32,91936,kb/s
	1,iozone,1,write,64,64,92347,kb/s
	[ ... ]

metric can extracted from the header of a table; something along the lines of

    metric = header.rsplit(' ', 1)[0].lower().rstrip('r')

"record_length" and "file_size" are classifiers extracted  extracted from 
the first row of a table (column headers) and the first column of a table 
(row headers) respectively.

wlauto/workloads/iozone/__init__.py
lisatn: I wasn't aware of list_of_ints, but I can make that change easily. I'll take your suggestion on setting allowed values.

wlauto/workloads/iozone/__init__.py
lisatn: I'll remove the defaults.

lisatn:
I'll see about getting the results to look similar to what you expect.

Thanks for the feedback!

lisatn:
Gentle ping in case you haven't seen the latest change I made to the iozone workload code.

setrofim:
Thanks for making the updates. Apologies for the delay in merging this.

PR #45, opened 2015-08-14
Androbench Storage Benchmark Workload Addition

bathepawan:
I have added initial draft for addition of Androbench v4 Benchmark for storage performance.
It measures Sequential Read, Sequential Write, Random Read and Write Performance.
More details on http://www.androbench.org/wiki/AndroBench


wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+# limitations under the License.
+#
+
+
+import os
+import re
+import time
+
+from wlauto import AndroidBenchmark
+from uiautomator import Device
```
setrofim: This introduces a new dependency on uiautomator module. Please could you move uiautomator-related code into the corresponding Java file -- as is currenlty done for every other workload that depends on it.

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class Androbench(AndroidBenchmark):
+
+    name = 'androbench'
+    description = """Androbench measures the storage performance of device"""
+    package = 'com.andromeda.androbench2'
+    activity = '.main'
+    device=''
```
setrofim: "device" attribute already gets intialized to be the instance of the WA Device interface that is used to interract with a device. Please do not overrided it to be something else.

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def setup(self, context):
+		global device
+		os.system('adb devices > deviceinfo')
+		devinf = open('deviceinfo','rb')
+		dev = devinf.readlines()[1].split('\t')[0]
+		devinf.close()
+		device=Device(dev)
+		os.system('rm deviceinfo')
```
setrofim: None of the above is necessary -- a connection to the device gets automatically intialized by the base class; UIAutomator interactions should be implemented in Java.

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+			if device(text="Cancel").exists :
+				device(text="Cancel").click()
+				time.sleep(1)
+				break
+			
+
+    def update_result(self, context):
+        super(Androbench, self).update_result(context)
+        os.system('adb shell cp /data/data/com.andromeda.androbench2/databases/history.db /sdcard/results.db')
+        os.system('adb pull /sdcard/results.db .')
```
setrofim: Pelase use the WA Device interface (instantiated as `self.device`) to interract with the device. Do not invoke adb directly. Also locations on the host should generally be specifed relative to `context.output_directory`; do not create files in the current working directory. The above two lines should be relace with something like

    device_results = self.device.path.join(self.device.package_data_directory, self.package, 'databases', 'history.db')
    host_results = os.path.join(context.output_directory, 'results.db')
    self.device.pull_file(device_results, host_results, as_root=True)

(the `as_root=True` will take care of the intermediate transfer via the SD card.)

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+        os.system('sqlite3 results.db "select * from history" > results.raw')
+        fhresults=open("results.raw","rb")
+        results=fhresults.readlines()[0].split('|')
```
setrofim: Please use the built-in Python sqlite3 module:

https://docs.python.org/2/library/sqlite3.html

rather than assuming that `sqlite3` binary is installed on the host. This will also remove the need for results.raw

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+        context.result.add_metric('Sequential Read MB/s ', results[8])
```
setrofim: There is a separate field for units, they should not be merged with the metric name. The above line should be

    context.result.add_metric('Sequential Read', results[8], 'MB/s')

Similar for the there lines below.

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+        context.result.add_metric('Sequential Write MB/s ', results[9])
+        context.result.add_metric('Random Read MB/s ', results[10])
+        context.result.add_metric('Random Write MB/s ', results[12])
+        os.system('rm results.raw')
+        
+
```
setrofim: You can remove the two-stubbed out methods below.

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+        context.result.add_metric('Random Write MB/s ', results[12])
+        os.system('rm results.raw')
+        
+    
+
+    def teardown(self, context):
+        pass
+
+    def validate(self):
+        pass
```
setrofim: Please also run the `pylint` and `pep8` checkers (see http://pythonhosted.org/wlauto/contributing.html) and fix the issues they report.

bathepawan:
I have fixed below issues.
1. Use sqllite3 to remove host dependencies , results update correction.
2. Moved UiAutomator code from python to java, removed obsolete device interfacing code.
3. Fixed issue reported by pep8 and pylint.

Could you please check if it is good to go?

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+class Androbench(AndroidUiAutoBenchmark):
+    name = 'androbench'
+    description = """Androbench measures the storage performance of device"""
+    package = 'com.andromeda.androbench2'
+    activity = '.main'
+    run_timeout = 10 * 60
+
+    def update_result(self, context):
+        super(Androbench, self).update_result(context)
+        db = '/data/data/com.andromeda.androbench2/databases/history.db'
```
setrofim: above line should be

    db = self.device.path.join(self.device.package_data_directory, self.package, 'databases/history.db')

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+        qs = 'select * from history'
+        res = 'results.raw'
+        os.system('adb shell sqlite3 %s "%s" > %s' % (db, qs, res))
+        fhresults = open("results.raw", "rb")
+        results = fhresults.readlines()[0].split('|')
```
setrofim: As per comments in the perivous pull request, please do not invoke either adb or sqilte3 in the code. *all* interactions with the device must be performed via `self.device`; and you should not assume the presence of sqlite3 binary either on the host or on the target device. Also, do not create files in the current working directory; files on the host must be relative to `context.output_directory`,  `context.run_output_directory` or `tempfile.gettempdir()` (for temporary files not intended to be included in the results).

As outlined previously, you can get history.db from the device with

     host_results = os.path.join(context.output_directory, 'results.db')
     self.device.pull_file(db, host_results, as_root=True)

You can then use the built-in Python module (https://docs.python.org/2/library/sqlite3.html) to query the database to extract the metrics.

bathepawan:
Added  change replacing sqlite with python sqlite3 library. please review. thanks!

setrofim:
Thanks for making the changes. The code looks good now. However when I tried to run this on a rooted S6, it failed with an error complaining that `/data/data/com.andromeda.androbench2/databases/history.db` does not exist. Having checked on the device, `databases` subdirectory appears to be missing. Does something need to be done to enable it?

bathepawan:
May I know , Which version of androbench have you used? In mean time I will try to run in on some more devices. I recommend latest i.e androbench v4.

setrofim:
I've installed version 4.0.1 (the one currently in Play Store).

On 20 Aug 2015, 09:49:26 +0100, Pawan Bathe wrote:
> May I know , Which version of androbench have you used? In mean time I will try to run in on some more devices. I recommend latest i.e androbench v4.
>
> —
> Reply to this email directly or view it on GitHub<https://github.com/ARM-software/workload-automation/pull/45#issuecomment-132942402>.

-- IMPORTANT NOTICE: The contents of this email and any attachments are confidential and may also be privileged. If you are not the intended recipient, please notify the sender immediately and do not disclose the contents to any other person, use it for any purpose, or store or copy the information in any medium.  Thank you.

ARM Limited, Registered office 110 Fulbourn Road, Cambridge CB1 9NJ, Registered in England & Wales, Company No:  2557590
ARM Holdings plc, Registered office 110 Fulbourn Road, Cambridge CB1 9NJ, Registered in England & Wales, Company No:  2548782

bathepawan:
I have no access to Samsung S6 but I have tested it on couple of more devices , it is working fine for me.

bathepawan:
Looks like production device which don't give root access from outside of device shell don't allow copying from /data/data/ directly so file from /data/data/ has to be moved /sdcard/ before pulling, to make it run on production device with limited root access from device shell. What would be command for moving file from one location to another in device ? can you please  help me with that?

wlauto/workloads/androbench/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    package = 'com.andromeda.androbench2'
+    activity = '.main'
+    run_timeout = 10 * 60
+
+    def update_result(self, context):
+        super(Androbench, self).update_result(context)
+        dbn = 'databases/history.db'
+        db = self.device.path.join(self.device.package_data_directory, self.package, dbn)
+        host_results = os.path.join(context.output_directory, 'results.db')
+        self.device.pull_file(db, host_results)
```
setrofim: Should be
    self.device.pull_file(db, host_results, as_root=True)

setrofim:
`device.pull_file` will take care of that intermediate copy, if you specify `as_root=True` (see the above commend).

The issue is that databases/history does not exist on the device (I've adb'd into it as root and examined the data directory for androbench -- the databases subdirectory wasn't there).

bathepawan:
OK, understood, then Samsung S6 seems to have special case because I checked it on Samsung S5, file is getting generated there, but was getting error due to lack of root access I believe

bathepawan:
Using self.device.pull_file(db, host_results, as_root=True)  solved issue on S5.

setrofim:
I was able to run this successfully on another device, so I'm going to assume the issue I'm seeing on the S6 is peculiar that specific device. Pull request merged.

Thanks a lot for this contribution!

bathepawan:
Thanks, S6 I believe uses UFS 2.0 and not eMMC thats could be the reason for the benchmark to be incompatible with it.

PR #46, opened 2015-08-28
trace: fix data file trace name in warning

JaviMerino:
The warning refers to trace.bin, which is not the extension that we use
for it.  Instead, use the variable with the default trace
name (trace.dat) for the warning.


setrofim:
Thanks!

PR #50, opened 2015-11-02
Add Gem5Device to WA

Sticklyman1936:
Added the Gem5Device to WA.
As part of this, also enabled password-less telnet connections.

### DIFF
--- a/wlauto/devices/android/gem5/__init__.py
+++ b/wlauto/devices/android/gem5/__init__.py
@@ -566,21 +566,9 @@ class Gem5Device(AndroidDevice):
                 # it quietly (not as an error/warning) and move on.
                 self.logger.debug('Could not pull property file "{}"'.format(propfile))
 
-        # This is duplicated from the AndroidDevice class, as we want to run
-        # this code without executing the BaseLinuxDevice implementation
-        props = {}
-        props['android_id'] = self.get_android_id()
-        buildprop_file = os.path.join(context.host_working_directory, 'build.prop')
-        if not os.path.isfile(buildprop_file):
-            self.pull_file('/system/build.prop', context.host_working_directory)
-        self._update_build_properties(buildprop_file, props)
-        context.add_run_artifact('build_properties', buildprop_file, 'export')
-
-        dumpsys_window_file = os.path.join(context.host_working_directory, 'window.dumpsys')
-        self.execute('{} > {}'.format('dumpsys window', 'window.dumpsys'))
-        self.pull_file('window.dumpsys', dumpsys_window_file)
-        context.add_run_artifact('dumpsys_window', dumpsys_window_file, 'meta')
-        return props
+        # As we do not call the super class, we explicitly need to
+        # process the Android properties.
+        props = self._get_android_properties(context)
 
     def get_directory(self, context, directory):
         """ Pull a directory from the device """
--- a/wlauto/devices/android/gem5/__init__.py
+++ b/wlauto/devices/android/gem5/__init__.py
@@ -28,7 +28,7 @@ from pexpect import EOF, TIMEOUT
 
 from wlauto import AndroidDevice, settings, Parameter
 from wlauto.core import signal as sig
-from wlauto.utils import ssh
+from wlauto.utils import ssh, types
 from wlauto.exceptions import DeviceError, ConfigError
 
 
@@ -84,14 +84,9 @@ class Gem5Device(AndroidDevice):
     """
 
     name = 'gem5_android'
-    default_working_directory = '/data/'
     has_gpu = False
     platform = 'android'
     path_module = 'posixpath'
-    default_package_data_directory = '/data/data/'
-    default_binaries_directory = '/system/bin'
-    default_external_storage_directory = '/data/data/'
-    default_adb_name = None
 
     # gem5 can be very slow. Hence, we use some very long timeouts!
     delay = 3600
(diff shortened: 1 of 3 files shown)


wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+
+    name = 'gem5_android'
+    default_working_directory = '/data/'
+    has_gpu = False
+    platform = 'android'
+    path_module = 'posixpath'
+    default_package_data_directory = '/data/data/'
+    default_binaries_directory = '/system/bin'
+    default_external_storage_directory = '/data/data/'
+    default_adb_name = None
```
setrofim: The above `default_`'s should not be necessary as they will be handled by the super

wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+    def __init__(self, **kwargs):
+        self.logger = logging.getLogger('gem5Device')
+
+        super(Gem5Device, self).__init__(**kwargs)
+        self.adb_name = kwargs.get('adb_name') or self.default_adb_name
+        self.binaries_directory = kwargs.get('binaries_directory', self.default_binaries_directory)
+        self.package_data_directory = kwargs.get('package_data_directory', self.default_package_data_directory)
+        self.external_storage_directory = kwargs.get('external_storage_directory',
+                                                     self.default_external_storage_directory)
+        self.logcat_poll_period = kwargs.get('logcat_poll_period')
```
setrofim: the above `kwargs` processing should not be necessary as it will be handled by super's `__init__`

wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+                  "boot."),
+        Parameter('run_delay', kind=int, default=0, mandatory=False,
+                  override=True, description="This sets the time that the "
+                  "system should sleep in the simulated system prior to "
+                  "running and workloads or taking checkpoints. This allows "
+                  "the system to quieten down prior to running the workloads. "
+                  "When this is combined with the checkpoint_post_boot"
+                  " option, it allows the checkpoint to be created post-sleep,"
+                  " and therefore the set of workloads resuming from this "
+                  "checkpoint will not be required to sleep.")
```
setrofim: `override` should only be set for `core_names` and `core_clusters` as they are overriding the parameters with the same names in the super. The other parameters aren't represent in the super, so for them, `override` should not be set.

wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+        self._available_governor_tunables = {}
+        self._number_of_cores = None
+        self._logcat_poller = None
+        self.gem5outdir = os.path.join(settings.output_directory, "gem5")
+        self.stdout_file = None
+        self.stderr_file = None
+        self.stderr_filename = None
+        self.checkpoint = False
+        self.sckt = None
+
```
setrofim: As above, the `kwargs` procesing is not necessary as it will be handled by super's `__init__`.

wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+    def push_file(self, source, dest, as_root=False):  # pylint: disable=W0221
+        """
+        Push a file to the gem5 device using VirtIO
+
+        The file to push to the device is copied to the temporary directory on
+        the host, before being copied within the simulation to the destination.
+        Checks, in the form of 'ls' with error code checking, are performed to
+        ensure that the file is copied to the destination.
+        """
+        filename = source.split('/')[-1]
```
setrofim: filename = os.path.basename(source)

wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+        try:
+            if int(output):
+                return True
+            else:
+                return False
+        except ValueError:
+            if output:
+                return True
+            else:
+                return False
```
setrofim: Not sure I understand the above logic. If `output` is not an integer, then as long as it is not empty, assume the file exists?

wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+                else:
+                    continue
+            except DeviceError:
+                # We pull these files "opportunistically", so if a pull fails
+                # (e.g. we don't have permissions to read the file), just note
+                # it quietly (not as an error/warning) and move on.
+                self.logger.debug('Could not pull property file "{}"'.format(propfile))
+
+        # This is duplicated from the AndroidDevice class, as we want to run
+        # this code without executing the BaseLinuxDevice implementation
```
setrofim: Rather than duplicating this code, I would factor it out into an internal `_get_android_properties` inside `AndroidDevice` and the invoke that here.

wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+        super(Gem5Device, self).capture_screen(filepath)
+
+    # gem5 might be slow. Hence, we need to make the ping timeout very long.
+    def ping(self):
+        self.logger.debug("Pinging gem5 to see if it is still alive")
+        self.gem5_shell('ls /', timeout=self.longdelay)
+
+    # Additional Android-specific methods.
+    def forward_port(self, from_port, to_port):
+        raise DeviceError('we do not need forwarding')
```
setrofim: This should just be a `self.logger.warning`

wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+        command_index = output.find(command)
+        if command_index != -1:
+            output = output[command_index + len(command):].strip()
+
+        self.logger.debug("gem5_shell output: {}".format(output))
+
+        # We get a second prompt. Hence, we need to eat one to make sure that we
+        # stay in sync. If we do not do this, we risk getting out of sync for
+        # slower simulations.
+        self.sckt.expect(r'\[PEXPECT\]\$', timeout=10000)
```
setrofim: self.sckt.expect(self.sckt.UNIQUE_PROMPT, timeout=10000)

wlauto/devices/android/gem5/__init__.py
```diff
@@ -0,0 +1,820 @@
...
+
+        Write some unique text to the gem5 device to allow us to synchronise
+        with the shell output. We actually get two prompts so we need to match
+        both of these.
+        """
+        self.sckt.send("echo \*\*sync\*\*\n")
+        self.sckt.expect(r"\*\*sync\*\*", timeout=self.delay)
+        self.sckt.expect(r'\[PEXPECT\]\$', timeout=self.delay)
+        self.sckt.expect(r'\[PEXPECT\]\$', timeout=self.delay)
+
```
setrofim: Use `self.sckt.UNIQUE_PROMPT` instead of hard-coded string, as above

wlauto/devices/android/gem5/__init__.py
Sticklyman1936: So, this is actually slightly modified from the original - I should update the comment. As we use the same terminal as is used for dmesg etc. for a gem5 system, I have changed the code to write  the output from 'dumpsys window' to a file and pull the file, as opposed to relying on the output from the shell command itself.

The way I see it, there are two choices. We keep the close-to duplicated code here for the gem5 device, or we do what you suggest above and move it to an internal method, but change it to redirect to a file and then pull the file. The latter would allow us to avoid code duplication, but does change how we process Android properties for all devices. Which do you prefer?

setrofim: I think it still makes sense to go with one internal method. get_properites writes to a file anyway, so may as well do it on the device and pull the file, if that works for all platforms.

PR #51, opened 2015-11-05
Added ability to merge DAQ channels

ep1cman:
If the parameter merge_channels is set to true (false by default).
 DAQ channels that have consecutive letters on the end of their names
 will be summed together. E.g: A7, A15a, A15b becomes A7, A15

### DIFF
--- a/wlauto/utils/ssh.py
+++ b/wlauto/utils/ssh.py
@@ -58,8 +58,8 @@ class TelnetConnection(pxssh.pxssh):
     # pylint: disable=arguments-differ
 
     def login(self, server, username, password='', original_prompt=r'[#$]', login_timeout=10,
-              auto_prompt_reset=True, sync_multiplier=1):
-        cmd = 'telnet -l {} {}'.format(username, server)
+              auto_prompt_reset=True, sync_multiplier=1, port=23):
+        cmd = 'telnet -l {} {} {}'.format(username, server, port)
 
         spawn._spawn(self, cmd)  # pylint: disable=protected-access
         i = self.expect('(?i)(?:password)', timeout=login_timeout)


wlauto/instrumentation/daq/__init__.py
```diff
@@ -158,6 +159,11 @@ class Daq(Instrument):
                   to facillitate syncing kernel trace events to DAQ power
                   trace.
                   """),
+        Parameter('merge_channels', kind=bool, default=False,
+                  description="""
+                  If set to true, channels with consecutive letter suffixes will be summed.
```
setrofim: If set to ``True``, channels [...]

wlauto/instrumentation/daq/__init__.py
```diff
@@ -158,6 +159,11 @@ class Daq(Instrument):
...
+                  e.g. If you have channels A7a, A7b, A7c, A15a, A15b they will be summed to A7, A15
```
setrofim: Rather than doing it this way, and constraining the naming of individual channels, why not specify the mapping explicitly, e.g.

    merged_channels:
                 a53: [a53dvfs, a53dram]
                 a72: [a72dvfs, a72dram]

wlauto/instrumentation/daq/__init__.py
```diff
@@ -315,6 +325,32 @@ def _execute_command(self, command, **kwargs):
...
+                    path = os.path.join(output_directory, name+suffix+".csv")
+                    with open(path) as fh:
+                        reader = csv.reader(fh)
+                        metrics = reader.next()
+                        rows = _get_rows(reader, None, self.negative_samples)
+                        if summed:
+                            summed = [[x+y for x, y in zip(a, b)] for a, b in zip(rows, summed)]
+                        else:
+                            summed = rows
+                    os.remove(path)
```
setrofim: Do not remove the original csv

PR #52, opened 2015-11-09
Added freqsweep instrument

ep1cman:
Added an instrument to sweep a workload across all available frequenices
for both big and little cores


Sticklyman1936:
I think it would be useful if this could be extended/generalised to make it work on non-big.LITTLE systems. There are plenty of systems which are non-big.LITTLE, but it would still be useful to get a workload vs CPU frequency profile.

Does it also make sense to enable runs on more than one core?

ep1cman:
Would an agenda interface like this work?

freqsweep:
  sweeps:
    - 0-3
    - 0,1,4
    - 4

Where each list entry is a sweep and numbers correspond to what cores will be active during that sweep

PR #56, opened 2015-11-16
Streamline improvements

ep1cman:


wlauto/instrumentation/streamline/__init__.py
```diff
@@ -156,26 +139,22 @@ def __init__(self, device, **kwargs):
 
         self._check_has_valid_display()
 
-    def on_run_start(self, context):
-        if subprocess.call('which caiman', stdout=subprocess.PIPE, shell=True):
-            raise InstrumentError('caiman not in PATH. Cannot enable Streamline tracing.')
-        p = subprocess.Popen('caiman --version 2>&1', stdout=subprocess.PIPE, shell=True)
+    def validate(self):
+        if subprocess.call('which streamline', stdout=subprocess.PIPE, shell=True):
```
setrofim: should be

    if not which('streamline'):

`which` can be imported from `wlauto.utils.misc`

PR #62, opened 2015-12-01
trace-cmd: Added ability to generate reports on target device

ep1cman:


wlauto/instrumentation/trace_cmd/__init__.py
```diff
@@ -146,6 +141,15 @@ class TraceCmdInstrument(Instrument):
                   not already a trace-cmd on the device, an error is raised.
 
                   """),
+        Parameter('report_on_target', kind=boolean, default=False,
+                  description="""
+                  When enabled generation of reports will be done host-side because the generated file is
+                  very large. If trace-cmd is not available on the host device this setting and be disabled
+                  and the report will be generated on the target device.
+
+                .. note:: This requires the latest version of trace-cmd to be installed on the host (the
```
setrofim: This looks misaligned. Make sure the html doc's build OK.

wlauto/instrumentation/trace_cmd/__init__.py
```diff
@@ -237,35 +241,49 @@ def update_result(self, context):  # NOQA pylint: disable=R0912
...
+                        if process.returncode:
+                            raise InstrumentError('trace-cmd returned non-zero exit code {}'.format(process.returncode))
+                        if error:
+                            # logged at debug level, as trace-cmd always outputs some
+                            # errors that seem benign.
+                            self.logger.debug(error)
+                    except OSError:
+                        raise InstrumentError('Could not find trace-cmd. Please make sure it is installed and is in PATH.')
+
+            else:
```
setrofim: For readability:

    else:  # not self.report_on_target

wlauto/instrumentation/trace_cmd/__init__.py
```diff
@@ -237,35 +241,49 @@ def update_result(self, context):  # NOQA pylint: disable=R0912
...
+                try:
+                    trace_file = self.output_file
+                    txt_trace_file = os.path.join(self.device.working_directory, OUTPUT_TEXT_FILE)
+                    command = 'trace-cmd report {} > {}'.format(trace_file, txt_trace_file)
+                    self.device.execute(command)
+                    self.device.pull_file(txt_trace_file, context.output_directory, timeout=pull_timeout)
+                except DeviceError:
+                    raise InstrumentError('trace-cmd returned non-zero exit code.')
```
setrofim: Error message should be updated to something like "Could not generate TXT report on target."

wlauto/instrumentation/trace_cmd/__init__.py
setrofim: Actually thinking about it, that whole construct is kinda large and unwieldly; might be worth factoring parts of it out into helper methods; something like


    if self.report_on_target:
        self._generate_report_on_target()
    else: 
        self._generate_report_on_host()

PR #70, opened 2016-01-08
Antutu6

ep1cman:


wlauto/common/android/workload.py
```diff
@@ -236,6 +236,26 @@ def reset(self, context):  # pylint: disable=W0613
         self.device.execute('am force-stop {}'.format(self.package))
         self.device.execute('pm clear {}'.format(self.package))
 
+        # As of android API level 23, apps can request permissions at runtime,
+        # this will grant all of them so requests do not pop up when running the app
+        if self.device.get_sdk_version() >= 23:
```
setrofim: I'd move the stuff below out into a helper method (`self._grant_requested_permissions()` or something)

PR #77, opened 2016-01-14
workloads: Add apklaunch workload

bjackman:
This is a workload to install and run an arbitrary .apk

### DIFF
--- a/wlauto/workloads/apklaunch/__init__.py
+++ b/wlauto/workloads/apklaunch/__init__.py
@@ -1,0 +1,63 @@
+#    Copyright 2015 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+
+from time import sleep
+
+from wlauto import Workload, Parameter
+from wlauto import File
+from wlauto.exceptions import ConfigError
+from wlauto.utils.android import ApkInfo
+
+
+class ApkLaunchWorkload(Workload):
+    name = 'apklaunch'
+    description = '''
+    Installs and runs a .apk file, waits wait_time_seconds, and tests if the app
+    has started successfully.
+    '''
+    supported_platforms = ['android']
+
+    parameters = [
+        Parameter('apk_file', description='Name to the .apk to run', mandatory=True),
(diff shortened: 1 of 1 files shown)


wlauto/workloads/apklaunch/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    name = 'apklaunch'
+    description = '''
+    Installs and runs a .apk file, waits wait_time_seconds, and tests if the app
+    has started successfully.
+    '''
+    supported_platforms = ['android']
+
+    parameters = [
+        Parameter('apk_file', description='Name to the .apk to run'),
+        Parameter('pkg_name', mandatory=True,
```
setrofim: I would change this to `package_name` or just `package`, for consistency with similar parameter namings elsewhere.

Also, this parameter isn't strictly necessary. You can obtain the package name from the APK with

    from wlauto.utils.android import ApkInfo
    self.package ApkInfo(self.apk_file).package

wlauto/workloads/apklaunch/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    def setup(self, context):
+        if self.install_required:
+            apk_file = context.resolver.get(File(self, self.apk_file))
+            self.logger.info("Installing {}".format(apk_file))
+            return self.device.install(apk_file)
+
+    def run(self, context):
+        self.logger.info("Starting {}".format(self.pkg_name))
+        self.device.execute('monkey -p {} -c android.intent.category.LAUNCHER 1'
+                            .format(self.pkg_name))
```
setrofim: Why use `monkey` rather than `am start` here? the `-W` flag would remove the need to sleep.

wlauto/workloads/apklaunch/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+        self.logger.info("Waiting {} seconds".format(self.wait_time_seconds))
+        sleep(self.wait_time_seconds)
+
+    def update_result(self, context):
+        # Check if the app is still running by grepping the output of "ps"
+        ps_lines = self.device.execute('ps').splitlines()
+        app_entries = [l for l in ps_lines if self.pkg_name in l.split()]
+        app_is_running = app_entries != []
```
setrofim: above can be replaced with

app_is_running = bool([p for p in self.device.ps() if p.name == self.pkg_name])

wlauto/workloads/apklaunch/__init__.py
bjackman: Ah, great; I didn't spot that in the API. Will remove the parameter.

wlauto/workloads/apklaunch/__init__.py
bjackman: Great. Also didn't spot ps().

wlauto/workloads/apklaunch/__init__.py
bjackman: Makes sense, this is just my Android noobiness showing. I'll keep the wait_time_seconds paramter, though, as apps could always crash after startup appears complete.

PR #78, opened 2016-01-14
BaseLinuxDevice: Tidied up the way binaries are handled

ep1cman:
Added:
get_binary_path: Checks binary_directory for the wanted binary, if
                 if its not there, it will use which to find a
                 system one. returns the full path

install_if_needed: will install a binary only if it is not present.

Changes:
 - Busybox is now deployed to non-rooted devices
 - is_installed has now been removed as the new functions supersede it
 - binaries will now always be installed to `binaries_directory` and
   not system folders.
 - updated workloads to use these new functions
   - rt-app and sysbench might still need work


wlauto/common/android/device.py
```diff
@@ -376,12 +389,11 @@ def uninstall(self, package):
 
     def uninstall_executable(self, executable_name):
         """
-        Requires root access.
 
         Added in version 2.1.3.
 
         """
-        on_device_executable = self.path.join(self.binaries_directory, executable_name)
+        on_device_executable = self.get_binary_path(executable_name, search_system_binaries=False)
```
setrofim: `on_device_executable` may now be `None` this should be detected and a meaningful error return (other wise, this would result in something like "could not uninstall None").

wlauto/common/linux/device.py
```diff
@@ -183,11 +185,14 @@ def validate(self):
...
-            else:
-                self.busybox = 'busybox'
+        if not self.binaries_directory:
+            self._set_binaries_dir()
+        self.execute('mkdir -p {}'.format(self.binaries_directory))
+        self.busybox = self.deploy_busybox(context)
+
+    def _set_binaries_dir(self):
+        # pylint: disable=attribute-defined-outside-init
+        self.binaries_directory = self.path.join(self.default_working_directory, "bin")
```
setrofim: This should be `working_directory` rather than `default_working_directory`.

wlauto/instrumentation/perf/__init__.py
```diff
@@ -91,11 +91,8 @@ class PerfInstrument(Instrument):
     ]
 
     def on_run_init(self, context):
-        if not self.device.is_installed('perf') or self.force_install:
-            binary = context.resolver.get(Executable(self, self.device.abi, 'perf'))
-            self.binary = self.device.install(binary)
-        else:
-            self.binary = 'perf'
+        binary = context.resolver.get(Executable(self, self.device.abi, 'perf'))
+        self.binary = self.device.install_if_needed(binary)
```
setrofim: I don't think this logic is equivalent to what has been removed -- if `self.force_install` is set, then the binary should be re-installed regardless of whether it is needed.

wlauto/instrumentation/trace_cmd/__init__.py
```diff
@@ -164,11 +164,12 @@ def on_run_init(self, context):
             raise InstrumentError('trace-cmd instrument cannot be used on an unrooted device.')
         if not self.no_install:
             host_file = context.resolver.get(Executable(self, self.device.abi, 'trace-cmd'))
-            self.trace_cmd = self.device.install_executable(host_file)
+            self.trace_cmd = self.device.install_if_needed(host_file)
```
setrofim: This should be just `install` rather than `install_if_needed` to be consistent with existing behavior.

wlauto/instrumentation/trace_cmd/__init__.py
```diff
@@ -164,11 +164,12 @@ def on_run_init(self, context):
...
         else:
-            if not self.device.is_installed('trace-cmd'):
+            self.trace_cmd = self.device.get_binary_path("trace-cmdS")
```
setrofim: typo: trailing `S` in `"trace-cmdS"`

wlauto/workloads/cyclictest/__init__.py
```diff
@@ -79,11 +79,8 @@ def setup(self, context):
         if not self.device.is_rooted:
             raise WorkloadError("This workload requires a device with root premissions to run")
 
-        if not self.device.is_installed('cyclictest'):
-            host_binary = context.resolver.get(Executable(self, self.device.abi, 'cyclictest'))
-            self.device_binary = self.device.install(host_binary)
-        else:
-            self.device_binary = 'cyclictest'
+        host_binary = context.resolver.get(Executable(self, self.device.abi, 'cyclictest'))
+        self.device_binary = self.device.install_if_needed(host_binary)
```
setrofim: Again should be just ``device.install` to be compatible with existing behavior

wlauto/workloads/hackbench/__init__.py
```diff
@@ -61,11 +61,8 @@ def setup(self, context):
         self.run_timeout = self.duration + timeout_buf
 
         self.binary_name = 'hackbench'
-        if not self.device.is_installed(self.binary_name):
-            host_binary = context.resolver.get(Executable(self, self.device.abi, self.binary_name))
-            self.device_binary = self.device.install(host_binary)
-        else:
-            self.device_binary = self.binary_name
+        host_binary = context.resolver.get(Executable(self, self.device.abi, self.binary_name))
+        self.device_binary = self.device.install_if_needed(host_binary)
```
setrofim: Should be `device.install`

PR #79, opened 2016-01-15
applaunch: Added support for IO scheduler test

chase-qi:
When IO is heavily loaded, the impact of IO schedulers on App launch
time varies. To measure the impact, added io_stress and io_scheduler two
parameters and related jinja2 blocks.

### DIFF
--- a/wlauto/workloads/applaunch/__init__.py
+++ b/wlauto/workloads/applaunch/__init__.py
@@ -77,6 +77,10 @@ class ApplaunchWorkload(Workload):
                   .. note:: This depends on appropriate sensors to be exposed through HWMON.
 
                   """),
+        Parameter('io_stress', kind=boolean, default=False,
+                  description='Specifies whether to stress IO during App launch.'),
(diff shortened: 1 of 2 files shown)


wlauto/workloads/applaunch/__init__.py
```diff
@@ -117,11 +123,20 @@ def setup(self, context):
         self.device.clear_logcat()
 
     def run(self, context):
-        self.device.execute('sh {}'.format(self.device_script_file), timeout=300)
+        self.device.execute('sh {}'.format(self.device_script_file), timeout=300, as_root=True)
```
setrofim: This change limits this test to only rooted devices. Since root is only required for io stress, please change to

     self.device.execute('sh {}'.format(self.device_script_file), timeout=300, as_root=self.io_stress)

chase-qi: Thanks for your review. That is a brilliant idea. I have updated the line.

PR #84, opened 2016-01-25
A couple of tweaks

bjackman:


setrofim:
Thanks for those! Didn't realize there were so may possessive "it's" in the docs...

PR #90, opened 2016-02-01
manual: Fixed trying to logcat on non linux devices

ep1cman:


wlauto/workloads/manual/__init__.py
```diff
@@ -103,3 +103,10 @@ def validate(self):
         if not self.user_triggered and not self.duration:
             raise ConfigError('Either user_triggered must be ``True`` or duration must be > 0.')
 
+        if self.enable_logcat is None:
+            if self.device.platform == "android":
+                self.enable_logcat = True
+            else:
+                self.enable_logcat = False
```
setrofim: the inner `if` can be replaced with

    self.enable_logcat = self.device.platform == "android"

Issue #97, opened 2016-02-09
wa fails to start when iPython version not recognised


stevebannister:
The following log shows what happens when running wa with Jupyter 4.1.0rc2 installed. Even without using the ipython exporter wa fails to start up. 

Is this expected behaviour? 

------- 
$ wa
CRITICAL:root:  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/wlauto/core/entry_point.py", line 54, in main
    load_commands(parser.add_subparsers(dest='command'))  # each command will add its own subparser
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/wlauto/core/entry_point.py", line 39, in load_commands
    ext_loader = ExtensionLoader(paths=settings.extension_paths)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/wlauto/core/extension_loader.py", line 162, in __init__
    self._load_from_packages(self.packages)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/wlauto/core/extension_loader.py", line 304, in _load_from_packages
    for module in walk_modules(package):
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/wlauto/utils/misc.py", line 138, in walk_modules
    mods.extend(walk_modules(submod_path))
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/wlauto/utils/misc.py", line 133, in walk_modules
    root_mod = __import__(path, {}, {}, [''])
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/wlauto/result_processors/ipynb_exporter/__init__.py", line 30, in <module>
    import wlauto.utils.ipython as ipython
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/wlauto/utils/ipython.py", line 39, in <module>
    if StrictVersion('5.0.0') > StrictVersion(IPython.__version__) >= StrictVersion('4.0.0'):
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/version.py", line 40, in __init__
    self.parse(vstring)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/version.py", line 107, in parse
[...]

PR #98, opened 2016-02-10
ipython: switched to using LooseVersion for version checks.

setrofim:
This is a fix for

https://github.com/ARM-software/workload-automation/issues/97

IPython can use rc tags in its version strings, which StrictVersion
can't handle.


stevebannister:
Perfect - works for me thanks

PR #103, opened 2016-02-15
ApkWorkload: Fixed runtime permission granting

ep1cman:
"Normal" android permissions are automatically granted and cannot
be changed. Trying to "pm grant" these caused an error, this should
no longer occur.

### DIFF
--- a/wlauto/common/android/workload.py
+++ b/wlauto/common/android/workload.py
@@ -22,7 +22,7 @@ from wlauto.core.workload import Workload
 from wlauto.core.resource import NO_ONE
 from wlauto.common.resources import ExtensionAsset, Executable
 from wlauto.exceptions import WorkloadError, ResourceError, ConfigError
-from wlauto.utils.android import ApkInfo
+from wlauto.utils.android import ApkInfo, ANDROID_NORMAL_PERMISSIONS
 from wlauto.utils.types import boolean
 import wlauto.common.android.resources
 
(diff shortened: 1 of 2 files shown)


wlauto/common/android/workload.py
```diff
@@ -29,6 +29,42 @@
 
 DELAY = 5
 
+ANDROID_NORMAL_PERMISSIONS = [
```
setrofim: This should be moved to `wlauto/utils/android.py` and imported from there

wlauto/common/android/workload.py
```diff
@@ -267,7 +303,12 @@ def _grant_requested_permissions(self):
                 break
 
         for permission in permissions:
-            self.device.execute("pm grant {} {}".format(self.package, permission))
+            # "Normal" Permisions are automatically granted and cannot be changed
+            for normal in ANDROID_NORMAL_PERMISSIONS:
+                if normal in permission:
```
setrofim: `if permission.endswith(nomal):` instead?

wlauto/common/android/workload.py
```diff
@@ -267,7 +303,12 @@ def _grant_requested_permissions(self):
...
+                    break
+            else:
+                self.device.execute("pm grant {} {}".format(self.package, permission))
```
setrofim: instead of the inner `for` loop:

    permission_name = permission.rsplit('.', 1)[1]
    if permission_name not in ANDROID_NORMAL_PERMISSIONS:
        self.device.execute("pm grant {} {}".format(self.package, permission))

PR #104, opened 2016-02-16
Moved WA over to devlib

ep1cman:
WA devices have been removed in favour of "DeviceManager"s. These basically wrap devlib functionality and add some WA specific features.


wlauto/common/android/workload.py
```diff
@@ -483,9 +483,9 @@ def _deploy_resource_tarball(self, context, resource_file, timeout=300):
                 raise WorkloadError(message.format(resource_file, self.name))
             # adb push will create intermediate directories if they don't
             # exist.
-            self.device.push_file(asset_tarball, ondevice_cache, timeout=timeout)
+            self.device.push(asset_tarball, ondevice_cache, timeout=timeout)
 
-        device_asset_directory = self.device.path.join(self.device.external_storage_directory, 'Android', kind)
+        device_asset_directory = self.device.path.join(self.context.device_manager.external_storage_directory, 'Android', kind)
```
setrofim: should be just `context` rather than `self.context`

PR #111, opened 2016-03-07
daq: Added check for duplicate channel labels

ep1cman:
The daq instrument will no longer accept duplicate channel names.
This caused issues where files sent from the daq sever were being
overwritten.

### DIFF
--- a/wlauto/instrumentation/daq/__init__.py
+++ b/wlauto/instrumentation/daq/__init__.py
@@ -286,6 +286,10 @@ class Daq(Instrument):
         if self.labels:
             if len(self.labels) != len(self.resistor_values):
                 raise ConfigError('Number of DAQ port labels does not match the number of resistor values.')
+
+            duplicates = set([x for x in self.labels if self.labels.count(x) > 1])
+            if len(duplicates) > 0:
+                raise ConfigError('Duplicate labels: {}'.format(', '.join(duplicates)))
         else:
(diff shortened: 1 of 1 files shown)


wlauto/instrumentation/daq/__init__.py
```diff
@@ -286,6 +286,10 @@ def validate(self):  # pylint: disable=too-many-branches
         if self.labels:
             if len(self.labels) != len(self.resistor_values):
                 raise ConfigError('Number of DAQ port labels does not match the number of resistor values.')
+
+            duplicates = set([x for x in self.labels if self.labels.count(x) > 1])
+            if len(duplicates) > 0:
```
setrofim: Could be just `if duplicates:`

wlauto/instrumentation/daq/__init__.py
```diff
@@ -286,6 +286,10 @@ def validate(self):  # pylint: disable=too-many-branches
...
+                raise ConfigError('Duplicate labels: {}'.format(', '.join(duplicates)))
```
setrofim: from wlauto.utils.misc import unique
and then report `unique(duplicates)§

ep1cman: does the set() in 290 not do this already?

setrofim: ah, so it does.

PR #113, opened 2016-03-09
freq_sweep: Improved documentation

ep1cman:
- Added explanation that this instrument does not taskset workloads
 - Fixed formatting issue with the agenda example


wlauto/instrumentation/freqsweep/__init__.py
```diff
@@ -33,6 +33,9 @@ class FreqSweep(Instrument):
         - Setting the runner to 'by_spec' increases the chance of successfully
           completing an agenda without encountering hotplug issues
         - If possible disable dynamic hotplug on the target device
+        - This instrument does not automatically taskset workloads to the cores
+          being swept, it is most likely that this is behaviour you want. To
```
setrofim: This is ambiguous -- suggests that *not* tasksetting is the "behaviour you want"; I assume the oposite is meant. Should be rephrased.

PR #133, opened 2016-04-06
CpuUtilisationTimeline added. This now will generate cpu utilisation …

mcgeagh:
…based on frequencies and a number of samples

### DIFF
--- a/wlauto/result_processors/cpustate.py
+++ b/wlauto/result_processors/cpustate.py
@@ -102,7 +102,13 @@ class CpuStatesProcessor(ResultProcessor):
                   Create a CSV with the timeline of core power states over the course of the run
                   as well as the usual stats reports.
                   """),
-
+        Parameter('create_utilization_timeline', kind=bool, default=False,
+                  description="""
+                  Create a CSV with the timeline of cpu(s) utilisation over the course of the run
+                  as well as the usual stats reports.
+                  The values generated are floating point numbers, normalised based on the maximum
+                  frequency of the cluster.
+                  """),
     ]
 
     def validate(self):
@@ -130,6 +136,7 @@ class CpuStatesProcessor(ResultProcessor):
         self.idle_state_names = [idle_states[i] for i in sorted(idle_states.keys())]
         self.num_idle_states = len(self.idle_state_names)
         self.iteration_reports = OrderedDict()
+        self.max_freq_list = []
         # priority -19: just higher than the slow_start of instrumentation
         signal.connect(self.set_initial_state, signal.BEFORE_WORKLOAD_EXECUTION, priority=-19)
 
@@ -139,12 +146,17 @@ class CpuStatesProcessor(ResultProcessor):
         # Write initial frequencies into the trace.
         # NOTE: this assumes per-cluster DVFS, that is valid for devices that
         # currently exist. This will need to be updated for per-CPU DFS.
+        # pylint: disable=attribute-defined-outside-init
         self.logger.debug('Writing initial frequencies into trace...')
         device = context.device
         cluster_freqs = {}
+        cluster_max_freqs = {}
+        self.max_freq_list = []
         for c in unique(device.core_clusters):
             cluster_freqs[c] = device.get_cluster_cur_frequency(c)
+            cluster_max_freqs[c] = device.get_cluster_max_frequency(c)
         for i, c in enumerate(device.core_clusters):
+            self.max_freq_list.append(cluster_max_freqs[c])
             entry = 'CPU {} FREQUENCY: {} kHZ'.format(i, cluster_freqs[c])
             device.set_sysfile_value('/sys/kernel/debug/tracing/trace_marker',
                                      entry, verify=False)
@@ -165,6 +177,10 @@ class CpuStatesProcessor(ResultProcessor):
             timeline_csv_file = os.path.join(context.output_directory, 'power_states.csv')
         else:
             timeline_csv_file = None
+        if self.create_utilization_timeline:
+            cpu_utilisation = os.path.join(context.output_directory, 'cpu_utilisation.csv')
+        else:
+            cpu_utilisation = None
         parallel_report, powerstate_report = report_power_stats(  # pylint: disable=unbalanced-tuple-unpacking
             trace_file=trace.path,
             idle_state_names=self.idle_state_names,
@@ -175,6 +191,8 @@ class CpuStatesProcessor(ResultProcessor):
             first_system_state=self.first_system_state,
             use_ratios=self.use_ratios,
             timeline_csv_file=timeline_csv_file,
+            cpu_utilisation=cpu_utilisation,
+            max_freq_list=self.max_freq_list,
         )
         if parallel_report is None:
             self.logger.warning('No power state reports generated; are power '
--- a/wlauto/utils/power.py
+++ b/wlauto/utils/power.py
@@ -515,6 +515,39 @@ class PowerStateStatsReport(object):
                                            for s in stats])
 
 
+class CpuUtilisationTimeline(object):
+
+    def __init__(self, filepath, core_names, max_freq_list):
+        self.filepath = filepath
+        self._wfh = open(filepath, 'w')
+        self.writer = csv.writer(self._wfh)
+        if core_names:
+            headers = ['ts'] + ['{} CPU{}'.format(c, i)
+                                for i, c in enumerate(core_names)]
+            self.writer.writerow(headers)
+        self._max_freq_list = max_freq_list
+
+    def update(self, timestamp, core_states):  # NOQA
+        row = [timestamp]
+        for core, [idle_state, frequency] in enumerate(core_states):
+            if idle_state == -1:
+                if frequency == UNKNOWN_FREQUENCY:
+                    frequency = 0
+            elif idle_state is None:
+                frequency = 0
+            else:
+                frequency = 0
+            if core < len(self._max_freq_list):
+                frequency /= float(self._max_freq_list[core])
+                row.append(frequency)
+            else:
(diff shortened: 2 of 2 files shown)


mcgeagh:
Added capability to write a CSV file containing CPU utilisation timeline information when using cpustates result_processor.

wlauto/utils/power.py
```diff
@@ -577,7 +614,7 @@ def build_idle_domains(core_clusters,   # NOQA
 def report_power_stats(trace_file, idle_state_names, core_names, core_clusters,
                        num_idle_states, first_cluster_state=sys.maxint,
                        first_system_state=sys.maxint, use_ratios=False,
-                       timeline_csv_file=None, filter_trace=False):
+                       timeline_csv_file=None, filter_trace=False, cpu_utilisation=None, cpu_averaging=1000):
```
ep1cman: I would move these two down to a new line

ep1cman:
Using the this agenda:
```
config:
    device: generic_android
    instrumentation: [trace-cmd]
    result_processors: [cpustates]
    trace-cmd:
        events: [cpu_idle, cpu_frequency]
    device-config:
        core_names: [a,a,a,a,b,b,b,b]
        core_clusters: [1,1,1,1,0,0,0,0]
workloads:
    - dhrystone
```


I get the following error:
```
ERROR    "cpustates" result processor has encountered an error
ERROR    TypeError("unsupported operand type(s) for +: 'int' and 'NoneType'")
ERROR    Traceback (most recent call last):
ERROR      File "/work/home/main_wa/workload-automation/wlauto/core/result.py", line 120, in _handle_errors
ERROR        yield
ERROR      File "/work/home/main_wa/workload-automation/wlauto/core/result.py", line 84, in add_result
ERROR        processor.process_iteration_result(result, context)
ERROR      File "/work/home/main_wa/workload-automation/wlauto/result_processors/cpustate.py", line 180, in process_iteration_result
ERROR        cpu_utilisation=cpu_utilisation
ERROR      File "/work/home/main_wa/workload-automation/wlauto/utils/power.py", line 642, in report_power_stats
ERROR        reporter.update(timestamp, states)
ERROR      File "/work/home/main_wa/workload-automation/wlauto/utils/power.py", line 547, in update
ERROR        row.append(sum(i) / len(i))
ERROR    TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'

```

mcgeagh:
Thanks for highlighting this. my test devices never hit that corner case so I didnt pick up this flaw. Last commit resolves it. Let me know if there is anything else.

setrofim:
please could you squash your commits.

wlauto/result_processors/cpustate.py
```diff
@@ -163,8 +163,10 @@ def process_iteration_result(self, result, context):
         self.logger.debug('Generating power state reports from trace...')
         if self.create_timeline:
             timeline_csv_file = os.path.join(context.output_directory, 'power_states.csv')
+            cpu_utilisation = os.path.join(context.output_directory, 'cpu_utilisation.csv')
```
ep1cman: This should have its own parameter with a proper description instead of being a part of create_timeline.

ep1cman:
Can you please clarify what this is supposed to show? From my understanding it is the "average frequency" which isn't CPU utilisation. Here are my results:

```
ts		        b CPU0	b CPU1	b CPU2	b CPU3	l CPU4	l CPU5	l CPU6	l CPU7
80964.044916	1002001	1467609	1522664	863577	1899495	1865682	0	0
80965.077646	1066494	1573000	1573000	814814	1989000	1989000	0	0
80965.892844	1088516	1573000	1573000	817960	1989000	1989000	0	0
80966.848816	1079078	1573000	1573000	827398	1989000	1989000	0	0
80967.69183		1027169	1573000	1573000	821106	900666	1157598	0	0
80968.356885	843661	937625	1171040	673062	801216	638976	0	0
80969.02221		539071	358852	365495	420329	1989000	1989000	0	0
80969.637148	551746	389584	450528	438945	1989000	1989000	0	0
80970.285714	572234	359216	346515	459082	1989000	1989000	0	0
```

wlauto/result_processors/cpustate.py
mcgeagh: the bool create_timeline is for creating a csv with the timeline of core power states. This is doing the same except for cpu frequency averaging. If you would like it to be its own parameter, I will first have to rename the existing one to be more explicit. Any recommendations? timeline_powerstate for example? In which case I will create a new one called timeline_cpufreq

mcgeagh:
cpustate doesnt give us utilisation information. As such, I infer utilisation by first setting all idle frequencies to zero and then averaging the frequencies over a number of samples. E.g. if 5 out of 10 samples are zero and the other 5 are 2GHz, i will get 1GHz, which is 50% utilisation for that 'sample period'. It is not exact sadly.
If you have alternative ideas (/proc/state is not accurate enough) then feel free to suggest :) I am happy to learn more.

setrofim:
You can normalize them against the highest frequency to get % utilization.

wlauto/result_processors/cpustate.py
ep1cman: for now leave existing prameter `create_timeline` for compatibility and the new parameter should be `create_utilization_timeline`. The difference between them should be made clear with the descriptions (updating `create_timelines`'s if need be too).

In WA v3 which we are currently working on we can add aliases to parameters and we will rename original parameter to `create_powerstate_timeline`, aliasing it as `create_timeline`

mcgeagh:
To get a % instead of the Frequency, yes. And I do that already in post processing of the data.
Perhaps my terminology is off. To me this gives me utilisation, not as a percentage but as a frequency, but not as a pre-defined freq value based on the dvfs table which steps...
It averages the frequency over a number of samples to smooth it out so the charts generated are smoother and fewer.
If you would like me to rename from utilisation, I am happy to (sounds like 'average' is what it should be instead.)

setrofim:
I'm fine with a rename -- the main thing is that the name accurately reflects what is being reported (and this should be clear in the parameter description as well).
Though, IMO, utilisation would be a more generally-useful metric...

mcgeagh:
So I've been thinking.... we could separate out the averaging and the utilisation, and have them as parameters. That way if the user wants a % instead of the value, they can set the utilisation parameter to true. And for me, I am more interested in the averaging. This is already a parameter currently, but i can start renaming things to make it all clearer that they are separate things that are interchangeable. 
Example: an average_rate of 1 (or 0/not_defined) and utilisation to false will just give you the raw numbers without modification. Any combination of these should be valid.

The next question is..... this sounds like its a generic results_processor and thus shouldn't live in power/cpustate, as one may want utilisation and/or averaging on any csv generated by any of the other instrumentation/result_processors. E.g. FPS

Let me know which route is preferable and I can start looking at it.

wlauto/result_processors/cpustate.py
mcgeagh: Didn't see this until now. Renamed

mcgeagh:
Averaging removed. Utilisation only now.
Renamed parameter to suggestion.
Squashed commits.

wlauto/utils/power.py
```diff
@@ -515,6 +515,39 @@ def write(self, filepath):
...
+        for idle_state, frequency in core_states:
+            if idle_state == -1:
+                if frequency == UNKNOWN_FREQUENCY:
+                    frequency = 0
+            elif idle_state is None:
+                frequency = 0
+            else:
+                frequency = 0
+            row.append(frequency)
+            self._max = max(self._max, frequency)
```
ep1cman: The max frequency can vary for different clusters meaning littles would never show 100% utilisation even though they are at 100%. Also if the system is running idle this would report higher utilisation than is actually occuring. You should get the max frequency out of cpufreq (which is accessible via the WA device). This should also error if cpufreq is not available on a device.

wlauto/result_processors/cpustate.py
```diff
@@ -102,6 +102,11 @@ class CpuStatesProcessor(ResultProcessor):
                   Create a CSV with the timeline of core power states over the course of the run
                   as well as the usual stats reports.
                   """),
+        Parameter('create_utilization_timeline', kind=bool, default=False,
+                  description="""
+                  Create a CSV with the timeline of cpu(s) utilisation over the course of the run
+                  as well as the usual stats reports.
+                  """),
```
ep1cman: Can you expand this description to explain that it is an normalised average of frequencies. When you look at the raw CSV you only see a few values and this can be confusing until a sliding window average is performed

mcgeagh:
Better description added.
Changed utilisation to be based on the cluster's maximum possible frequency

wlauto/result_processors/cpustate.py
```diff
@@ -142,9 +149,13 @@ def set_initial_state(self, context):
         self.logger.debug('Writing initial frequencies into trace...')
         device = context.device
         cluster_freqs = {}
+        cluster_max_freqs = {}
+        del self.max_freq_list[:]
```
ep1cman: `self.max_freq_list = []`

ep1cman:
Once you have addressed any comments, checked the code with pylint/pep8 (scripts can be found in `dev_scripts`) and tested to make sure it all works, could you squash your commits and we can then merge this in

wlauto/result_processors/cpustate.py
mcgeagh: I did this originally, but pylint gave me this:
`W:153, 8: Attribute 'max_freq_list' defined outside __init__ (attribute-defined-outside-init)`
As such i changed it to del instead.
If you prefer this behaviour, i will add:
`# pylint: disable=attribute-defined-outside-init`

ep1cman: the pylint ignore should be fine

mcgeagh:
Done.

PR #135, opened 2016-04-11
Only check for crashed content if crash_check is true.

mcgeagh:


ep1cman:
good catch, could you rename the commit to start with "fps:" so we know what it relates to. Once thats done I'll merge it

mcgeagh:
Done.

PR #138, opened 2016-04-13
fps: Can now process multiple 'view' attributes

mcgeagh:
Previously fps only captured a single 'view' attribute given to it by the workload.
This change allows you to pass a list of 'views' and the latency command will be run for every view in the list - if that view is currently visible.

### DIFF
--- a/wlauto/instrumentation/fps/__init__.py
+++ b/wlauto/instrumentation/fps/__init__.py
@@ -230,11 +230,10 @@ class LatencyCollector(threading.Thread):
     #command_template = 'while (true); do dumpsys SurfaceFlinger --latency {}; sleep 2; done'
     command_template = 'dumpsys SurfaceFlinger --latency {}'
 
-    def __init__(self, outfile, device, activity, keep_raw, logger):
+    def __init__(self, outfile, device, activities, keep_raw, logger):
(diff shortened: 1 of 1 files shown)


wlauto/instrumentation/fps/__init__.py
```diff
@@ -244,6 +243,9 @@ def __init__(self, outfile, device, activity, keep_raw, logger):
         self.drop_threshold = self.refresh_period * 1000
         self.exc = None
         self.unresponsive_count = 0
+        if isinstance(activity, basestring):
+            activity = [activity]
+        self.activity = activity
```
setrofim: Should be renamed to `activities` as is now a list

mcgeagh:
Done

wlauto/instrumentation/fps/__init__.py
```diff
@@ -254,7 +256,10 @@ def run(self):
             wfh = os.fdopen(fd, 'wb')
             try:
                 while not self.stop_signal.is_set():
-                    wfh.write(self.device.execute(self.command))
+                    view_list = self.device.execute('dumpsys SurfaceFlinger --list').split()
+                    for i in self.activities:
```
ep1cman: I would change i for activity

PR #144, opened 2016-04-15
servo_power: Added support for chromebook servo boards

ep1cman:
Servo is a debug board used for Chromium OS test and development. Among other uses, it allows
access to the built in power monitors (if present) of a Chrome OS device. More information on
Servo board can be found in the link bellow:

 https://www.chromium.org/chromium-os/servo

based on: https://github.com/vflouris/workload-automation/commit/03ede10739add543251b7827621370e647878a26
and: https://github.com/freedomtan/workload-automation/commit/9a0dc55b55fa7a806703a1a43901ea91f14c2e6a

### DIFF
--- a/wlauto/instrumentation/servo_power_monitors/__init__.py
+++ b/wlauto/instrumentation/servo_power_monitors/__init__.py
@@ -1,0 +1,228 @@
+#    Copyright 2016 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+# pylint: disable=W0613,E1101,attribute-defined-outside-init
+from __future__ import division
+import os
+import subprocess
+import signal
+import csv
+import threading
+import time
+import getpass
+import logging
+import xmlrpclib
+from datetime import datetime
+
+from wlauto import Instrument, Parameter, Executable
+from wlauto.exceptions import InstrumentError, ConfigError
+from wlauto.utils.types import list_of_strings
+from wlauto.utils.misc import check_output
+from wlauto.utils.cros_sdk import CrosSdkSession
+from wlauto.utils.misc import which
+
+
+class ServoPowerMonitor(Instrument):
+
+    name = 'servo_power'
+    description = """
+    Collects power traces using the Chromium OS Servo Board.
+
+    Servo is a debug board used for Chromium OS test and development. Among other uses, it allows
+    access to the built in power monitors (if present) of a Chrome OS device. More information on
+    Servo board can be found in the link bellow:
+
+    https://www.chromium.org/chromium-os/servo
+
+    In order to use this instrument you need to be a sudoer and you need a chroot environment. More
+    information on the chroot environment can be found on the link bellow:
+
+    https://www.chromium.org/chromium-os/developer-guide
+
+    If you wish to run servod on a remote machine you will need to allow it to accept external connections
+    using the `--host` command line option, like so:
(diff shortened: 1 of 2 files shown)


wlauto/instrumentation/servo_power_monitors/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+from wlauto.utils.types import list_of_strings
+from wlauto.utils.misc import check_output
+from wlauto.utils.cros_sdk import CrosSdkSession
+from wlauto.utils.misc import which
+
+
+class ServoPowerMonitor(Instrument):
+
+    name = 'servo_power'
+    description = """Collects power traces using the Chromium OS Servo Board.
```
setrofim: Indent to align with "description"

wlauto/instrumentation/servo_power_monitors/__init__.py
```diff
@@ -0,0 +1,178 @@
+#    Copyright 2015 ARM Limited
```
setrofim: copyright should be this year

wlauto/utils/cros_sdk.py
```diff
@@ -0,0 +1,123 @@
+#    Copyright 2015 ARM Limited
```
setrofim: copyright should be this year

wlauto/utils/cros_sdk.py
```diff
@@ -0,0 +1,123 @@
...
+    def __init__(self, cros_path, password=''):
+        self.in_chroot = True if which('dut-control') else False
+        ON_POSIX = 'posix' in sys.builtin_module_names
+        if self.in_chroot:
+            self.cros_sdk_session = Popen(['/bin/sh'], bufsize=1, stdin=PIPE, stdout=PIPE, stderr=PIPE,
+                                          cwd=cros_path, close_fds=ON_POSIX, shell=True)
+        else:
+            cros_sdk_bin_path = which('cros_sdk')
+            if not cros_sdk_bin_path:
+                raise HostError("Failed to locate 'cros_sdk' make sure it is in your PATH")
```
JaviMerino: you don't need to have `cros_sdk` in your PATH, you can find it in `chromium/tools/depot_tools`.  If `which` fails, you could try to locate it there and if it's an executable, use it.

wlauto/instrumentation/servo_power_monitors/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+            self.logger.info('Instrument {} requires sudo acces on this machine'.format(self.name))
+            self.logger.info('You need to be sudoer to use it.')
+            password = getpass.getpass()
+            check = subprocess.call('echo {} | sudo -S ls > /dev/null'.format(password), shell=True)
+            if check:
+                raise InstrumentError('Given password was either wrong or you are not a sudoer')
+        self.server_session = CrosSdkSession(self.chroot_path, password=password)
+        self.cros_session = CrosSdkSession(self.chroot_path, password=password)
+        password = ''
+        self.server_session.send_command('sudo servod -b {b} -c {b}.xml&'.format(b=self.board_name))
```
JaviMerino: It would be good if starting servo was optional (a configuration parameter).  Sometimes I already have my servod running for serial and I don't want to kill it for `wa`.

JaviMerino:
Another minor nit: you create one csv file per power rail.  If you're going to generate a csv, why not put all the data collected in the file?

JaviMerino:
I can work around these nits for now so from my point of view, this can be merged as is.  They can be addressed later.

wlauto/instrumentation/servo_power_monitors/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+        self.cros_session = cros_session
+
+    def run(self):
+        while self.term_signal.is_set():
+            time.sleep(self.period)
+            self.cros_session.send_command(self.command)
+            lines = self.cros_session.get_lines()
+            for i, line in enumerate(lines):
+                _, value = line.split(':')
+                self.power_data[i].append(value)
```
JaviMerino: You can't discard the time.  The output csv needs to have the time at which the measurement was taken for us to properly calculate the energy consumed.  I know I can estimate the time from the `sampling_rate` but there is no point in estimating when we already have the information.

ep1cman: This isnt time thats being discarded its the name of the power domain the reading is from e.g:

PR #153, opened 2016-05-03
Juno fixes

ep1cman:
juno:
- The default boot args have been removed since these cause issues with the latest Linaro builds, which boot correctly without any bootargs.
- made a regex string a raw-string.

uboot: 
- Now detects the U-Boot version to use correct line endings

### DIFF
--- a/wlauto/utils/uboot.py
+++ b/wlauto/utils/uboot.py
@@ -16,7 +16,7 @@
 import re
 import time
 import logging
-
+from distutils.version import LooseVersion
 from wlauto.utils.serial_port import TIMEOUT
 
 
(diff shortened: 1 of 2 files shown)


wlauto/utils/uboot.py
```diff
@@ -16,6 +16,8 @@
 import re
 import time
 import logging
+import re
```
setrofim: already imported on line 16

wlauto/utils/uboot.py
```diff
@@ -56,6 +61,7 @@ def open(self, timeout=default_timeout):
 
         """
         self.conn.expect(self.start_prompt, timeout)
+        self._set_new_line_sep()
```
setrofim: should be just `_set_line_sep`, or better, `_set_line_separator`

wlauto/utils/uboot.py
```diff
@@ -114,3 +120,14 @@ def empty_buffer(self):
             pass
         self.conn.buffer = ''
 
+    def _set_new_line_sep(self):
+        uboot_text = self.conn.before
+        uboot_ver = self.uboot_regex.findall(uboot_text)
+
+        if uboot_ver:
+            year, month = uboot_ver[0].split(".")
```
setrofim: This is brittle -- assumes a very specific version format. use `distutils.version.LooseVersion` for comparison instead.

wlauto/utils/uboot.py
```diff
@@ -32,19 +32,22 @@ class UbootMenu(object):
     option_regex = re.compile(r'^\[(\d+)\]\s+([^\r]+)\r\n', re.M)
     prompt_regex = re.compile(r'^([^\r\n]+):\s*', re.M)
     invalid_regex = re.compile(r'Invalid input \(max (\d+)\)', re.M)
+    uboot_regex = re.compile(r"U-Boot\s(\d[\w.-]+)\s")
```
setrofim: I would change parenthesized part of the regex to `(\d\S*)`

PR #162, opened 2016-05-10
workloads: add blogbench workload

chase-qi:
Blogbench is a portable filesystem benchmark that tries to reproduce the
load of a real-world busy file server.


wlauto/workloads/blogbench/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+                  description='The number of iterations to run'),
+        Parameter('directory', kind=str,
+                  default='/mnt/blogbench',
+                  description='Specify an empty and writable directory')
+    ]
+
+    def initialize(self, context):
+        host_exe = context.resolver.get(Executable(self, self.device.abi,
+                                                   'blogbench'))
+        BlogbenchWorkload.binary = self.device.install(host_exe)
```
ep1cman: Use install_if_needed instead (this will only push the binary if its not already on the device

wlauto/workloads/blogbench/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+            for line in fh:
+                if any('Final score for ' + x in line
+                       for x in ['writes', 'reads']):
+                    line = line.split(':')
+                    metric = line[0].lower().strip().replace(' ', '_')
+                    score = int(line[1].strip())
+                    context.result.add_metric(metric, score, 'blogs')
+
+    def finalize(self, context):
+        self.device.uninstall(self.binary)
```
ep1cman: We don't usually remove binaries from the device, is there a reason you need to for this workload?

wlauto/workloads/blogbench/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Source code are available from:
+        https://download.pureftpd.org/pub/blogbench/
+    """
+
+    parameters = [
+        Parameter('iterations', kind=int, default=30,
+                  description='The number of iterations to run'),
+        Parameter('directory', kind=str,
+                  default='/mnt/blogbench',
```
ep1cman: Might be better to put this the working directory by default

wlauto/workloads/blogbench/__init__.py
chase-qi: Thanks for the tip! Updated.

wlauto/workloads/blogbench/__init__.py
chase-qi: No, not really, removed this line.

wlauto/workloads/blogbench/__init__.py
chase-qi: Agree, updated the code, not sure if it is the right approach ;)
Please review.

wlauto/workloads/blogbench/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Source code are available from:
+        https://download.pureftpd.org/pub/blogbench/
+    """
+
+    parameters = [
+        Parameter('iterations', kind=int, default=30,
+                  description='The number of iterations to run'),
+        Parameter('directory', kind=str,
+                  description='Specify an empty and writable directory')
```
setrofim: Please remove this parameter as this is no longer necessary

wlauto/workloads/blogbench/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+        host_binary = context.resolver.get(Executable(self, self.device.abi,
+                                                   'blogbench'))
+        self.binary = self.device.install_if_needed(host_binary)
+
+    def setup(self, context):
+        # Total test duration equal to iteration*frequency
+        # The default frequency is 10 seconds, plus 5 here as a buffer.
+        self.timeout = self.iterations * 15
+        if self.directory is None:
+            self.directory = self.device.working_directory + '/blogbench'
```
setrofim: This should be

    self.directory = self.device.path.join(self.device.working_directory, 'blogbench')

chase-qi: Thanks for your review! Updated.

wlauto/workloads/blogbench/__init__.py
chase-qi: The total size of the data produced by the benchmark equal to memory capacity. When the memory capacity is big, it may run into low disk space issue. IMHO, we should keep this parameter so that user can specify another partition or storage device if needed. Please let me know if it is ok to keep it.

setrofim: If this is an issue, you can already set an alternative working directory for WA via `device_config`. However everything generated/pushed by WA to the target device should be kept under its working directory. We do not want to start leaving stuff all over the place.

chase-qi: Cool, that solves this issue. Parameter removed.

wlauto/workloads/blogbench/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        self.binary = self.device.install_if_needed(host_binary)
+        # Total test duration equal to iteration*frequency
+        # The default frequency is 10 seconds, plus 5 here as a buffer.
+        self.timeout = self.iterations * 15
+        # An empty and writable directory is needed.
+        self.directory = self.device.path.join(self.device.working_directory,
+                                               'blogbench')
+        self.device.execute('rm -rf {}'.format(self.directory), timeout=60)
+        self.device.execute('mkdir -p {}'.format(self.directory))
+        self.results = os.path.join(self.device.working_directory,
```
ep1cman: Should be `self.device.path.join`

chase-qi: Sorry that I missed this comment. It is fixed now.

wlauto/workloads/blogbench/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+            for line in blogbench_output:
+                if any('Final score for ' + x in line
+                       for x in ['writes', 'reads']):
+                    line = line.split(':')
+                    metric = line[0].lower().strip().replace(' ', '_')
+                    score = int(line[1].strip())
+                    context.result.add_metric(metric, score, 'blogs')
+
+    def finalize(self, context):
+        self.device.execute('rm -rf {}'.format(self.directory), timeout=60)
```
ep1cman: I just tested this and this command timed out on me, I would recommend setting the timeout to 300

chase-qi: Thanks for your review! I have increased the timeout to 300 on this line. Did the same update on line 50.

PR #164, opened 2016-05-10
Poller: Added an instrument to poll files and output a csv of their v…

ep1cman:
…alues


wlauto/instrumentation/poller/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+# limitations under the License.
+# pylint: disable=access-member-before-definition,attribute-defined-outside-init,unused-argument
+
+from wlauto import Instrument, Parameter, Executable
+from wlauto.exceptions import ConfigError
+from wlauto.utils.types import list_of_strs
+
+
+class Poller(Instrument)
+    name = 'poller'
```
setrofim: I would make the name less generic. E.g. "sysfs_poller" or "file_poller".

wlauto/instrumentation/poller/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    description = """
+    Polls the given files at a set sample interval. The values are output in CSV format.
```
setrofim: It's worth elaborating in the following paragraphs -- how/where will the results appear (e.g. on per-iteration basis or in the top-level output)? What files can be specified? (e.g. does this assume that there will be a single integer or float in the file)? etc.

wlauto/instrumentation/poller/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    parameters = [
+        Parameter('sample_interval', kind=int, default=1000,
+                  description="""The interval between samples in mS."""),
+        Parameter('files', kind=list_of_strs,
```
setrofim: I would use `list_or_string` as the kind here. If a user is interested in polling a single file (a common use case), they should be able to specify it without needing to make it into a list.

wlauto/instrumentation/poller/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+                  description="""A list of paths to the files to be polled"""),
+        Parameter('as_root', kind=bool, default=True,
```
setrofim: Default to `False`?

wlauto/instrumentation/poller/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    def validate(self):
+        if not self.device.is_rooted and self.as_root:
+            raise ConfigError('The device is not rooted, cannot run poller as root.')
+
+    def initialize(self, context):
+        #for file_ in self.files:
+        #    print self.device.execute("if [ -e '{}' ]; then echo 1; else echo 0; fi".format(file_))
+        #    if not self.device.is_file(file_):
+        #        raise ConfigError('The file {} does not exist'.format(file_))
+
```
setrofim: What is this?

wlauto/instrumentation/poller/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+                                              self.target_output_path)
+
+    def start(self, context):
+        self.device.kick_off(self.command, as_root=self.as_root)
+
+    def stop(self, context):
+        self.device.killall('poller', signal='TERM', as_root=self.as_root)
+
+    def update_result(self, context):
+        self.device.pull_file(self.target_output_path, context.output_directory)
```
setrofim: Should delete `self.target_output_path` inside `teardown()`. Depending on the number of files monitored, polling rate and the duration of the run, these files can potentially be quite sizeable -- we don't want to lave them hanging around.

wlauto/instrumentation/poller/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+
```
setrofim: Superfluous empty line at the end (should be just one).

wlauto/instrumentation/poller/poller.c
```diff
@@ -0,0 +1,58 @@
...
+    double time_float;
+   
+    //Print headers and open files to poll  
+    printf("time");
+    int filesToPoll[argc - 2];
+    int i;
+    for (i = 0; i < argc-2; i++)
+    {
+        filesToPoll[i] = open(argv[i+2], O_RDONLY);
+	printf(",%s", argv[i+2]);
```
setrofim: alignment issues in the above two lines

wlauto/instrumentation/poller/poller.c
```diff
@@ -0,0 +1,58 @@
...
+    //Setup SIGTERM handler
+    struct sigaction action;
+    memset(&action, 0, sizeof(struct sigaction));
+    action.sa_handler = term;
+    sigaction(SIGTERM, &action, NULL);
+
+    //Poll files 
+    while (!done) {
+        gettimeofday(&current_time, NULL); 
+       	time_float = (double)current_time.tv_sec;
```
setrofim: Indentation.

wlauto/instrumentation/poller/poller.c
```diff
@@ -0,0 +1,58 @@
...
+        time_float += ((double)current_time.tv_usec)/1000/1000;
+        printf("%f", time_float);
+	for (i = 0; i < argc-2; i++) {
+            read(filesToPoll[i], buf, 1024);
+            lseek(filesToPoll[i], 0, SEEK_SET);
+            buf[strcspn(buf, "\n")] = 0; //Removes "\n"
+	    printf(",%s", buf);
+        }
+	printf("\n");
+        usleep((useconds_t)atoi(argv[1]));
```
setrofim: Indentation. Also, I'd keep all argument parsing in one place, at the begging. So you would have

    useconds_t polling_period =  atoi(argv[1])

inside your `parse_args`, followed by just a 

     usleep(polling_period);

wlauto/instrumentation/poller/poller.c
```diff
@@ -0,0 +1,58 @@
...
+    }
```
setrofim: You should close the files you've opened (this will also make sure that everything you've written has been flushed).

wlauto/instrumentation/poller/poller.yaml
```diff
@@ -0,0 +1,8 @@
+config:
```
setrofim: Rather than having this in a separate file that no one will look at -- user's aren't expected to look thorough the code, this should be part of the description. Also, I would remove everything not germane to the example (e.g. your current example can be taken to imply that this only works with android devices).

wlauto/instrumentation/poller/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+        if not self.device.is_rooted and self.as_root:
+            raise ConfigError('The device is not rooted, cannot run poller as root.')
+
+    def initialize(self, context):
+        host_poller = context.resolver.get(Executable(self, self.device.abi,
+                                                      "poller"))
+        target_poller = self.device.install_if_needed(host_poller)
+
+        self.target_output_path = self.device.path.join(self.device.working_directory, 'poller.csv')
+        self.command = '{} {} {} > {}'.format(target_poller,
```
setrofim: From the poller code, it looks like the interval is specified with `-t`. Shouldn't this format string be `'{} -t {} {} > {}'`?

wlauto/instrumentation/poller/poller.c
```diff
@@ -0,0 +1,108 @@
...
+    extern char *optarg;
+    extern int optind;
+    int c = 0;
+    int show_help = 0;
+    useconds_t interval = 1000000;
+    char buf[1024];
+    memset(buf, 0, sizeof(buf));
+    struct timeval current_time;
+    double time_float;
+    int filesToPoll[argc-optind];
```
setrofim: Should be `files_to_poll` to be consistent with the naming of other variables.

PR #165, opened 2016-05-13
workloads: add stress_ng workload

chase-qi:
stress-ng will stress test a computer system in various selectable ways.
It was designed to exercise various physical subsystems of a computer as
well as the various operating system kernel interfaces.


wlauto/workloads/stress_ng/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+                              'of online processors.'),
+        Parameter('duration', kind=int, default=60,
+                  description='Timeout for test execution in seconds')
+    ]
+
+    def setup(self, context):
+        host_binary = context.resolver.get(Executable(self, self.device.abi,
+                                                      'stress-ng'))
+        self.binary = self.device.install_if_needed(host_binary)
+        self.log = os.path.join(self.device.working_directory,
```
ep1cman: should be `self.device.path.join`

wlauto/workloads/stress_ng/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+                                'stress_ng_output.txt')
+        self.results = os.path.join(self.device.working_directory,
```
ep1cman: Same as above

wlauto/workloads/stress_ng/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+                                    'stress_ng_results.yaml')
+        self.command = ('{} --{} {} --timeout {}s --log-file {} --yaml {} '
+                        '--metrics-brief --verbose'
+                        .format(self.binary, self.stressor, self.threads,
+                                self.duration, self.log, self.results))
+        self.timeout = self.duration + 10
+
+    def run(self, context):
+        self.output = self.device.execute(self.command, timeout=self.timeout,
+                                          as_root=True)
```
setrofim: Since you're running as root, you should add an `initialize()` method and in it check that the device is rooted (`self.device.is_rooted`), raising `WorkloadError` if it isn't.

wlauto/workloads/stress_ng/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+            results = yaml.load(stress_ng_results)
+
+        try:
+            metric = results['metrics'][0]['stressor']
+            throughput = results['metrics'][0]['bogo-ops']
+            context.result.add_metric(metric, throughput, 'ops')
+        # For some stressors like vm, if test duration is too short, stress_ng
+        # may not able to produce test throughput rate.
+        except TypeError:
+            pass
```
setrofim: It's probably worth logging a warning here stating that a result could not be obtained because the test duration was set too short. Better yet, if you know which stressors require longer tests, it's worth adding a `validate()` method where you check that the duration specified is compatible with the specified stressor.

setrofim:
Please also include a LICENSE file for the binaries that you're adding. It should state what license the binaries are under, link to the full license text and to the binaries' source.

wlauto/workloads/stress_ng/__init__.py
chase-qi: Fixed. Thanks!

wlauto/workloads/stress_ng/__init__.py
chase-qi: Fixed. Thanks!

wlauto/workloads/stress_ng/__init__.py
chase-qi: Added. Thanks!

wlauto/workloads/stress_ng/__init__.py
chase-qi: Thank you for the advice! 

I tested all of these 20 stressors on pandaboard, seems vm is the only one have this issue, I have added the validate method to raise config error if the duration is less than 60s. 

However, the condition may vary depending on hardware speed and configuration. I guess we still should keep the try..except... clauses. stress-ng main focus on stress test, throughput rate missing probably is fine.

chase-qi:
I have copied the "COPYING" file from stress-ng git repo to here. It uses GPL v2.0, too. That is why I didn't include it before. Please let me know if it is not needed.

wlauto/workloads/stress_ng/__init__.py
chase-qi: As you suggested, just added a sentence in except clause to log a warning if it is not covered by validate() method.

chase-qi:
Ping ;)
Is there any plan to merge this workload?

PR #167, opened 2016-05-13
cameracapture & camerarecord: Updated workloads to work with Android M+

ep1cman:
The stock camera app as of Android M has changed. This commit updates
the ui automation to work with this new app. As part of this change
it was required to bump the API level of the ui automation to 18.

Also made the teardown of the capture workload close the app like the record workload does.


wlauto/workloads/cameracapture/__init__.py
```diff
@@ -39,13 +39,20 @@ def __init__(self, device, **kwargs):
         super(Cameracapture, self).__init__(device, **kwargs)
         self.uiauto_params['no_of_captures'] = self.no_of_captures
         self.uiauto_params['time_between_captures'] = self.time_between_captures
+        self.uiauto_params['version'] = "button"
```
setrofim: Don't use "version" for this, as this is used for version numbers.

wlauto/workloads/cameracapture/__init__.py
```diff
@@ -39,13 +39,20 @@ def __init__(self, device, **kwargs):
...
 
     def setup(self, context):
         super(Cameracapture, self).setup(context)
         self.device.execute('am start -n {}/{}'.format(self.package, self.activity))
 
+    def initialize(self, context):
+        if self.device.get_sdk_version() >= 23:
+            self.package = "com.google.android.GoogleCamera"
+            self.uiauto_params['version'] = "swipe"
```
setrofim: I would do this via a mapping -- that way if we have to update this in future android versions, we won't have a mass of if/elif statements.

wlauto/utils/types.py
```diff
@@ -300,3 +300,12 @@ def extend(self, values):
...
 
+
+class range_dict(dict):
+
+    def __getitem__(self, i):
+        i = int(i)
+        while i not in self and i > 0:
+            i -= 1
+        if i > 0:
+            return dict.__getitem__(self, i)
```
setrofim: I would also implement a `__setitem__` to make sure that only integers can be used as keys.

wlauto/utils/types.py
```diff
@@ -300,3 +300,12 @@ def extend(self, values):
     def __str__(self):
         return ' '.join(self)
 
+
+class range_dict(dict):
```
setrofim: Add a doc string describing the behavior of this type

wlauto/utils/types.py
```diff
@@ -300,3 +300,31 @@ def extend(self, values):
...
+
+    If a key is not in the dict it will search downward until
+    the next key and return its value. E.g:
+
+    If:
+        a[5] = "Hello"
+        a[10] = "There"
+
+    Then:
+        a[2] > None
```
setrofim: Greater then? `==`

wlauto/utils/types.py
```diff
@@ -111,7 +111,7 @@ def list_of_numbers(value):
     """
     Value must be iterable. All elements will be converted to numbers (either ``ints`` or
     ``float``\ s depending on the elements).
-
+ibm bar
```
setrofim: "ibm bar" ?

wlauto/utils/types.py
```diff
@@ -300,3 +299,32 @@ def extend(self, values):
...
+        a[7] == "Hello"
+        a[999] == "There"
+
+    """
+    def __getitem__(self, i):
+        i = int(i)
+        while i not in self and i > 0:
+            i -= 1
+        if i <= 0:
+            raise KeyError(i)
```
setrofim: `KeyError` should be raised with the original i

PR #169, opened 2016-05-24
AndroidDevice: Improved gathering of build props

ep1cman:
These are now gathered via `getprop` rather than trying to parse the
build.prop file directly.

This fixes issues with build.prop files that have imports.

### DIFF
--- a/wlauto/common/android/device.py
+++ b/wlauto/common/android/device.py
@@ -21,6 +21,7 @@ import time
 import tempfile
 import shutil
 import threading
+import json
 from subprocess import CalledProcessError
 
 from wlauto.core.extension import Parameter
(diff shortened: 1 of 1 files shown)


wlauto/common/android/device.py
```diff
@@ -508,17 +509,18 @@ def get_properties(self, context):
...
-
+        self._update_build_properties(props)
+        
         dumpsys_target_file = self.path.join(self.working_directory, 'window.dumpsys')
         dumpsys_host_file = os.path.join(context.host_working_directory, 'window.dumpsys')
         self.execute('{} > {}'.format('dumpsys window', dumpsys_target_file))
         self.pull_file(dumpsys_target_file, dumpsys_host_file)
         context.add_run_artifact('dumpsys_window', dumpsys_host_file, 'meta')
+        
+        prop_file = os.path.join(context.host_working_directory, 'props.json')
```
setrofim: Maybe call it `'android-props.json'` -- would make it more clear for people browsing WA results who aren't necessarily familiar with WA.

PR #175, opened 2016-06-01
Revent fixes

ep1cman:
- Device model name is now used when searching for revent files
- All revent instances are killed during teardown (kills revent if run aborted)

### DIFF
--- a/wlauto/common/android/workload.py
+++ b/wlauto/common/android/workload.py
@@ -321,19 +321,24 @@ class ReventWorkload(Workload):
             super(ReventWorkload, self).__init__(device, **kwargs)
         devpath = self.device.path
         self.on_device_revent_binary = devpath.join(self.device.binaries_directory, 'revent')
-        self.on_device_setup_revent = devpath.join(self.device.working_directory, '{}.setup.revent'.format(self.device.name))
-        self.on_device_run_revent = devpath.join(self.device.working_directory, '{}.run.revent'.format(self.device.name))
         self.setup_timeout = kwargs.get('setup_timeout', self.default_setup_timeout)
         self.run_timeout = kwargs.get('run_timeout', self.default_run_timeout)
         self.revent_setup_file = None
         self.revent_run_file = None
+        self.on_device_setup_revent = None
+        self.on_device_run_revent = None
 
-    def init_resources(self, context):
+    def initialize(self, context):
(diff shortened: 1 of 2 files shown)


wlauto/common/android/workload.py
```diff
@@ -321,18 +321,21 @@ def __init__(self, device, _call_super=True, **kwargs):
...
         self.run_timeout = kwargs.get('run_timeout', self.default_run_timeout)
         self.revent_setup_file = None
         self.revent_run_file = None
+        self.on_device_setup_revent = None
+        self.on_device_run_revent = None
 
-    def init_resources(self, context):
+    def setup(self, context):
         self.revent_setup_file = context.resolver.get(wlauto.common.android.resources.ReventFile(self, 'setup'))
         self.revent_run_file = context.resolver.get(wlauto.common.android.resources.ReventFile(self, 'run'))
```
setrofim: This should be done inside `initialize` rather than `setup`.

wlauto/resource_getters/standard.py
```diff
@@ -380,11 +385,15 @@ def resolve_resource(self, resource):
...
         elif resource.name == 'revent':
-            filename = '.'.join([resource.owner.device.name, resource.stage, 'revent']).lower()
-            for asset in assets:
-                pathname = os.path.basename(asset['path']).lower()
-                if pathname == filename:
-                    return asset
+            device_model = resource.owner.device.get_device_model()
+            wa_device_name = resource.owner.device.name
+            for name in [device_model, wa_device_name]:
+                if name:
```
setrofim: Rather than doing

    for name in list:
        if name:
            # large block

Generally, the fewer indentation levels three are, the better.
I'd go with

    for name in list:
         if not name:
              continue
         # large block

wlauto/resource_getters/standard.py
```diff
@@ -446,26 +455,31 @@ def try_get_resource(self, resource, version, remote_path, local_path):
...
+                    # 'revent_files' subdirectory under it, if it exists.
+                    if os.path.isdir(alternate_location):
+                        for candidate in os.listdir(alternate_location):
+                            if candidate.lower() == filename.lower():
+                                return os.path.join(alternate_location, candidate)
+                    if os.path.isdir(location):
+                        for candidate in os.listdir(location):
+                            if candidate.lower() == filename.lower():
+                                return os.path.join(location, candidate)
         else:
```
setrofim: This `else` could use a comment

    else:  # unknown resource name

PR #177, opened 2016-06-08
servo_power: Added check for device platform.

ep1cman:
Now checks to see if the device is running chromeOS.

### DIFF
--- a/wlauto/instrumentation/servo_power_monitors/__init__.py
+++ b/wlauto/instrumentation/servo_power_monitors/__init__.py
@@ -117,6 +117,10 @@ class ServoPowerMonitor(Instrument):
         self.poller = None
         self.data = None
         self.stopped = True
+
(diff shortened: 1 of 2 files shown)


wlauto/instrumentation/servo_power_monitors/__init__.py
```diff
@@ -117,6 +117,10 @@ def initialize(self, context):
         self.poller = None
         self.data = None
         self.stopped = True
+
+        if self.device.platform != "chromeos":
+            raise DeviceError("servo_power instrument only supports chromeOS devices.")
```
setrofim: `chromeOS` --> `Chrome OS`

setrofim: Also this should be raising an `IntrumentError` rather than a `DeviceError` (the latter is reserved for when an error occurs on/with the device).

PR #190, opened 2016-06-29
Apk version check

muendelezaji:
Opening this for discussion, originally was [this PR](https://github.com/jimboatarm/workload-automation/pull/118)

**Replace APK during install**
 - Use `-g` and `-r` flags with `adb install` if necessary. The `-g` flag is used to grant all runtime permissions at installation time on API 23 and above. The `-r` flag tells `adb` to replace existing APK if found (sometimes uninstall does not fully remove the package e.g. for built-in apps).

**Check APK version and ABI**
 - When the `version` attribute is specified in the workload, the execution will now validate it against the provided APK's `versionName` property, and raise a `WorkloadError` if they do not match.

 - Added `check_apk_abi` parameter to `ApkWorkload` which is disabled by default. If enabled in the workload/agenda, the resource resolver will try to locate the APK in the ABI-specific folder (see note below) and raise an error if not found.
By default (i.e. when disabled), if APK is not found in this path the resolver falls back to the `root` path as per current behaviour.

**Note:**
- The ABI check follows the same principle as that for deploying binary files [described here]
(https://github.com/jimboatarm/workload-automation/blob/master/doc/source/writing_extensions.rst#deploying-executables-to-a-device). Folder path is in the form `<root>/apk/<abi>/<apk_name>`. Where `root` is the base resource location e.g. `~/.workload_automation/dependencies/<extension_name>`.

### DIFF
--- a/wlauto/common/android/device.py
+++ b/wlauto/common/android/device.py
@@ -349,18 +349,24 @@ class AndroidDevice(BaseLinuxDevice):  # pylint: disable=W0223
                            timeout=self.default_timeout)
         return bool(int(output))
 
-    def install(self, filepath, timeout=default_timeout, with_name=None):  # pylint: disable=W0221
+    def install(self, filepath, timeout=default_timeout, with_name=None, replace=False):  # pylint: disable=W0221
         ext = os.path.splitext(filepath)[1].lower()
         if ext == '.apk':
-            return self.install_apk(filepath, timeout)
+            return self.install_apk(filepath, timeout, replace)
         else:
             return self.install_executable(filepath, with_name)
 
-    def install_apk(self, filepath, timeout=default_timeout):  # pylint: disable=W0221
+    def install_apk(self, filepath, timeout=default_timeout, replace=False):  # pylint: disable=W0221
         self._check_ready()
         ext = os.path.splitext(filepath)[1].lower()
         if ext == '.apk':
-            return adb_command(self.adb_name, "install '{}'".format(filepath), timeout=timeout)
+            flags = []
+            if replace:
+                flags.append('-r') # Replace existing APK
+            if self.get_sdk_version() >= 23:
+                flags.append('-g') # Grant all runtime permissions
+            self.logger.debug("Replace APK = {}, ADB flags = '{}'".format(replace, ' '.join(flags)))
+            return adb_command(self.adb_name, "install {} '{}'".format(' '.join(flags), filepath), timeout=timeout)
         else:
             raise DeviceError('Can\'t install {}: unsupported format.'.format(filepath))
 
--- a/wlauto/common/android/device.py
+++ b/wlauto/common/android/device.py
@@ -362,9 +362,9 @@ class AndroidDevice(BaseLinuxDevice):  # pylint: disable=W0223
         if ext == '.apk':
             flags = []
             if replace:
-                flags.append('-r') # Replace existing APK
+                flags.append('-r')  # Replace existing APK
             if self.get_sdk_version() >= 23:
-                flags.append('-g') # Grant all runtime permissions
+                flags.append('-g')  # Grant all runtime permissions
             self.logger.debug("Replace APK = {}, ADB flags = '{}'".format(replace, ' '.join(flags)))
             return adb_command(self.adb_name, "install {} '{}'".format(' '.join(flags), filepath), timeout=timeout)
         else:
--- a/wlauto/common/android/workload.py
+++ b/wlauto/common/android/workload.py
@@ -20,6 +20,7 @@ import time
 from wlauto.core.extension import Parameter
 from wlauto.core.workload import Workload
 from wlauto.core.resource import NO_ONE
+from wlauto.common.android.resources import ApkFile
 from wlauto.common.resources import ExtensionAsset, Executable
 from wlauto.exceptions import WorkloadError, ResourceError, ConfigError
 from wlauto.utils.android import ApkInfo, ANDROID_NORMAL_PERMISSIONS
@@ -160,6 +161,11 @@ class ApkWorkload(Workload):
                   '''),
         Parameter('uninstall_apk', kind=boolean, default=False,
                   description='If ``True``, will uninstall workload\'s APK as part of teardown.'),
+        Parameter('check_abi', kind=bool, default=False,
+                  description='''
+                  If ``True``, workload will check that the APK matches the target
+                  device ABI, otherwise any APK found will be used.
+                  '''),
     ]
 
     def __init__(self, device, _call_super=True, **kwargs):
@@ -169,12 +175,14 @@ class ApkWorkload(Workload):
         self.apk_version = None
         self.logcat_log = None
 
-    def init_resources(self, context):
-        self.apk_file = context.resolver.get(wlauto.common.android.resources.ApkFile(self),
+    def initialize(self, context):
+        # Get APK for the correct version and device ABI
+        self.apk_file = context.resolver.get(ApkFile(self, self.device.abi),
                                              version=getattr(self, 'version', None),
+                                             check_abi=getattr(self, 'check_abi', False),
+                                             variant_name=getattr(self, 'variant_name', None),
                                              strict=self.check_apk)
-
-    def validate(self):
+        # Validate the APK
         if self.check_apk:
             if not self.apk_file:
                 raise WorkloadError('No APK file found for workload {}.'.format(self.name))
@@ -223,7 +231,7 @@ class ApkWorkload(Workload):
             if installed_version:
                 self.device.uninstall(self.package)
             # It's possible that the uninstall above fails, which might result in a warning
-            # and/or failure during installation. Hower execution should proceed, so need
+            # and/or failure during installation. However execution should proceed, so need
             # to make sure that the right apk_vesion is reported in the end.
             if self.install_apk(context):
                 self.apk_version = host_version
--- a/wlauto/common/android/workload.py
+++ b/wlauto/common/android/workload.py
@@ -155,8 +155,8 @@ class ApkWorkload(Workload):
                   '''),
         Parameter('force_install', kind=boolean, default=False,
                   description='''
-                  Always re-install the APK, even if matching version is found
-                  on already installed on the device.
+                  Always re-install the APK, even if matching version is found already installed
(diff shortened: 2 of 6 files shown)


wlauto/common/android/device.py
```diff
@@ -109,6 +109,15 @@ def is_rooted(self):
...
 
+    """
+    Returns the ABI for this device.
+    This is the same as the ``abi`` property except it does not rely on the ``_is_ready`` flag
+    being set, and thus can be run directly from the host console. Useful when we need to provide
+    ABI-specific handling during initialization before the property is set.
+    """
+    def get_abi(self):
+        return adb_shell(self.adb_name, 'getprop ro.product.cpu.abi').split('-')[0]
+
```
setrofim: You should be using abi() above. The ``_is_ready`` flag is set once a connection to the device has been established; You cannot assume you're able to execute arbitrary commands on the device before the flag is set.

wlauto/common/android/workload.py
```diff
@@ -170,14 +175,25 @@ def __init__(self, device, _call_super=True, **kwargs):
...
             if not self.apk_file:
                 raise WorkloadError('No APK file found for workload {}.'.format(self.name))
+
+            # If workload specifies version, ensure APK version is the expected one
+            if getattr(self, 'version', None):
+                host_version = ApkInfo(self.apk_file).version_name
+                self.logger.debug('workload.version={}, host_version={}'.format(self.version, host_version))
+                if self.version != host_version:
+                    message = 'APK dependency version mismatch. Required: {}; Found: {}'
+                    raise WorkloadError(message.format(self.version, host_version))
```
setrofim: Why is this check necessary? It would already  have been done when resolving `self.apk_file`.

wlauto/resource_getters/standard.py
```diff
@@ -45,6 +45,9 @@ class PackageFileGetter(ResourceGetter):
     Looks for exactly one file with the specified extension in the owner's directory. If a version
     is specified on invocation of get, it will filter the discovered file based on that version.
     Versions are treated as case-insensitive.
+    If an ABI is specified in the resource request, then the getter will try to locate the file in
+    the ABI-specific folder in the form ``<root>/<file_extension>/<abi>/<filename>``. Where ``root``
+    is the base resource location e.g. ``~/.workload_automation/dependencies/<extension_name>``
```
setrofim: All of this should be implemented inside `PackageApkGetter` rather than here.

wlauto/resource_getters/standard.py
```diff
@@ -522,3 +540,16 @@ def get_owner_path(resource):
         return os.path.join(os.path.dirname(__base_filepath), 'common')
     else:
         return os.path.dirname(sys.modules[resource.owner.__module__].__file__)
+
+
+def get_abi_specific_path(resource, base_dir, extension, enforce=False):
+    # Note: Using device.get_abi() method instead of the property device.abi property
+    # The property is not yet set or available at this point in the execution flow
+    abi_dir = os.path.join(base_dir, extension, resource.owner.device.get_abi())
```
setrofim: This should not be assuming that resource's owner will have a "device". The abi (or more generally, "platform") should be set in the `resource` description object (as is done for executable binaries).

wlauto/common/android/device.py
```diff
@@ -349,18 +349,24 @@ def file_exists(self, filepath):
...
 
-    def install_apk(self, filepath, timeout=default_timeout):  # pylint: disable=W0221
+    def install_apk(self, filepath, timeout=default_timeout, replace=False):  # pylint: disable=W0221
         self._check_ready()
         ext = os.path.splitext(filepath)[1].lower()
         if ext == '.apk':
-            return adb_command(self.adb_name, "install '{}'".format(filepath), timeout=timeout)
+            flags = []
+            if replace:
+                flags.append('-r') # Replace existing APK
```
setrofim: Two spaces between code and inline comment.

wlauto/common/android/device.py
```diff
@@ -349,18 +349,24 @@ def file_exists(self, filepath):
...
+            if self.get_sdk_version() >= 23:
+                flags.append('-g') # Grant all runtime permissions
```
setrofim: Two spaces between code and inline comment.

wlauto/common/android/workload.py
```diff
@@ -222,8 +222,8 @@ def initialize_with_host_apk(self, context, installed_version):
         if self.force_install:
             if installed_version:
                 self.device.uninstall(self.package)
-            # It's possible that that the uninstall above fails, which will result in
-            # install failing and a warning, hower execution would the proceed, so need
+            # It's possible that the uninstall above fails, which might result in a warning
+            # and/or failure during installation. Hower execution should proceed, so need
```
setrofim: Hower --> However

wlauto/resource_getters/standard.py
muendelezaji: Changed. Would it also make sense to rename this method to `get_platform_specific_path`, and the `kwargs` argument as `check_platform` (or something similar)?

setrofim: Yes

wlauto/resource_getters/standard.py
```diff
@@ -522,3 +540,16 @@ def get_owner_path(resource):
...
+    if enforce:
+        return abi_dir # Only if explicitly requested to enforce the resource ABI check
+    else:
+        # Otherwise for backwards compatibility, default to previous behaviour and log the event
+        message = 'Resource {} not found in {}. Defaulting to {}'
+        logging.getLogger('ResourceResolver').debug(message.format(resource, abi_dir, base_dir))
```
setrofim: This should be the logger for the specific resource getter, not for resolver.

wlauto/common/android/workload.py
muendelezaji: The resolver currently only checks the APK file name, 
`filelist = [ff for ff in filelist if version.lower() in os.path.basename(ff).lower()]`
so can be misled to use any APK as long as it contains the version string in its name.

This verifies `version` against the APK's `versionName` property, so is independent of the file name and thus harder to "cheat" (if a bit overkill).

setrofim: That check should be updated to use the real version instead of doing this. We should not have different checks of different versions conducted in different way in different parts of the code. That makes thing unnecessarily complicated.

wlauto/resource_getters/standard.py
muendelezaji: What's the suggested way of getting the resource getter context here? (as it is not in scope)

wlauto/common/android/device.py
muendelezaji: Done

wlauto/common/android/device.py
muendelezaji: Done

wlauto/common/android/workload.py
muendelezaji: Done

wlauto/resource_getters/standard.py
setrofim: Pass the logger as a parameter?

wlauto/common/android/workload.py
muendelezaji: So just to clarify, we replace the old (filename) check in the resolver with the `version_name` property check?

wlauto/resource_getters/standard.py
muendelezaji: I tried to avoid that if possible, but works if it's fine by you.

wlauto/common/android/workload.py
setrofim: Yes. The intention behind the two checks is the same; just the method for executing them is different (the original check was implemented before we were using aapt)

wlauto/resource_getters/standard.py
setrofim: it's either that, or raise an exception here; catch and log at the point of invocation. I'm fine with either approach.

wlauto/resource_getters/standard.py
muendelezaji: There are two ways of implementing this:
 1. Move the ABI check to `{Environment,Package}ApkGetter` and override the `get()` method, which means duplicating the code, or
 1. Add a method to provide `resource_dir` in `{Environment,Package}FileGetter` with a default implementation that returns the current value. The `ApkGetter` subclasses will override this to return the platform specific path, as would any other subclasses that might need this (good extensibility).

setrofim: `get()` method is three lines, one of which needs to change; so you end up duplicating just two lines of code; to me, this seems preferable to the additional complexity of method 2

muendelezaji:
Changes:
- [x] Add `platform` attribute to `ApkFile` resource class
- [x] Remove `get_abi()` from `AndroidDevice` and use resource `platform` instead
- [x] In `ApkWorkload`, get and validate `apk_file` after device is ready and ABI can be determined
- [x] Replace old APK version check (filename only) with `aapt`'s `versionName` check
- [x] Move ABI-specific resource resolution to `{Environment,Package}ApkGetter`
- [x] Remove redundant `get_abi_specific_path()` in resource resolver

setrofim:
The updated changes look good. However, please could you update the previous commits with the changes, rather than submitting them in a separate commit.

Also, have you tested this with existing versioned apk workloads (such as Antutu, Geekbench, etc) to make sure that the work with the new version matching?

Finally, please address pylint/pep8 issues

muendelezaji:
Cleaned up pylint and combined review commit into previous.

Regarding existing workloads, apart from `dex2oat` and `peacekeeper`, all other versioned workloads run as expected with the new check in place.
These two actually utilise the `version` param as a "hack" to provide the app *name* (see below).

**OK**
- [x] antutu
- [x] cameracapture
- [x] camerarecord
- [x] geekbench
- [x] glbenchmark
- [x] vellamo

**Fail**
- [ ] dex2oat  
resolver is invoked as follows  
`self.apk_file = context.resolver.get(wlauto.common.android.resources.ApkFile(peacekeeper), version='chrome')`
- [ ] peacekeeper  
`self.version = self.browser`, where `browser` is `'firefox'` or `'chrome'`

What do you suggest we do to these two?

setrofim:
For these, the old method based on file name should be used. However, there should be a different parameter used for that rather than "version" (e.g. "variant").

muendelezaji:
Updated and re-tested. Please review.

PR #196, opened 2016-07-11
file-poller: Improved csv output

### LINKED ISSUE #159: Temporal temperature, frequency measurements
WA - Feature Request:

We need below measurements over time (with configurable sampling interval) from Android devices.
1) Each cluster frequency.
2) SoC Temperature.
Both above parameters sampled in the same time scale - so that they can be plotted (frequency vs Temperature) together.

ep1cman:
All ',' and '\n' will now be stripped from the files contents so it doesn't
effect csv formatting

Also fixed some whitespace

Fixes #159


wlauto/instrumentation/poller/poller.c
```diff
@@ -14,8 +14,21 @@ void term(int signum)
     done = 1;
 }
 
+// From: http://stackoverflow.com/questions/1515195/how-to-remove-n-or-t-from-a-given-string-in-c
+void strip(char *s) {
+    char *p2 = s;
+    while(*s != '\0') {
+        if(*s != ',' && *s != '\n') {
+            *p2++ = *s++;
```
setrofim: Why `*p2++` ?

PR #197, opened 2016-07-11
file_poller fixes.

setrofim:


wlauto/instrumentation/poller/poller.c
```diff
@@ -115,15 +128,24 @@ int main(int argc, char ** argv) {
...
         printf("%f", time_float);
-        for (i = 0; i < (argc - optind); i++) {
-            read(files_to_poll[i], buf, 1024);
-            lseek(files_to_poll[i], 0, SEEK_SET);
+        for (i = 0; i < num_files; i++) {
+            lseek(files_to_poll[i].fd, 0, SEEK_SET);
+            bytes_read = read(files_to_poll[i].fd, buf, 1024);
+
+            if (bytes_read < 0) {
+                fprintf(stderr, "WARNING: Read nothing form \"%s\"\n",
```
ep1cman: "form" > "from"

wlauto/instrumentation/poller/poller.c
```diff
@@ -115,15 +128,24 @@ int main(int argc, char ** argv) {
...
+                        files_to_poll[i].path);
+                printf(",");
+                continue;
+            }
+
             strip(buf);
             printf(",%s", buf);
             buf[0] = '\0'; // "Empty" buffer
```
ep1cman: is this necessary with your changes?
(previously if nothing was read it would print the previous value, is this still the case?)

setrofim: You're right, it's no longer necessary, but it does no harm and may prevent issues if the above code is modified further in the future -- "belts and suspenders".

PR #201, opened 2016-07-13
Add per-action instrumentation for UX performance

johnrchrdsn:
- Implement a new Marker API in the BaseUiAutomation so workload can
  generate start and end markers with string name. Outputs to logcat.

- Document the Marker output log format in the WA source.

- Create a results processor to take existing instrument fps
  logs and parse them based on the workload markers. Produce per-action
  fps metrics.

- Add simple timing results based on the workload markers

- Move common processing logic used in result processor and
  instrumentation into a separate utility file.


wlauto/utils/fps.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+
+    def __init__(self, context, data, generate_csv=True, action=None):
+        """
+        context      - the wa execution context
+        data         - a pandas.DataFrame object with frame data (e.g. frames.csv)
+        generate_csv - if true create a csv file containing the metrics
+        action       - output metrics names with additional action specifier
+        """
+        self.context = context
```
setrofim: Utils modules are intended to be stand-alone bits of re-usable functionality. They cannot rely on the rest of WA framework. This should not be ware of "context" or "metrics" . Rather than adding the metrics to the contest's results, this should be returning the the calculated results and the jtg

wlauto/result_processors/uxperf.py
```diff
@@ -0,0 +1,229 @@
...
+    '''
+    Parses logcat messages for UX Performance markers.
+
+    UX Performance markers are output from logcat under a debug priority. The
+    logcat tag for the marker messages is UX_PERF. The messages associated with
+    this tag consist of a name for the action to be recorded and a timestamp.
+    These fields are delimited by underscores. e.g.
+
+    <TAG>   : <MESSAGE>
+    UX_PERF : gestures_swipe_left_start_861975087367
```
setrofim: Would it not be easier/clearer to have the timestamp separated from the tag by a space?

wlauto/external/uiauto/src/com/arm/wlauto/uiauto/BaseUiAutomation.java
```diff
@@ -37,6 +38,28 @@
...
+        private String testTag;
+        private Bundle parameters;
+
+        public ActionLogger(String testTag, Bundle parameters) {
+            this.testTag = testTag;
+            this.parameters = parameters;
+        }
+
+        public void start() {
+            if (Boolean.parseBoolean(parameters.getString("dumpsys_enabled"))) {
```
setrofim: I would parse the parameters inside the constructor and save the parsed values as members, it will result in both clearer code, and early failure in case of bad parameter values been passed.

wlauto/external/uiauto/src/com/arm/wlauto/uiauto/BaseUiAutomation.java
```diff
@@ -37,6 +38,28 @@
 
 public class BaseUiAutomation extends UiAutomatorTestCase {   
 
+    public class ActionLogger {
```
setrofim: Where is this being instantiated?

setrofim: Maybe add a docstring  with some documentation describing what it is good for, and how it is intended to be used.

setrofim:
The modification to FPS and the UXPerf stuff are two logically-separate sets of changes; they should be split into two commits.

PR #202, opened 2016-07-13
camerarecord: add possibility to select slow_motion recording mode

mdigiorgio:
Signed-off-by: Michele Di Giorgio <michele.digiorgio@arm.com>

### DIFF
--- a/wlauto/workloads/camerarecord/__init__.py
+++ b/wlauto/workloads/camerarecord/__init__.py
@@ -15,6 +15,7 @@
 
 from wlauto import UiAutomatorWorkload, Parameter
 from wlauto.utils.types import range_dict
+from wlauto.exceptions import WorkloadError
 
 
 class Camerarecord(UiAutomatorWorkload):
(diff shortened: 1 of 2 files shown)


setrofim:
The code looks good. Are you reasonably sure this will work across different android versions (at least KitKat onwards), or is a version check needed?

mdigiorgio:
I've implemented this feature only inside the `recordVideoGoogleV3_2()` and tested it on the N5X (so Android M). I'm not hundred percent sure about previous versions, are you able to give it a try? I don't have other devices on which to test this.

setrofim:
Yes, you're right, that will only be invoked for M onwards, so it should be fine. However you should add a check in `setup` of the workload to raise a `WorkloadError` if `self.device.get_android_sdk_version()` is less than 23.

mdigiorgio:
OK, I provided a check for that in `camerarecord/__init__.py:setup()`.

wlauto/workloads/camerarecord/__init__.py
```diff
@@ -48,8 +51,14 @@ def initialize(self, context):
         version = self.device.get_installed_package_version(self.package)
         version = version.replace(' ', '_')
         self.uiauto_params['version'] = version
+        self.uiauto_params['recording_mode'] = self.recording_mode
 
     def setup(self, context):
+        if self.recording_mode != 'normal':
+            if self.device.get_sdk_version() < 23:
+                raise WorkloadError('{} recording mode only supported by '\
```
setrofim: Format specifier never used.

Have you run pylint/pep8 checkers for this?

mdigiorgio: Oops, forgot it sorry. Going to change it and push again.

PR #204, opened 2016-07-19
External libs and initial workload for UX Performance

johnrchrdsn:
Extensions to the BaseUiAutomation class and a new UxPerfUIAutomation class for using with UX performance workloads. Includes the googlephotos workload to demonstrate the use of the action logger.

Note: API level for the build has been increased to 18 to allow resourceId call for UiSelector objects.


setrofim:
The "Add UxPerfUiAutomation class and googlephotos workload" should be split into two commits as UxPerfUi stuff and the googlephotos are two logically separate changes, and the former does not depend on the latter.

wlauto/workloads/googlephotos/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+        super(Googlephotos, self).validate()
+        self.uiauto_params['package'] = self.package
+        self.uiauto_params['output_dir'] = self.device.working_directory
+        self.uiauto_params['dumpsys_enabled'] = self.dumpsys_enabled
+
+    def initialize(self, context):
+        super(Googlephotos, self).initialize(context)
+
+        # Check for workload dependencies before proceeding
+        jpeg_files = [entry for entry in os.listdir(self.dependencies_directory) if entry.lower().endswith(('.jpg', '.jpeg'))]
```
setrofim: You should not expect your assets to be in a particular location. Instead, you should you the resource resource mechanism to discover them.

wlauto/workloads/googlephotos/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if len(jpeg_files) < 4:
+            raise NotFoundError("This workload requires a minimum of four {} files in {}".format('jpe?g',
+                                self.dependencies_directory))
+        else:
+            for entry in jpeg_files:
+                self.device.push_file(os.path.join(self.dependencies_directory, entry),
+                                      os.path.join(self.device.working_directory, entry),
```
setrofim: The on-device path should be constructed using `self.device.path.join`

johnrchrdsn:
The googlephotos workload extends from UxPerfUiAutomation and uses its confirmAccess method that is why I grouped them into one commit. With that in mind would you still like to have the workload and base class as two separate commits?

setrofim:
Yes. googlephotos depends on UxPerfUi, but not the other way round. They are two logical changes -- one enhances and expands core UI workload functionality, the other uses that new functionality to implement a specific workload.

Presumably, there will be other commits in the future adding workloads that utilize the new common functionality that have nothing to do with googlephotos. It is conceivable that in the future, for whatever reason, we may need to to revert the googlemaps changes, but keep the common functionality so that the other workloads keep working. Hence, it's useful to have them in separate commits.

wlauto/workloads/googlephotos/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+       degrees and 270 degrees.
+    """
+
+    parameters = [
+        Parameter('dumpsys_enabled', kind=bool, default=True,
+                  description="""
+                  If ``True`` turns on the action logger which outputs
+                  timestamps to logcat for actions recorded in the workload.
+                  """),
+        Parameter('test_images', kind=list_or_string, mandatory=True,
```
setrofim: Why is th kind `list_or_string`, when only a list is expected? Should probably be 'list_of_strings' instead.

setrofim: Why is this a mandatory parameter?

Presumably, images used will have a significant impact on the measurements collected; results collected using two different image sets would not be comparable. Therefore the images are an intrinsic part of the test, rather than an input into it. 

While I can see someone wanted to use this to examine issues with specific images, so it is useful to have the *option* to override the default set your own, the default should be provided. Moreover, any results collected using something other than the images that are part of the tests, need to be identified as such in the output.

wlauto/workloads/googlephotos/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+
+        for image in self.test_images:
+            if not image.endswith(('jpg', 'jpeg')):
+                raise ValidationError('{} must be a jpeg file'.format(image))
+
+    def initialize(self, context):
+        super(Googlephotos, self).initialize(context)
+
+        for image in self.test_images:
+            image_path = context.resolver.get(wlauto.common.resources.File(self, image))
```
ep1cman: Would also make sense to allow users to specify the absolute path to images. Check if the path is a file, if not then try the resource resolvers?

wlauto/workloads/googlephotos/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+            self.device.push_file(image_path, self.device.working_directory, timeout=300)
+
+        # Force a re-index of the mediaserver cache to pick up new files
+        self.device.execute('am broadcast -a android.intent.action.MEDIA_MOUNTED -d file:///sdcard')
+
+    def teardown(self, context):
+        super(Googlephotos, self).teardown(context)
+
+        for entry in self.device.listdir(self.device.working_directory):
+            if entry.endswith(".log"):
```
ep1cman: This seems a bit broad, there may be log file from things other than this workload in the working directory.

setrofim:
Also, just spotted, please remove `template-agenda.mustache` from the PR -- this does not belong in the code.

wlauto/external/uiauto/src/com/arm/wlauto/uiauto/UxPerfUiAutomation.java
```diff
@@ -0,0 +1,66 @@
...
+        public GestureTestParams(GestureType gesture, PinchType pinchType, int steps, int percent) {
+            this.gestureType = gesture;
+            this.gestureDirection = Direction.NULL;
+            this.pinchType = pinchType;
+            this.steps = steps;
+            this.percent = percent;
+        }
+    }
+
+    public void confirmAccess() throws Exception {
```
ep1cman: WA should handle granting permissions when an app is being installed. This shouldn't need to be part of the automation.

wlauto/external/uiauto/project.properties
```diff
@@ -11,4 +11,4 @@
 #proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
 
 # Project target.
-target=android-17
+target=android-18
```
ep1cman: You have upgraded the required API level for the base class to 18, but not for all of the existing automation that would now break. I would recommend creating a separate commit upgrading to API level 18 because this change breaks compatibility and would affect a lot of files.

wlauto/workloads/googlephotos/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+                  """),
+    ]
+
+    def validate(self):
+        super(Googlephotos, self).validate()
+        self.uiauto_params['package'] = self.package
+        self.uiauto_params['output_dir'] = self.device.working_directory
+        self.uiauto_params['dumpsys_enabled'] = self.dumpsys_enabled
+
+        self._check_image_numbers()
```
ep1cman: Why were these split out into functions?

wlauto/workloads/googlephotos/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+       the crop markers so the image is tilted positively, reset and then negatively to get a
+       similar cropping effect.
+    6. A Rotate test is performed on a selected image, rotating anticlockwise 90 degrees, 180
+       degrees and 270 degrees.
+    """
+
+    default_test_images = [
+        'uxperf_1200x1600.png', 'uxperf_1600x1200.jpg',
+        'uxperf_2448x3264.png', 'uxperf_3264x2448.jpg',
+    ]
```
ep1cman: Can you also include these default images with the workload in this commit

wlauto/external/uiauto/project.properties
johnrchrdsn: Took on board your suggestions and after discussing with the rest of the team added some default test image names we will be using. These will be pushed at a later date in a separate dependency repo. At the moment we are trying to introduce these changes gradually.

I updated the .properties files to the new API requirement but I didn't rebuild the workloads since nothing in them has changed and they should run as before. Thank you for your input so far.

wlauto/workloads/googlephotos/__init__.py
johnrchrdsn: I thought it was more explicit and readable that way. Any particular objections?

wlauto/workloads/googlephotos/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def _check_image_duplicates(self):
+        if len(self.test_images) != len(set(self.test_images)):
+            raise ValidationError('Duplicate image names not allowed')
+
+    def _info_image_used(self):
+        if set(self.test_images) & set(self.default_test_images):
+            self.logger.info('Using default test images')
+        else:
+            self.logger.warning('Using custom test images')
```
setrofim: This is not the right way to do this.

1. A warning implies something has gone wrong; which is not the case here.
2. Someone looking at results is unlikely to go through the log file, so they won't see.

My suggestion is that either the images themselves are saved as part of the results or their names and hashes are (e.g. in a file in the output called "USED_IMAGES").

wlauto/workloads/googlephotos/__init__.py
setrofim: This was discussed earlier and the images will be hosted separately on a publicly-visible assets server (most likely a separate GitHub repo).

wlauto/external/uiauto/project.properties
setrofim: I'm not comfortable merging `googlephotos` workload until it's assets are available, as we'll have a workload that users are unable to run. Please move the commit into a separate pull request.

wlauto/workloads/googlephotos/__init__.py
ep1cman: I think some comments would achieve this better than moving the code to the bottom of the file

wlauto/external/uiauto/project.properties
setrofim: Regarding the .properities

   - I don't see a commit with that change; have you pushed it?
   - Are *certain* that no APIs used by existing automations remove/renamed/changed between levels 17 and 18? If not please rebuild everything to make sure.

PR #206, opened 2016-07-21
[WIP] Add script to get external assets for workloads

muendelezaji:
Note:
WA's `HttpGetter` class downloads assets to `~/.workload_automation/dependencies/__remote/<workload>`, while workloads seem to check for dependencies in `~/.workload_automation/dependencies/<workload>` or their own working directory.

Currently testing this against a workload with dependencies and will update the PR with the outcome.

### DIFF
--- a/scripts/get_assets
+++ b/scripts/get_assets
@@ -1,0 +1,103 @@
+#!/usr/bin/env python
+#    Copyright 2013-2016 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""
+This module downloads external dependencies of assets used by Workload Automation.
+Works by first downloading a directory index of the assets, then iterating through
+it to get assets for the specified workloads.
+"""
+import sys
+import argparse
+import logging
+
+from wlauto import Workload
+from wlauto.common.resources import File
+from wlauto.resource_getters.standard import HttpGetter
+from wlauto.utils.log import init_logging
+
+__version__ = '0.1.0'
+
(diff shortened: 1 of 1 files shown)


setrofim:
> WA's HttpGetter class downloads assets to ~/.workload_automation/dependencies/__remote/<workload>, while workloads seem to check for dependencies in ~/.workload_automation/dependencies/<workload> or their own working directory.

Both locations are checked by different getters at different stages. The latter location takes priority over the former. This is done this way so that `HttpGetter` doesn't override assets that have been provided manually.

scripts/get_assets
```diff
@@ -0,0 +1,103 @@
...
+
+from wlauto import Workload
+from wlauto.common.resources import File
+from wlauto.resource_getters.standard import HttpGetter
+from wlauto.utils.log import init_logging
+
+__version__ = '0.1.0'
+
+# Get index.json from this URL
+ASSETS_REPO = 'https://github.com/jimboatarm/wa-assets/raw/master/workloads'
```
setrofim: Assets should be hosted on ARM-Software rather than a personal account.

It should also be settable in configuration rather than hard-coded.

setrofim:
This should be implemented as a command rather than a script, as this concerns core WA functionality.

scripts/get_assets
muendelezaji: Agreed, it will be configurable.
It's only like this because at the moment we need a way to change/test the folder structure without asking for commit access to ARM-software.

setrofim: Then it's too early to merge this into mainline.

muendelezaji:
> This should be implemented as a command rather than a script, as this concerns core WA functionality.

The reasoning behind having it as a script is so that someone (`pip`?) could run this as part of the initial setup to get all external workload assets and not need to do it on-demand. For example, if the WA installation will later be run on a device not connected to the internet.

Alternatively, it could also be automatically run as a post-install script, though I'm not familiar whether Python packaging tools (`pip`/`setuptools`) support that.

setrofim:
The fetching of all assets should not be done as part of installation, or at all in any kind of automatic way. WA will already automatically fetch assets on as-needed basis -- that is what `HttpGetter` is for. That way only the assets that that user needs will be fetched, transparently to the user.

There may be situations when when the user would want to fetch all or a subset of the assets in advance (e.g. when planning to be using WA without access to the Internet), so the ability to pre-fetch assets is usefully, but it needs to be an explicit decision taken by the user and not the default mode of operation.

PR #207, opened 2016-07-21
Revent fixes

ep1cman:


wlauto/external/revent/revent.c
```diff
@@ -251,7 +251,9 @@ void dump(const char *logfile)
     //Read magic
     len = strlen(magic);
     size_t rb = read(fdin, &buf[0], len);
+    buf[len] = '\0';
     if (rb != len) die("problems reading eventlog\n");
+    printf("magic: %s\n const: %s", buf, magic);
```
setrofim: Is this meant to be committed? This looks like a debug print

PR #208, opened 2016-07-21
Implemented visual state detection functionality for revent workloads

drcef:
- Added statedetect.py in utils which is a standalone module that
    contains all the methods needed for state detection

  - Modified the setup() and run() methods of the GameWorkload class
    in common/android/workload.py to have a parameter that enables
    state checks and run the check after setup and run if requested.

  - Modified angrybirds workload to enable state detection for
    testing purposes

State detection uses the template matching method available in
OpenCV to determine the state of the workload by detecting
predefined unique elements on a screenshot from the device.

### DIFF
--- a/wlauto/common/android/workload.py
+++ b/wlauto/common/android/workload.py
@@ -16,7 +16,6 @@
 import os
 import sys
 import time
-from math import ceil
 
 from wlauto.core.extension import Parameter
 from wlauto.core.workload import Workload
@@ -26,8 +25,10 @@ from wlauto.common.resources import ExtensionAsset, Executable
 from wlauto.exceptions import WorkloadError, ResourceError, ConfigError
 from wlauto.utils.android import ApkInfo, ANDROID_NORMAL_PERMISSIONS
 from wlauto.utils.types import boolean
-from wlauto.utils.revent import ReventParser
 import wlauto.common.android.resources
+from wlauto import File
+
+import wlauto.utils.statedetect as stateDetector
 
 
 DELAY = 5
@@ -324,13 +325,16 @@ AndroidBenchmark = ApkWorkload  # backward compatibility
 
 class ReventWorkload(Workload):
 
+    default_setup_timeout = 5 * 60  # in seconds
+    default_run_timeout = 10 * 60  # in seconds
+
     def __init__(self, device, _call_super=True, **kwargs):
         if _call_super:
             super(ReventWorkload, self).__init__(device, **kwargs)
         devpath = self.device.path
         self.on_device_revent_binary = devpath.join(self.device.binaries_directory, 'revent')
-        self.setup_timeout = kwargs.get('setup_timeout', None)
-        self.run_timeout = kwargs.get('run_timeout', None)
+        self.setup_timeout = kwargs.get('setup_timeout', self.default_setup_timeout)
+        self.run_timeout = kwargs.get('run_timeout', self.default_run_timeout)
         self.revent_setup_file = None
         self.revent_run_file = None
         self.on_device_setup_revent = None
@@ -345,10 +349,6 @@ class ReventWorkload(Workload):
         self.on_device_run_revent = devpath.join(self.device.working_directory,
                                                  os.path.split(self.revent_run_file)[-1])
         self._check_revent_files(context)
-        default_setup_timeout = ceil(ReventParser.get_revent_duration(self.revent_setup_file)) + 30
-        default_run_timeout = ceil(ReventParser.get_revent_duration(self.revent_run_file)) + 30
-        self.setup_timeout = self.setup_timeout or default_setup_timeout
-        self.run_timeout = self.run_timeout or default_run_timeout
 
     def setup(self, context):
         self.device.killall('revent')
@@ -447,9 +447,12 @@ class GameWorkload(ApkWorkload, ReventWorkload):
     view = 'SurfaceView'
     loading_time = 10
     supported_platforms = ['android']
+    check_game_states = None
 
     parameters = [
         Parameter('install_timeout', default=500, override=True),
+	Parameter('check_states', kind=bool, default=False, global_alias='check_game_states',
+		  description='Use visual state detection to verify the state of the workload after setup and run'),
         Parameter('assets_push_timeout', kind=int, default=500,
                   description='Timeout used during deployment of the assets package (if there is one).'),
         Parameter('clear_data_on_reset', kind=bool, default=True,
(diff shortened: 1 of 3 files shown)


wlauto/common/android/workload.py
```diff
@@ -16,7 +16,6 @@
 import os
 import sys
 import time
-from math import ceil
```
ep1cman: Looks like you might need to rebase your commits

wlauto/common/android/workload.py
```diff
@@ -476,6 +479,19 @@ def setup(self, context):
         time.sleep(self.loading_time)
         ReventWorkload.setup(self, context)
 
+	# state detection check if it's enabled in the config
+	if self.check_game_states:
+		try:
+			self.logger.info("\tChecking workload state...")
+			statedefs_dir = context.resolver.get(File(self, 'state_definitions'))
+			self.device.capture_screen(context.output_directory+"/aftersetup.png")
+			stateCheck = stateDetector.verify_state(context.output_directory+"/aftersetup.png", statedefs_dir, "setupComplete")
```
ep1cman: Split this line up, its too long

wlauto/common/android/workload.py
```diff
@@ -476,6 +479,19 @@ def setup(self, context):
...
+			if not stateCheck: raise WorkloadError("Unexpected state after setup")
+		except ResourceError:
+			self.logger.warning("State definitions directory not found. Skipping state detection.")
+		except stateDetector.StateDefinitionError, errorMsg:
+			self.logger.warning("State definitions or template files missing or invalid (" + errorMsg + "). Skipping state detection.")
```
ep1cman: Split this line up, its too long

setrofim:
`check_state`

wlauto/utils/statedetect.py
```diff
@@ -0,0 +1,120 @@
...
+State detection functionality for revent workloads. Uses OpenCV to analyse screenshots from the device.
+Requires a 'statedetection' directory in the workload directory that includes the state definition yaml file,
+and the 'templates' folder with PNGs of all templates mentioned in the yaml file.
+
+Requires the following python plugins:
+numpy, pyyaml (yaml), imutils and opencv (cv2)
+
+"""
+
+import cv2
```
setrofim: third-party imports should come after standard library imports, separated by a line

wlauto/utils/statedetect.py
```diff
@@ -0,0 +1,120 @@
...
+# limitations under the License.
+#
+
+
+"""
+State detection functionality for revent workloads. Uses OpenCV to analyse screenshots from the device.
+Requires a 'statedetection' directory in the workload directory that includes the state definition yaml file,
+and the 'templates' folder with PNGs of all templates mentioned in the yaml file.
+
+Requires the following python plugins:
```
setrofim: python --> Python

setrofim: plugins --> libraries

wlauto/utils/statedetect.py
```diff
@@ -0,0 +1,120 @@
...
+numpy, pyyaml (yaml), imutils and opencv (cv2)
```
setrofim: Please add the new dependencies to setup.py.

Also, since some of these are quite "heavy" (in particular we do *not* want to depend on numpy by default), these should be added as optional dependencies in their own group. Which means that here, you cannot assume that the imports will succeed. You need to guard against that and report and error only if the user tried to use the optional functionality without those modules installed. 

For example, see `fps` instrument which uses the optional `pandas` dependency:

https://github.com/ARM-software/workload-automation/blob/master/wlauto/instrumentation/fps/__init__.py

wlauto/utils/statedetect.py
```diff
@@ -0,0 +1,120 @@
...
+	lower = int(max(0, (1.0 - sigma) * v))
+	upper = int(min(255, (1.0 + sigma) * v))
+	edged = cv2.Canny(image, lower, upper)
+ 
+	# return the edged image
+	return edged
+
+def match_state(screenshotFile, defpath):
+	# load and parse state definition file
+	if not os.path.isfile(defpath+'/definition.yaml'): raise StateDefinitionError("Missing state definitions yaml file")
```
setrofim: the `raise` should be on it's own line. Do not put the suite of the if statement onthe same lime

wlauto/utils/statedetect.py
```diff
@@ -0,0 +1,120 @@
...
+	stateDefinitions = yaml.load(file(defpath+'/definition.yaml', 'r'))
```
setrofim: should be

    with open(os.path.join(defpath, 'definition.yaml')) as fh:
            state_definitions = yaml.load(fh)

wlauto/utils/statedetect.py
setrofim: `os.path.join` instead of '+' -- do not assume that '/' is the path separator

setrofim: The missing path should be reported as part of the error message

setrofim: since the definition path is used in multiple places below, it's probably worth saving it in a variable

setrofim:
Do not use camel case for variable names in Python -- local variables should be lower case separated by '_' as appropriate. Please also see

https://pythonhosted.org/wlauto/contributing.html

wlauto/utils/statedetect.py
```diff
@@ -0,0 +1,120 @@
...
+import cv2
+import numpy as np
+import imutils
+import yaml
+import os
+
+class StateDefinitionError(RuntimeError):
+	def __init__(self, arg):
+		self.args = arg
+
```
setrofim: Two lines between definitions at top level

wlauto/common/android/workload.py
```diff
@@ -476,6 +479,19 @@ def setup(self, context):
         time.sleep(self.loading_time)
         ReventWorkload.setup(self, context)
 
+	# state detection check if it's enabled in the config
+	if self.check_game_states:
```
setrofim: use `self.check_states` here -- that's the parameter name

wlauto/common/android/workload.py
```diff
@@ -447,9 +447,12 @@ class GameWorkload(ApkWorkload, ReventWorkload):
     view = 'SurfaceView'
     loading_time = 10
     supported_platforms = ['android']
+    check_game_states = None
```
setrofim: Get rid of this -- use `self.check_states` instead

wlauto/workloads/angrybirds/__init__.py
```diff
@@ -27,4 +27,4 @@ class AngryBirds(GameWorkload):
     """
     package = 'com.rovio.angrybirds'
     activity = 'com.rovio.ka3d.App'
-
+    check_game_states = True
```
setrofim: This is not necessary

setrofim:
Closing as this is superseded by https://github.com/ARM-software/workload-automation/pull/210

PR #210, opened 2016-07-22
Implemented visual state detection functionality for revent workloads

drcef:
- Added statedetect.py in utils which is a standalone module that
    contains all the methods needed for state detection

  - Modified the setup() and run() methods of the GameWorkload class
    in common/android/workload.py to have a parameter that enables
    state checks and run the check after setup and run if requested.

State detection uses the template matching method available in
OpenCV to determine the state of the workload by detecting
predefined unique elements on a screenshot from the device.

### DIFF
--- a/doc/source/revent.rst
+++ b/doc/source/revent.rst
@@ -210,3 +210,100 @@ https://www.kernel.org/doc/Documentation/input/event-codes.txt
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     |                          Event Value                          |
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
+
+Using state detection with revent
+=================================
+
+State detection can be used to verify that a workload is executing as expected.
+This utility, if enabled, and if state definitions are available for the
+particular workload, takes a screenshot after the setup and the run revent
+sequence, matches the screenshot to a state and compares with the expected
+state. A WorkloadError is raised if an unexpected state is encountered.
+
+To enable state detection, make sure a valid state definition file and
+templates exist for your workload and set the check_states parameter to True.
+
+State definition directory
+--------------------------
+
+State and phase definitions should be placed in a directory of the following
+structure inside the dependencies directory of each workload (along with
+revent files etc):
+
+   dependencies/
+      <workload_name>/
+         state_definitions/
+            definition.yaml
+            templates/
+               <oneTemplate>.png
+               <anotherTemplate>.png
+               ...
+
+definition.yaml file
+--------------------
+
+This defines each state of the workload and lists which templates are expected
+to be found and how many are required to be detected for a conclusive match. It
+also defines the expected state in each workload phase where a state detection
+is run (currently those are setupComplete and runComplete).
+
+Templates are picture elements to be matched in a screenshot. Each template
+mentioned in the definition file should be placed as a file with the same name
+and a .png extension inside the templates folder. Creating template png files
+is as simple as taking a screenshot of the workload in a given state, cropping
+out the relevant templates (eg. a button, label or other unique element that is
+present in that state) and storing them in PNG format.
+
+Please see the definition file for Angry Birds below as an example to
+understand the format. Note that more than just two states (for the afterSetup
+and afterRun phase) can be defined and this helps track the cause of errors in
+case an unexpected state is encountered.
+
+.. code-block:: python
+
+    workload_name: angrybirds
+
+    workload_states:
+      - state_name: titleScreen
+        templates:
+          - play_button
+          - logo
+        matches: 2
+      - state_name: worldSelection
+        templates:
+          - first_world_thumb
+          - second_world_thumb
+          - third_world_thumb
+          - fourth_world_thumb
+        matches: 3
+      - state_name: level_selection
+        templates:
+          - locked_level
+          - first_level
+        matches: 2
+      - state_name: gameplay
+        templates:
+          - pause_button
+          - score_label_text
+        matches: 2
+      - state_name: pause_screen
+        templates:
+          - replay_button
+          - menu_button
+          - resume_button
+          - help_button
+        matches: 4
+      - state_name: level_cleared_screen
+        templates:
+          - level_cleared_text
+          - menu_button
+          - replay_button
+          - fast_forward_button
+        matches: 4
+
+    workload_phases:
+      - phase_name: setup_complete
+        expected_state: gameplay
+      - phase_name: run_complete
+        expected_state: level_cleared_screen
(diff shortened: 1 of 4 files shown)


wlauto/common/android/workload.py
```diff
@@ -476,6 +481,22 @@ def setup(self, context):
         time.sleep(self.loading_time)
         ReventWorkload.setup(self, context)
 
+        # state detection check if it's enabled in the config
+        if self.check_states:
+            try:
+                self.logger.info("\tChecking workload state...")
+                statedefs_dir = context.resolver.get(File(self, 'state_definitions'))
+                self.device.capture_screen(os.path.join(context.output_directory, "aftersetup.png"))
```
ep1cman: you are using this path twice, might be worth saving it to a variable

wlauto/common/android/workload.py
```diff
@@ -494,6 +515,22 @@ def reset(self, context):
     def run(self, context):
         ReventWorkload.run(self, context)
 
+        # state detection check if it's enabled in the config
```
ep1cman: This whole section is basically a duplicate of the above, split it out into a function

ep1cman: Since GameWorkload sub classes ReventWorkload it should be a method of ReventWorkload

wlauto/common/android/workload.py
```diff
@@ -476,6 +481,22 @@ def setup(self, context):
...
+        if self.check_states:
+            try:
+                self.logger.info("\tChecking workload state...")
+                statedefs_dir = context.resolver.get(File(self, 'state_definitions'))
+                self.device.capture_screen(os.path.join(context.output_directory, "aftersetup.png"))
+                stateCheck = stateDetector.verify_state(os.path.join(context.output_directory,
+                    "aftersetup.png"), statedefs_dir, "setupComplete")
+                if not stateCheck:
+                    raise WorkloadError("Unexpected state after setup")
+            except ResourceError:
```
ep1cman: These checks should be happening as part of the init_resources method. That way if the assets cannot be found or are corrupt the workload will not even attempt to run. This will present itself to the end user much sooner than waiting till the workload setup.

ep1cman:
Can you also update the revent documentation explaining this new functionality, and what the format of the definitions file is (including examples) because as it stands right now I would not know how to use this feature.

doc/source/revent.rst
```diff
@@ -210,3 +210,99 @@ https://www.kernel.org/doc/Documentation/input/event-codes.txt
...
+        matches: 4
+      - stateName: levelClearedScreen
+        templates:
+          - levelClearedText
+          - menuButton
+          - replayButton
+          - fastForwardButton
+        matches: 4
+
+    workloadPhases:
```
ep1cman: I notice below you are only using the `gameplay` and `levelClearScreen` states, why are there more above?

wlauto/common/android/workload.py
```diff
@@ -494,6 +524,10 @@ def reset(self, context):
     def run(self, context):
         ReventWorkload.run(self, context)
 
+        # state detection check if it's enabled in the config
+        if self.check_states:
```
ep1cman: I would move this check to the teardown because this would effect what instrumentation measures for the apps "run"

doc/source/revent.rst
```diff
@@ -210,3 +210,99 @@ https://www.kernel.org/doc/Documentation/input/event-codes.txt
...
+present in that state) and storing them in PNG format.
+
+Please see the definition file for Angry Birds below as an example to
+understand the format:
+
+.. code-block:: python
+
+    workloadName: angrybirds
+
+    workloadStates:
```
setrofim: Unless there is a very good reason to use camel case here, the format should follow all-lower-case, words-separated-by-underscores naming to be consistent with the rest of WA configuration.

setrofim:
Please keep the commit history tidy -- if you're pushing fixes to the functionality new to this pull request, they should be merged into the existing commit(s) rather than pushed as new ones -- each commit should represent a clear logical change.

doc/source/revent.rst
drcef: While only those two states are required, more can be defined so that the current state of the workload can be indicated even in a mismatch (for example, "Unexpected state: pauseScreen instead of levelClearScreen"). I defined most of those states for demo purposes in Angry Birds.

ep1cman: In which case add a bit to the documentation explaining this feature

wlauto/common/android/workload.py
```diff
@@ -469,13 +492,19 @@ def __init__(self, device, **kwargs):  # pylint: disable=W0613
     def init_resources(self, context):
         ApkWorkload.init_resources(self, context)
         ReventWorkload.init_resources(self, context)
+        if self.check_states:        
+            ReventWorkload._check_statedetection_files(self, context)
```
ep1cman: Any reason you are calling from the class rather than `self`

wlauto/utils/statedetect.py
```diff
@@ -0,0 +1,147 @@
...
+    # make a list of all templates defined in the state definitions
+    template_list = []
+    for state in state_definitions["workload_states"]:
+        template_list.extend(state["templates"])
+
+    # check all template PNGs exist
+    missing_files = 0
+    for template_png in template_list:
+        if not os.path.isfile(os.path.join(defpath, 'templates', template_png+'.png')):
+            missing_files += 1
```
ep1cman: rather than counting the missing files then raising an exception later why not raise it here, you can even specify which file is missing

wlauto/utils/statedetect.py
```diff
@@ -0,0 +1,147 @@
...
+
+    return matched_state
+
+
+def verify_state(screenshot_file, stateDefsPath, workloadPhase):
+    # run a match on the screenshot
+    matched_state = match_state(screenshot_file, stateDefsPath)
+
+    # load and parse state definition file
+    statedefs_file = os.path.join(stateDefsPath, 'definition.yaml')
```
ep1cman: Rather than parsing the yaml file both in `match_state` and here, just parse it here and pass the parsed yaml to `match_state`

wlauto/utils/statedetect.py
drcef: Good point, this was left from when the script was interactive

doc/source/revent.rst
```diff
@@ -210,3 +210,100 @@ https://www.kernel.org/doc/Documentation/input/event-codes.txt
...
+Templates are picture elements to be matched in a screenshot. Each template
+mentioned in the definition file should be placed as a file with the same name
+and a .png extension inside the templates folder. Creating template png files
+is as simple as taking a screenshot of the workload in a given state, cropping
+out the relevant templates (eg. a button, label or other unique element that is
+present in that state) and storing them in PNG format.
+
+Please see the definition file for Angry Birds below as an example to
+understand the format. Note that more than just two states (for the afterSetup
+and afterRun phase) can be defined and this helps track the cause of error in
```
ep1cman: cause of error > cause of errors

drcef: Corrected

PR #212, opened 2016-07-25
Additional changes to FpsProcessor, fps instrument and uxperf result processor

johnrchrdsn:
Includes :
- Fix for FpsProcessor logic when data for filtered_vsyncs_to_compose is empty
- Adds dumpsys_period parameter for controlling the frequency of dumpsys in the fps instrument
- New logging calls and exception handling in uxperf result processor.

### DIFF
--- a/wlauto/result_processors/uxperf.py
+++ b/wlauto/result_processors/uxperf.py
@@ -15,6 +15,7 @@
 
 import os
 import re
+import logging
 
 from collections import defaultdict
 from distutils.version import LooseVersion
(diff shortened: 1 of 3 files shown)


wlauto/utils/fps.py
```diff
@@ -19,8 +19,6 @@ class FpsProcessor(object):
     Provide common object for processing surfaceFlinger output for frame
     statistics.
 
-    This processor adds four metrics to the results:
```
ep1cman: Why was this line removed?

johnrchrdsn: Originally the code added the metrics to the context but that has now changed. This comment is therefore a bit misleading.

ep1cman: Rather than removing it I would change it to say that it returns the values below

wlauto/result_processors/uxperf.py
```diff
@@ -86,12 +87,15 @@ def export_iteration_result(self, result, context):
         logfile = os.path.join(context.output_directory, 'logcat.log')
         framelog = os.path.join(context.output_directory, 'frames.csv')
 
+        self.logger.info('Parsing logcat.log for UX_PERF markers')
```
ep1cman: This should also go to the debug log

PR #213, opened 2016-07-26
Extensions to baselib for UX Performance

johnrchrdsn:
Includes:
- Minor code changes for consistent java formatting
- Further convenience methods for use in new workloads
- Upgrades Android API level dependency from 17 to 18. Build properties in affected workloads updated.


ep1cman:
As per the previous pull request, can you please make the API bump its own commit to make it clear that this has happened. Also, have you tried rebuilding all the affected workloads to make sure they still build successfully?

johnrchrdsn:
The API bump is now a separate commit. I've also rebuilt all the affected workloads (those that extend from BaseUiAutomation) and they all build successfully.

PR #214, opened 2016-07-26
Add network connectivity check for devices

johnrchrdsn:
Add a general check that pings a network host rather than checking the
status of dumpsys wifi due to the fact that not all devices are
connected via wifi. Intended for workloads that require an internet
connection to operate.

### DIFF
--- a/wlauto/common/linux/device.py
+++ b/wlauto/common/linux/device.py
@@ -318,6 +318,26 @@ class BaseLinuxDevice(Device):  # pylint: disable=abstract-method
 
             please see: https://pythonhosted.org/wlauto/writing_extensions.html""")
 
+    def is_network_connected(self, ip_address='8.8.8.8'):
+        """
+        Checks for internet connectivity on the device
+        by pinging IP address provided.
+
+        :param ip_address: the IP address to ping
+                           (default - google-public-dns-a.google.com)
+
+        """
+        self.logger.debug('Checking for internet connectivity.')
+        output = self.execute('ping -q -w 1 -c 1 {}'.format(ip_address),
+                              check_exit_code=False)
+
+        if 'network is unreachable' not in output.lower():
+            self.logger.debug('Found IP address {}'.format(ip_address))
+            return True
+        else:
+            self.logger.debug('Cannot find IP address {}'.format(ip_address))
+            return False
+
     def get_binary_path(self, name, search_system_binaries=True):
         """
         Searches the devices ``binary_directory`` for the given binary,


wlauto/common/android/device.py
```diff
@@ -670,6 +670,19 @@ def get_device_model(self):
         except KeyError:
             return None
 
+    def is_network_connected(self):
+            network_host = 'www.google.com'
```
setrofim: `network_host` should probably be a parameter defaulting to `'www.google.com'` (also -- is there not a dedicated test domain for these things)?

wlauto/common/android/device.py
```diff
@@ -670,6 +670,19 @@ def get_device_model(self):
...
+            self.logger.debug('Checking for internet connectivity.')
+            output = adb_shell(self.adb_name, 'ping -q -w 1 -c 1 {}'.format(network_host),
```
setrofim: Is `ping` part of standard Android build? Will it be present on all targets?

wlauto/common/android/device.py
johnrchrdsn: As for a dedicated test domain there isn't one that I know off. Most of the examples I've seen use google as a default.

wlauto/common/android/device.py
johnrchrdsn: I believe this is present on all android devices and is part of the standard binaries included. It is part of busybox anyway (which I believe is a requirement for WA?) so it should be in the path anyway.

setrofim: On 26 Jul 2016, 02:16:30 -0700, johnrchrdsn wrote:
> > @@ -670,6 +670,19 @@ def get_device_model(self):
> >          except KeyError:
> >              return None
> >
> > +    def is_network_connected(self):
> > +            network_host = 'www.google.com'
> > +            self.logger.debug('Checking for internet connectivity.')
> > +            output = adb_shell(self.adb_name, 'ping -q -w 1 -c 1 {}'.format(network_host),
>
> I believe this is present on all android devices and is part of the standard binaries included. It is part of busybox anyway (which I believe is a requirement for WA?) so it should be in the path anyway.

busybox is not a requirement of WA -- it's in WA and will be deployed by
WA itself, however it won't be in PATH. That's where I was headed with
this -- if you're not sure it will be on all devices it would safer to
invoke `ping` via busybox explicitly:

         output = adb_shell(self.adb_name, '{} ping -q -w 1 -c 1 {}'.format(self.busybox, network_host),

>
> ---
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/ARM-software/workload-automation/pull/214/files/2c199966e414d33bea8eb8a0c758a716c732039b#r72214248
IMPORTANT NOTICE: The contents of this email and any attachments are confidential and may also be privileged. If you are not the intended recipient, please notify the sender immediately and do not disclose the contents to any other person, use it for any purpose, or store or copy the information in any medium. Thank you.

wlauto/common/android/device.py
```diff
@@ -670,6 +670,18 @@ def get_device_model(self):
         except KeyError:
             return None
 
+    def is_network_connected(self, network_host='www.google.com'):
```
setrofim: Just spotted -- the method body is over-indented; each indentation level must be exactly four spaces; have you run the pep8/pylint checkers on this?

johnrchrdsn: My apologies that was as copy and paste error! Should be correct now. Ran the pylint script and that file is clean.

wlauto/common/android/device.py
```diff
@@ -670,6 +670,18 @@ def get_device_model(self):
...
+        self.logger.debug('Checking for internet connectivity.')
+        output = adb_shell(self.adb_name, 'ping -q -w 1 -c 1 {}'.format(network_host),
```
ep1cman: Why are you using adb_shell rather than self.execute?

johnrchrdsn: Changed to use self.execute instead which has a self._check_ready() call rather than adb_shell().

wlauto/common/android/device.py
```diff
@@ -670,6 +670,18 @@ def get_device_model(self):
         except KeyError:
             return None
 
+    def is_network_connected(self, network_host='www.google.com'):
+        self.logger.debug('Checking for internet connectivity.')
+        output = self.execute('ping -q -w 1 -c 1 {}'.format(network_host),
```
ep1cman: As per sergei's previous comment this should be:

        output = self.execute('{} ping -q -w 1 -c 1 {}'.format(self.busybox, network_host),

johnrchrdsn: The ping for busybox requires a rooted device whereas the built in does not. The targets we have tested so far all have a ping command built in and I would expect this to be included as standard for android devices. If this is still not acceptable than let me know.

ep1cman:
Any reason this was added to AndroidDevice and not BaseLinuxDevice?

johnrchrdsn:
No particular reason other than I am testing on android devices. I will move to the BaseLinuxDevice class and re-test.

PR #216, opened 2016-07-27
Add UxPerfUiAutomation and googleplaybooks workload

johnrchrdsn:


wlauto/workloads/googleplaybooks/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+# limitations under the License.
+#
+
+import os
+import re
+
+from wlauto import AndroidUiAutoBenchmark, Parameter
+from wlauto.exceptions import DeviceError
+
+__version__ = '1.0.0'
```
ep1cman: Why is this here

wlauto/workloads/googleplaybooks/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+    def update_result(self, context):
+        super(Googleplaybooks, self).update_result(context)
+
+        # Add classifiers for workload specific input parameters
+        d = {}
+        for param in self.parameters:
+            if param.name not in getattr(AndroidUiAutoBenchmark, "parameters"):
+                d[param.name] = getattr(self, param.name)
+
+        context.add_classifiers(**d)
```
johnrchrdsn: This logic will be common to a number of uxperf workloads if we want to capture the classifiers. Are you happy for it to be duplicated when required, or should this be moved to the base class? The latter approach would add classifiers to non-uxperf workloads too.

setrofim: This logic is too generic. We want to keep the classifiers as low-noise as possible, so the decision to add a param  as a classifier should be made explicitly. One should not assume that "every param that's not in super should be added".

In general, params should not be  added at all, as they are already recorded in the results meta data (and if the user decides they want them in the main results for a particular set of runs, they can always add them via the agenda).

Typically, classifiers should only be automatically added for things that aren't already being captured, e.g. the version of the app, as read from the device.

johnrchrdsn: Considering that we expect the APK version to be fixed I will remove the update_result method completely and remove the classifiers.

wlauto/workloads/googleplaybooks/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+     6. Opens selected  book
+     7. Gestures are performed to swipe between pages and pinch zoom in and out of a page
+     8. Selects a specified chapter based on page number from the navigation view
+     9. Selects a word in the centre of screen and adds a test note to the page
+    10. Removes the test note from the page (clean up)
+    11. Searches for the number of occurrences of a common word throughout the book
+    12. Switches page styles from 'Day' to 'Night' to 'Sepia' and back to 'Day'
+    13. Uses the 'About this book' facility on the currently selected book
+
+    NOTE: This workload requires a network connection (ideally, wifi) to run.
```
ep1cman: It also requires a google account to be setup on the device

wlauto/workloads/googleplaybooks/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    parameters = [
+        Parameter('search_book_title', kind=str, mandatory=False, default="Hamlet",
+                  description="""
+                  The book title to search for within Google Play Books archive.
+                  The book must either be already in the account's library, or free to purchase.
+                  Note: spaces must be replaced with underscores in the book title.
```
ep1cman: why not just do this in the code?

johnrchrdsn: When UiAutomator is invoked on the command line the arguments are separated by spaces. Hence to make this one argument we need the underscores.

ep1cman: so when passing that string why not do 
```
self.uiauto_params['book_title'] = self.search_book_title.replace(" ", "_")
```

rather than requiring the user to do it themselves

johnrchrdsn: You are right. Changed accordingly.

wlauto/utils/android.py
```diff
@@ -100,6 +100,15 @@
     'UNINSTALL_SHORTCUT',
 ]
 
+GOOGLE_KEYBOARD_PACKAGE = 'com.android.inputmethod.latin'
+
+# List of known Google Keyboard versions that have problems
+# when key presses are used in AndroidUiAutoBenchmark workloads
+GOOGLE_KEYBOARD_BLACKLIST = [
```
ep1cman: I would make this a dict with the key being the package and the values being lists of bad versions

wlauto/utils/android.py
```diff
@@ -100,6 +100,15 @@
...
+    '5.0.25.122319759-armeabi-v7a',
```
ep1cman: we should probably only go down to major.minor.path with the version numbers, things like the ABI at the end will change but it will still have the same bug

wlauto/common/android/workload.py
```diff
@@ -424,6 +429,21 @@ def teardown(self, context):
...
 
+    def _check_unsupported_packages(self):
+        """
+        Check for any unsupported package versions and raise an
+        exception if detected.
+
+        """
+        for package in UNSUPPORTED_PACKAGES:
+            version = self.device.get_installed_package_version(package)
+            if '-' in version:
```
ep1cman: This errors if version is `None`

```
2016-08-09 09:38:42,113 DEBUG    android: adb  shell "dumpsys package com.google.android.inputmethod.latin"
2016-08-09 09:38:42,175 CRITICAL root:   File "/work/home/wa2/workload-automation/wlauto/core/entry_point.py", line 66, in main
2016-08-09 09:38:42,175 CRITICAL root:     sys.exit(command.execute(args))
2016-08-09 09:38:42,175 CRITICAL root:   File "/work/home/wa2/workload-automation/wlauto/commands/run.py", line 104, in execute
2016-08-09 09:38:42,175 CRITICAL root:     executor.execute(agenda, selectors={'ids': args.only_run_ids})
2016-08-09 09:38:42,175 CRITICAL root:   File "/work/home/wa2/workload-automation/wlauto/core/execution.py", line 346, in execute
2016-08-09 09:38:42,175 CRITICAL root:     runner.run()
2016-08-09 09:38:42,175 CRITICAL root:   File "/work/home/wa2/workload-automation/wlauto/core/execution.py", line 484, in run
2016-08-09 09:38:42,175 CRITICAL root:     self._initialize_run()
2016-08-09 09:38:42,175 CRITICAL root:   File "/work/home/wa2/workload-automation/wlauto/core/execution.py", line 543, in _initialize_run
2016-08-09 09:38:42,175 CRITICAL root:     workload_spec.workload.initialize(self.context)
2016-08-09 09:38:42,175 CRITICAL root:   File "/work/home/wa2/workload-automation/wlauto/core/extension.py", line 486, in wrapper
2016-08-09 09:38:42,175 CRITICAL root:     dm(self, *args, **kwargs)
2016-08-09 09:38:42,175 CRITICAL root:   File "/work/home/wa2/workload-automation/wlauto/core/extension.py", line 486, in wrapper
2016-08-09 09:38:42,175 CRITICAL root:     dm(self, *args, **kwargs)
2016-08-09 09:38:42,175 CRITICAL root:   File "/work/home/wa2/workload-automation/wlauto/common/android/workload.py", line 414, in initialize
2016-08-09 09:38:42,175 CRITICAL root:     self._check_unsupported_packages()
2016-08-09 09:38:42,175 CRITICAL root:   File "/work/home/wa2/workload-automation/wlauto/common/android/workload.py", line 440, in _check_unsupported_packages
2016-08-09 09:38:42,175 CRITICAL root:     if '-' in version:
2016-08-09 09:38:42,175 CRITICAL root: 
2016-08-09 09:38:42,176 CRITICAL root: TypeError(argument of type 'NoneType' is not iterable)
```

johnrchrdsn: Yep my mistake. Added a conditional for the None condition.

PR #217, opened 2016-08-02
Dump hierarchy view on error

johnrchrdsn:
Dump window hierarchy view from uiautomator to a file when WA fails
during execution. Note: the xml file are pre-formatted after dump.
Implementation specific to android.device.

### DIFF
--- a/wlauto/core/execution.py
+++ b/wlauto/core/execution.py
@@ -731,6 +731,13 @@ class Runner(object):
             filepath = os.path.join(settings.output_directory, filename)
         self.device.capture_screen(filepath)
 
+    def _take_uiautomator_dump(self, filename):
+        if self.context.output_directory:
+            filepath = os.path.join(self.context.output_directory, filename)
+        else:
+            filepath = os.path.join(settings.output_directory, filename)
+        self.device.capture_view_hierachy(filepath)
+
     @contextmanager
     def _handle_errors(self, action, on_error_status=IterationResult.FAILED):
         try:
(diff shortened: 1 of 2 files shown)


wlauto/common/linux/device.py
```diff
@@ -897,6 +897,9 @@ def capture_screen(self, filepath):
             message = e.message.split('OUTPUT:', 1)[1].strip()
             self.logger.debug('Could not take screenshot: {}'.format(message))
 
+    def capture_view_hierachy(self, filepath):
```
ep1cman: this only makes sense for android devices, not linux devices

wlauto/common/gem5/device.py
```diff
@@ -494,6 +494,9 @@ def capture_screen(self, filepath):
                 pass
         return False
 
+    def capture_view_hierachy(self, filepath):
```
ep1cman: this only makes sense for android devices, not linux devices

wlauto/core/device.py
```diff
@@ -318,6 +318,10 @@ def capture_screen(self, filepath):
         """Captures the current device screen into the specified file in a PNG format."""
         raise NotImplementedError()
 
+    def capture_view_hierachy(self, filepath):
```
ep1cman: this only makes sense for android devices, should not be in the device base class

wlauto/core/execution.py
```diff
@@ -746,6 +753,7 @@ def _handle_errors(self, action, on_error_status=IterationResult.FAILED):
                 self.current_job.result.add_event(str(we))
             try:
                 self._take_screenshot('error.png')
+                self._take_uiautomator_dump('error.xml')
```
ep1cman: you now need to check self.device.platform is android

PR #221, opened 2016-08-18
workloads: added octane2

mwasilew:
Initial support for Octane2 JS benchmark. The workload is designed for
Chromium browser as default AOSP browser crashes when trying to load the
page.

Signed-off-by: Milosz Wasilewski milosz.wasilewski@linaro.org


wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+                  description='Specifies the timeout (in seconds) before the uiautomator script is terminated.'),
+        Parameter('clear_file_cache', kind=boolean, default=True,
+                  description='Clear the the file cache on the target device prior to running the workload.'),
+    ]
+
+    supported_platforms = ['android']
+
+    def setup(self, context):  # NOQA
+        UiAutomatorWorkload.setup(self, context)
+        self.octane2_on_device = '/'.join([self.device.working_directory, self.name])
```
setrofim: Should be

    self.octane2_on_device = self.device.path.join(self.device.working_directory, self.name)

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+        self.index_octane = 'file:///{}'.format(self.octane2_on_device) + '/index.html'
```
setrofim: Too many `/` in `file:///`? won't the third `/` be part of the path?

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+
+        if not os.path.isdir(self.dependencies_directory):
```
setrofim: Should not assume that if the dependencies directory exists then octane has been downloaded. Should check either for the zip itself, or extracted contents.

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+            self._download_octane2_file()
+
+        # Push the octane2
+        if self.force_dependency_push or not self.device.file_exists(self.octane2_on_device):
+            self.logger.debug('Copying octane2 to device.')
+            self.device.push_file(self.dependencies_directory, self.octane2_on_device, timeout=300)
+
+        # Open the browser with default page
+        self.device.execute('am start -n  {}/{} about:blank'.format(self.browser_package, self.browser_activity))
+        time.sleep(5)
```
setrofim: Do not sleep here. Use `-W` option in  the `am` call to wait for launch to complete instead.

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # Stop the browser if already running and wait for it to stop
+        self.device.execute('am force-stop {}'.format(self.browser_package))
+        time.sleep(5)
```
setrofim: Why is this sleep necessary? Isn't `force-stop` blocking?

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+        #On android 6+ the web browser requires permissions to access the sd card
+        if self.device.get_sdk_version() >= 23:
+            self.device.execute("pm grant {} android.permission.READ_EXTERNAL_STORAGE".format(self.browser_package))
+            self.device.execute("pm grant {} android.permission.WRITE_EXTERNAL_STORAGE".format(self.browser_package))
+
+
+    def run(self, context):
+        # Launch browser
+        self.device.execute('am start -n  {}/{} {}'.format(self.browser_package, self.browser_activity, self.index_octane))
+        time.sleep(5)  # WA1 parity
```
setrofim: As above, use `-W` instead of sleeping

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+        # Launch automation script
+        command = 'uiautomator runtest {} -e workdir {} -c {}'.format(self.device.working_directory + self.octane_jar_file,
```
setrofim: As above, use `self.device.path.join` instead of `+`

setrofim:
Changes to `BaseUiAutomator` should be in a separate commit from the workload.

wlauto/workloads/octane2/__init__.py
mwasilew: Fixed.

wlauto/workloads/octane2/__init__.py
mwasilew: Good point. It actually doesn't show up in the browser (auto correction?) but can be spotted from the log. Fixed.

wlauto/workloads/octane2/__init__.py
mwasilew: I added a check for index.html. Zipfile is removed after extraction.

wlauto/workloads/octane2/__init__.py
mwasilew: This is a bit problematic with chromium as it takes a bit of time to display the T&Cs dialog. I had to add additional timeout in the uiautomator script instead.

wlauto/workloads/octane2/__init__.py
mwasilew: Fixed. This was a copy/paste error.

wlauto/workloads/octane2/__init__.py
mwasilew: Fixed.

wlauto/workloads/octane2/__init__.py
mwasilew: Fixed.

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    Octane 2 is a JavaScript benchmark for browser
+
+    """
+    octane_jar_file = 'org.linaro.wlauto.uiauto.octane2.jar'
+    octane_run_string = 'org.linaro.wlauto.uiauto.octane2.UiAutomation#runUiAutomation'
+    octane_extract_string = 'org.linaro.wlauto.uiauto.octane2.UiAutomation#runUiAutomation'
+
+    parameters = [
+        Parameter('force_dependency_push', kind=boolean, default=False,
+                  description=('Specifies whether to push dependency files to the device to the device '
```
ep1cman: `to the device` repeated twice

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+            os.path.join(self.device.working_directory, self.octane_jar_file),
+            self.device.working_directory,
+            self.octane_run_string)
+        self.device.execute(command, timeout=self.octane_timeout)
+
+    def update_result(self, context):
+        # Stop the browser
+        self.device.execute('am force-stop {}'.format(self.browser_package))
+        # extract results
+        command = 'uiautomator runtest {} -e workdir {} -c {}'.format(
```
ep1cman: This is not used anywhere

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+        z.extractall(self.dependencies_directory)
+
+        # move files from 'octane-master' one level up
+        octane_master = os.path.join(self.dependencies_directory, "octane-master")
+        for filename in os.listdir(octane_master):
+            shutil.move(os.path.join(octane_master, filename), self.dependencies_directory)
+        os.rmdir(octane_master)
+        os.unlink(full_file_path)
+
+def _parse_metrics(logfile, output_directory):  # pylint: disable=R0914
```
ep1cman: `output_directory` is unused

wlauto/workloads/octane2/__init__.py
mwasilew: Fixed.

wlauto/workloads/octane2/__init__.py
mwasilew: Correct, removed.

wlauto/workloads/octane2/__init__.py
mwasilew: Removed.

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
```diff
@@ -0,0 +1,93 @@
...
+    public void runUiAutomation() throws Exception {
+        final int timeout = 10;
+        Bundle status = new Bundle();
+        status.putString("product", getUiDevice().getProductName());
+
+        sleep(timeout);
+        UiSelector selector = new UiSelector();
+        /* Accept terms of service */
+        try {
+            UiObject acceptButton = new UiObject(selector.resourceId("org.chromium.chrome:id/lightweight_fre_terms_accept")
```
ep1cman: This is hard coded for chromium, when I tried using chrome this failed

mwasilew: Hmm, it's in a try-catch block, so should just go on if it's not on Chromium. BTW, which device did you use? Default browser crashes on my board, so I can't test with anything else than Chromium

ep1cman: Im testing on a nexus10. The problem here is that chrome also has such a dialog and it also needs to be dismissed.

mwasilew: OK, I'll find some device that browser doesn't crash on and try to fix it.

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+        self.octane2_on_device = self.device.path.join(self.device.working_directory, self.name)
+        self.index_octane = 'file://{}'.format(self.octane2_on_device) + '/index.html'
+
+        # assume that it's enough to have index.html in the dependencies
+        if not os.path.isdir(self.dependencies_directory) or \
+              not os.path.exists(os.path.join(self.dependencies_directory, "index.html")):
+            self._download_octane2_file()
+
+        # Push the octane2
+        if self.force_dependency_push or not self.device.file_exists(self.octane2_on_device):
```
ep1cman: I think this also needs to check for index.html because for some reason I had the directory on my device but not that file

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+            self.logger.debug('Copying octane2 to device.')
+            self.device.push_file(self.dependencies_directory, self.octane2_on_device, timeout=300)
```
ep1cman: This end up being something like:
` adb  push '/work/home/.workload_automation/dependencies/octane2' '/sdcard/wa-working/octane2'`

This ends up with the octane files being place into `/sdcard/wa-working/octane2/octane2/` you should replace `self.octane2_on_device` with `self.device.working_directory`

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
mwasilew: I'm close to giving up on Chrome. I got some nexus10 and downloaded Android 5.1.1 image on it. There is additional dialog about setting up Google account (can be dismissed easily) but finally uiautomator isn't able to inspect the contents of the webview in Chrome. Maybe I'm doing sth wrong. It works with Chromium, but not with Chrome :(

ep1cman: That's weird because on the device I have once I manually dismissed the dialog the automation did click to start the benchmark. Push your changes and I can test it.

wlauto/workloads/octane2/__init__.py
mwasilew: That doesn't seem to be the case for me. I'm getting similar log:
`push '/home/milosz/.workload_automation/dependencies/octane2' '/sdcard/wa-working/octane2'`

but it creates the proper directory structure:

> root@hikey:/sdcard/wa-working/octane2 # pwd
> /sdcard/wa-working/octane2
> root@hikey:/sdcard/wa-working/octane2 # ls
> CHANGELOG.md
> CONTRIBUTING.md
> LICENSE
> README.md
> base.js
> box2d.js
> code-load.js
> crypto.js
> css
> deltablue.js
> earley-boyer.js
> gbemu-part1.js
> gbemu-part2.js
> ico
> index.html
> js
> mandreel.js
> navier-stokes.js
> pdfjs.js
> raytrace.js
> regexp.js
> richards.js
> run.js
> splay.js
> typescript-compiler.js
> typescript-input.js
> typescript.js
> zlib-data.js
> zlib.js
> 

When there is no trailing / adb pushes the contents of the directory into another dir. At least this is how it works for me.

wlauto/workloads/octane2/__init__.py
mwasilew: Added.

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
mwasilew: Pushed changes. Still works for me and works on Nexus10 (with stock image) to the point where the octane2 is loaded. I've no idea how to resolve uiautomator dump problem with nexus10.

wlauto/workloads/octane2/uiauto/project.properties
```diff
@@ -0,0 +1,14 @@
...
+#
+# To customize properties used by the Ant build system edit
+# "ant.properties", and override values to adapt the script to your
+# project structure.
+#
+# To enable ProGuard to shrink and obfuscate your code, uncomment this (available properties: sdk.dir, user.home):
+#proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
+
+# Project target.
+target=android-24
```
ep1cman: is there a reason this needs `android-24` rather than `android-19` like all other uiautomator workloads?

mwasilew: No, it just didn't build for me with android-18/android-19 (probably some missing bits in my sdk). I'll try to update and check the build locally. There shouldn't be any issues.

mwasilew: Just managed to fix my SDK and it builds with android-19 :)

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
```diff
@@ -0,0 +1,102 @@
...
+        try {
+            UiObject acceptButton = new UiObject(selector.text("Accept & continue")
+                .className("android.widget.Button"));
+            acceptButton.click();
+        } catch(UiObjectNotFoundException e) {
+            /* Do nothing. Apparently dialog wasn't there */
+        }
+        /* Don't set up account */
+        try {
+            UiObject acceptButton = new UiObject(selector.text("No, thanks")
```
ep1cman: does this have a comma in the chromium version? On the device im using its just `No Thanks`

mwasilew: This dialog doesn't come up on the chromium I'm using (on HiKey with Android M). I found it on Nexus10 when trying the default browser (probably some flavour of Chrome). I'll change it to textMatches with proper regex to account for version with and without ','.

wlauto/external/uiauto/src/com/arm/wlauto/uiauto/BaseUiAutomation.java
```diff
@@ -115,6 +115,17 @@ public boolean takeScreenshot(String name) {
...
+    public void waitResourceId(String text, int second, String className) throws UiObjectNotFoundException {
+        UiSelector selector = new UiSelector();
+        UiObject resourceObj = new UiObject(selector.resourceId(text)
+                                       .className(className));
+        waitObject(resourceObj, second);
+    }
+
     public void waitText(String text) throws UiObjectNotFoundException {
         waitText(text, 600);
     }
```
ep1cman: Can you include rebuilt BaseUiAutomation classes with this commit

ep1cman:
This doesnt seem to pick up results from logcat properly, I can see them in the log though:

```
08-26 13:39:06.739 W/ActivityManager(  437): getRunningAppProcesses: caller 10101 does not hold REAL_GET_TASKS; limiting output
08-26 13:39:16.787 I/DownloadManagerWrapper(  734): onReceive() : Charging = true
08-26 13:39:17.502 I/QueryController(27029): Matched selector: UiSelector[CLASS=android.view.View, RESOURCE_ID=run-octane] <<==>> [android.view.accessibility.AccessibilityNodeInfo@9735; boundsInParent: Rect(55, 60 - 1225, 180); boundsInScreen: Rect(110, 362 - 2450, 602); packageName: com.android.chrome; className: android.view.View; text: null; error: null; maxTextLength: -1; contentDescription: Start Octane 2.0 ; viewIdResName: run-octane; checkable: false; checked: false; focusable: true; focused: false; selected: false; clickable: true; longClickable: false; enabled: true; password: false; scrollable: false; actions: [AccessibilityAction: ACTION_FOCUS - null, AccessibilityAction: ACTION_CLICK - null, AccessibilityAction: ACTION_ACCESSIBILITY_FOCUS - null, AccessibilityAction: ACTION_NEXT_AT_MOVEMENT_GRANULARITY - null, AccessibilityAction: ACTION_PREVIOUS_AT_MOVEMENT_GRANULARITY - null, AccessibilityAction: ACTION_NEXT_HTML_ELEMENT - null, AccessibilityAction: ACTION_PREVIOUS_HTML_ELEMENT - null]]
08-26 13:39:17.503 D/InteractionController(27029): clickAndSync(1280, 482)
08-26 13:39:19.737 I/chromium(26815): [INFO:CONSOLE(61)] "Richards: 6654", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:19.911 E/WifiStateMachine(  437): WifiStateMachine CMD_START_SCAN source -2 txSuccessRate=2.40 rxSuccessRate=1.82 targetRoamBSSID=84:d4:7e:90:94:c0 RSSI=-52
08-26 13:39:21.833 I/chromium(26815): [INFO:CONSOLE(61)] "DeltaBlue: 12186", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:26.010 I/chromium(26815): [INFO:CONSOLE(61)] "Crypto: 7668", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:28.129 I/chromium(26815): [INFO:CONSOLE(61)] "RayTrace: 9906", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:32.326 I/chromium(26815): [INFO:CONSOLE(61)] "EarleyBoyer: 8010", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:35.982 I/chromium(26815): [INFO:CONSOLE(61)] "RegExp: 918", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:36.838 W/ActivityManager(  437): getRunningAppProcesses: caller 10101 does not hold REAL_GET_TASKS; limiting output
08-26 13:39:38.497 I/chromium(26815): [INFO:CONSOLE(61)] "Splay: 4404", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:38.500 I/chromium(26815): [INFO:CONSOLE(61)] "SplayLatency: 6547", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:39.912 E/WifiStateMachine(  437): WifiStateMachine CMD_START_SCAN source -2 txSuccessRate=0.08 rxSuccessRate=0.06 targetRoamBSSID=84:d4:7e:90:94:c0 RSSI=-55
08-26 13:39:40.606 I/chromium(26815): [INFO:CONSOLE(61)] "NavierStokes: 8286", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:41.768 E/WifiConfigStore(  437):  rewrite network history for "AccessNG"WPA_EAP
08-26 13:39:43.223 I/chromium(26815): [INFO:CONSOLE(61)] "PdfJS: 3843", source: file:///sdcard/wa-working/octane2/index.html (61)
08-26 13:39:46.259 I/lowmemorykiller(  123): Killing '.esfm' (21708), uid 10083, adj 15
08-26 13:39:46.259 I/lowmemorykiller(  123):    to free 33748kB because cache 182684kB is below limit 184320kB for oom_adj 15
08-26 13:39:46.259 I/lowmemorykiller(  123):    Free memory is 43160kB above reserved
08-26 13:39:46.418 I/ActivityManager(  437): Process .esfm (pid 21708) has died
08-26 13:39:46.418 W/ActivityManager(  437): Scheduling restart of crashed service com.estrongs.android.pop/com.estrongs.android.scanner.service.FileMonitorService in 1000ms
08-26 13:39:46.450 I/chromium(26815): [INFO:CONSOLE(61)] "Mandreel: 4119", source: file:///sdcard/wa-working/octane2/index.html (61)

```

mwasilew:
The results reported by uiautomator script should look like this:

`01-01 00:09:04.965 V/octane2 ( 3186): OCTANE2 RESULT: Richards 2800`

If they're not in the logcat, something went wrong.

ep1cman:
if you look closely they are in there, but they dont have the prefix:
`08-26 13:39:19.737 I/chromium(26815): [INFO:CONSOLE(61)] "Richards: 6654", source: file:///sdcard/wa-working/octane2/index.html (61)`

mwasilew:
i have both in my logfile:

> 01-01 00:06:03.271 I/chromium( 3197): 
> 01-01 00:06:03.321 I/chromium( 3154): [INFO:CONSOLE(61)] "Richards: 2800", source: file:///sdcard/wa-working/octane2/index.html (61)
> 01-01 00:06:03.347 I/chromium( 3197): [INFO:render_accessibility_impl.cc(344)] Accessibility event: textChanged on node id 45
> 01-01 00:06:03.347 I/chromium( 3197): id=45 staticText READONLY (154, 282)-(45, 24) name_from=contents background_color=&FFB8B7B6 color=&FF5E3534 name=2800 value= display=block font_size=20
> 01-01 00:06:03.347 I/chromium( 3197): 
> 

and then:

> 
> 01-01 00:09:04.965 I/QueryController( 3186): Matched selector: UiSelector[CHILD=UiSelector[CLASS=android.view.View, INDEX=1], RESOURCE_ID=Box-Richards] <<==>> [android.view.accessibility.AccessibilityNodeInfo@b052; boundsInParent: Rect(116, 2 - 161, 26); boundsInScreen: Rect(154, 348 - 199, 372); packageName: org.chromium.chrome; className: android.view.View; text: null; error: null; maxTextLength: -1; contentDescription: 2800; viewIdResName: null; checkable: false; checked: false; focusable: false; focused: false; selected: false; clickable: true; longClickable: false; contextClickable: false; enabled: true; password: false; scrollable: false; actions: [AccessibilityAction: ACTION_CLICK - null, AccessibilityAction: ACTION_ACCESSIBILITY_FOCUS - null, AccessibilityAction: ACTION_NEXT_AT_MOVEMENT_GRANULARITY - null, AccessibilityAction: ACTION_PREVIOUS_AT_MOVEMENT_GRANULARITY - null, AccessibilityAction: ACTION_NEXT_HTML_ELEMENT - null, AccessibilityAction: ACTION_PREVIOUS_HTML_ELEMENT - null]]
> 01-01 00:09:04.965 V/octane2 ( 3186): OCTANE2 RESULT: Richards 2800

I think the 'matched selector' log is where this code:
https://github.com/mwasilew/workload-automation/blob/octane2/wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java#L96
is executed. So if you don't have this line in the code, it might mean that uiautomator script didn't catch the event of finishing the benchmark :( Would it be possible to send the full log?

mwasilew:
OK, I think I found the reason why result parsing fails on Chrome. The resource IDs are not populated for all parts of the webpage. Should be possible to fix.

wlauto/workloads/octane2/uiauto/project.properties
mwasilew: Fixed.

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
mwasilew: Changed to regexp. It works for me. Please check.

wlauto/external/uiauto/src/com/arm/wlauto/uiauto/BaseUiAutomation.java
mwasilew: Fixed.

mwasilew:
I fixed the result collection from UI. It requires patching octane. Pull request here: https://github.com/chromium/octane/pull/39. Fixed version can be downloaded from my repo: https://github.com/mwasilew/octane/tree/fix-richards-box. The fix is just one line: https://github.com/mwasilew/octane/commit/62e6627a456cd0accdea71652f4b371160c35cd1. With this fix result collection works fine for me.

mwasilew:
Here is snippet from my agenda file:

```
workloads:
- name: octane2
  id: octane2
  workload_parameters:
    octane_timeout: 1000
    force_dependency_push: True
    octane_archive: https://github.com/mwasilew/octane/archive/fix-richards-box.zip
    octane_filename: fix-richards-box.zip

```
For some reason I have issues running the workload in full on Nexus10. The call:
```
am start -W -n  com.android.chrome/com.google.android.apps.chrome.Main about:blank

```
doesn't exit although the browser is launched.

ep1cman:
I had the same issue with am blocking even though the browser launched. Clearing the app data seemed to solve it

mwasilew:
Thanks. Works now on Nexus10. The regex I uses is wrong apparently. There will be one more update to the PR.

mwasilew:
I finally found the way the regexs work in Java. Fixed all issues. Works well on Chrome on Nexus10 and custom built Chromium on HiKey. It's important to use the octane2 from my repo until the patch fixing the index.html is merged to master (hopefully not long).

ep1cman:
Have your changes been merged into octane2 yet?

mwasilew:
No, it's extremely slow. I got some reply from octane developers only today (after a week of silence). I'm still keeping my fingers crossed tightly (in hope it helps :P).

mwasilew:
Sebastian,
My PR to octane was merged. Should I rebase this patch now or is it OK to merge as it is?

ep1cman:
Hi, sorry for taking a while to get back to you, I will now start testing this workload

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
```diff
@@ -0,0 +1,101 @@
...
+            UiObject acceptButton = new UiObject(selector.textMatches("(?i:No.?\\sThanks)")
+                .className("android.widget.Button"));
+            acceptButton.click();
+        } catch(UiObjectNotFoundException e) {
+            /* Do nothing. Apparently dialog wasn't there */
+        }
+
+        /* wait for browser to load index */
+        sleep(timeout);
+        UiObject runButton = new UiObject(selector.resourceId("run-octane")
```
ep1cman: It seems in the latest chrome that the resourceID is not sent, why not look for the text `Start Octane 2.0` instead?

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
```diff
@@ -0,0 +1,101 @@
...
+            .className("android.view.View"));
+        runButton.click();
+
+        try {
+            waitResourceId("bottom-text", "android.view.View");
+            extractDetailedScores();
+        } catch(UiObjectNotFoundException e) {
+            takeScreenshot("octane2-error");
```
ep1cman: WA already takes a screenshot when a workload errors. Let the error occur, this also ensures the workload does not try to process results that don't exist.

wlauto/workloads/octane2/__init__.py
```diff
@@ -0,0 +1,183 @@
...
+                os.path.join(self.octane2_on_device, "index.html")):
+            self.logger.debug('Copying octane2 to device.')
+            self.device.push_file(self.dependencies_directory, self.octane2_on_device, timeout=300)
+
+        # Open the browser with default page
+        self.device.execute('am start -W -n  {}/{} about:blank'.format(self.browser_package, self.browser_activity))
+
+        # Stop the browser if already running and wait for it to stop
+        self.device.execute('am force-stop {}'.format(self.browser_package))
+
```
ep1cman: Why are you opening and closing the browser at the start?

mwasilew:
Sebastian,
Sorry for delay. I'll try to get back to this work this week.

wlauto/workloads/octane2/__init__.py
mwasilew: This is copy/paste from bbench. Probably not necessary here but ensures the browser activity is loaded and unloaded before actual benchmark starts.

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
mwasilew: OK, so what is the proper thing to do in this case? Just throw error (remove try/catch block)?

setrofim: Yes, just remove the try/catch block. The error will be detected and handled by the framework.

wlauto/workloads/octane2/__init__.py
setrofim: Yes, this isn't necessary here. Please remove.

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
mwasilew: Fixed that. The 'Start Octane 2.0' text comes from contentDescription field.

wlauto/workloads/octane2/uiauto/src/org/linaro/wlauto/uiauto/UiAutomation.java
mwasilew: OK, removed

wlauto/workloads/octane2/__init__.py
mwasilew: I removed start, but left the force-stop to make sure browser is always start fresh.

mwasilew:
The code was updated including last comments. Please check if anything else needs fixing.

setrofim:
I get
```

2016-10-10 17:33:03,996 ERROR    Runner: Error while running octane2:
2016-10-10 17:33:03,996 ERROR    Runner:        Could not start activity; got the following:
2016-10-10 17:33:03,996 ERROR    Runner: Error: Activity class {com.android.chrome/com.android.chrome.browser.ChromeTabbedActivity} does not exist
```

When I try to run this (`com.android.chrome` is the correct browser package for the devices I'm trying this on -- an S5 and a Mate8). Could it be that the activity name also needs to be adjusted?

mwasilew:
You can set this in agenda file. Variable name is: `browser_activity`. I don't know any means of extracting the activity name at runtime.

setrofim:
I've set the activity and it has gotten further (it's probably worth changing the defaults for package/activity to be for the production version of chrome, as that is the common case). However, it is till failing. It runs Octane and gets to the score screen, but then gets stuck there. From, what I can tell, it is failing on 

        waitResourceId("bottom-text", "android.view.View");

According to the UI tree dump, there is nothing there with ID "bottom-text" (see attached files). 
[ui-dump.xml.txt](https://github.com/ARM-software/workload-automation/files/521827/ui-dump.xml.txt)
screen

I'm trying this on a Galaxy S5.

Also, it looks like there are a few PEP8  issues in Python code. Please re-run the checker and address.

mwasilew:
The problem will happen on the small screens. The most bottom part of the page is not displayed, so uiautomator doesn't dump it's structure. Is there any helper in your library that allows searching with scrolling? It works fine for me on a bigger screen, when all items can be displayed.

setrofim:
There isn't. You'd have to use the UIAutomator's `UIScrollable` APIs or similar. In this case, you'd first need to wait on something that is guaranteed to be on the screen when the test finishes (e.g. the "Octane Score" string) and then extract the results, scrolling as necessary, as you cannot assume they would all be on the screen.

mwasilew:
Caveat is that 'Octane Score' doesn't appear in uiautomator dump. So the only way to check whether the testing finished is to check for the bottom section :( JS doesn't work well with uiautomator.

setrofim:
Hm strange. It does seem to appear in the dumped UI hierarchy (see the previously attached dump). A work around for this might be to periodically call `UiDevice.dumpWindowHierarchy` and parse the resulting XML to look for the scores. This will also remove the need for scrolling, as that method should dump the complete hierarchy, rather than just what is currently in the view port.

mwasilew:
Here is my dump from Nexus10. It's pretty similar to what I have on HiKey. 
[nexus10_dump.xml.txt](https://github.com/ARM-software/workload-automation/files/524114/nexus10_dump.xml.txt)

There box with total score is not updated. At least UIAutomator can't see this change. However the test works fine on Nexus10 even though in horizontal mode the bottom of the screen isn't visible. 

My primary goal is Chromium, so I'll leave the defaults as they are. I'm also not targeting small screens. Primary focus is on HiKey. It looks like covering all browsers and all possible devices is a bit too complicated.

mwasilew:
It seems there is no more interest in this workload. Closing the PR.

PR #235, opened 2016-08-31
Add method to check AND raise if device is not connected to internet

muendelezaji:
This ensures consistency across workloads by allowing them to replace:
  ```
  if not self.device.is_network_connected():
    # handle error here e.g. raise DeviceError
  ```

with a simple call to: `self.device.check_network_connected()`

The former approach can still be used by workloads that choose to do so

### DIFF
--- a/wlauto/common/linux/device.py
+++ b/wlauto/common/linux/device.py
@@ -320,11 +320,11 @@ class BaseLinuxDevice(Device):  # pylint: disable=abstract-method
 
     def is_network_connected(self, ip_address='8.8.8.8'):
         """
-        Checks for internet connectivity on the device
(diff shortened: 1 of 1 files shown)


setrofim:
Do we really need a separate method for this two-liner?

muendelezaji:
Not entirely necessary, but we thought it useful to have a consistent message if anyone else will use it. At the moment it is used in 4 workloads - skype, books, youtube and gmail.

setrofim:
This shouldn't be a `DeviceError` -- that implies something has gone wrong on the device. Where as what you're actually trying to signal is that requirements for a specific workload are not met, so a `WorkloadError` should be raised.

PR #237, opened 2016-09-01
Add network check methods to Device and Workload classes

muendelezaji:
- Device subclasses should provide their own implementation.
    Default behaviour is to raise a `NotImplementedError`
  - Workload subclasses can set `requires_network` to `True` and
    network connectivity check will be performed during `setup()`

### DIFF
--- a/wlauto/common/linux/device.py
+++ b/wlauto/common/linux/device.py
@@ -39,6 +39,8 @@ FstabEntry = namedtuple('FstabEntry', ['device', 'mount_point', 'fs_type', 'opti
 PsEntry = namedtuple('PsEntry', 'user pid ppid vsize rss wchan pc state name')
 LsmodEntry = namedtuple('LsmodEntry', ['name', 'size', 'use_count', 'used_by'])
 
+GOOGLE_DNS_SERVER_ADDRESS = '8.8.8.8'
+
 
 class BaseLinuxDevice(Device):  # pylint: disable=abstract-method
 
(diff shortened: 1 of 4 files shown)


wlauto/core/workload.py
```diff
@@ -78,6 +79,8 @@ def setup(self, context):
         the workload.
 
         """
+        if self.requires_network:
+            self.check_network_connected()
         pass
```
ep1cman: This should be removed

wlauto/common/linux/device.py
```diff
@@ -318,13 +318,13 @@ def is_installed(self, name):  # pylint: disable=unused-argument,no-self-use
 
             please see: https://pythonhosted.org/wlauto/writing_extensions.html""")
 
-    def is_network_connected(self, ip_address='8.8.8.8'):
+    def is_network_connected(self, ip_address='8.8.8.8'):  # pylint: disable=arguments-differ
```
ep1cman: The arguments shouldn't differ either they all should have ip or none of them

wlauto/core/workload.py
muendelezaji: Done

wlauto/common/linux/device.py
muendelezaji: Updated. Removed the argument for the interface method. Device subclasses can have their own internal implementation to check for internet e.g. `LinuxDevice` uses `_ping_server()`

wlauto/core/workload.py
```diff
@@ -99,5 +101,10 @@ def teardown(self, context):
     def finalize(self, context):
         pass
 
+    def check_network_connected(self):
+        if not self.device.is_network_connected():
+            message = 'Workload {} requires internet. Device {} does not appear to be connected to the internet.'
```
ep1cman: i would replace `{}` with `"{}"`

PR #238, opened 2016-09-01
Add check_app_version method to AndroidUxPerfWorkload

johnrchrdsn:
Implement new method to enforce that a valid apk version is used to run
the workload. Based on a min and max range of apk versions tested
during development.

### DIFF
--- a/wlauto/common/android/workload.py
+++ b/wlauto/common/android/workload.py
@@ -18,6 +18,8 @@ import sys
 import time
 from math import ceil
 
+from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


setrofim:
- Why are `min`/`max_app_version` implemented as properties rather than class attributes in the base?
- The code assumes that they are always set, which may not be the case.

setrofim:
- This should be added to `ApkWorkload` rather than `AndroidUxPerfWorkload` as this is a mechanism for dealing with APK's

PR #241, opened 2016-09-06
Add Skype workload

johnrchrdsn:
- Add Skype workload
- Add ApkWorkload extension support to launch non-default packages names.


wlauto/common/android/workload.py
```diff
@@ -603,7 +616,7 @@ class GameWorkload(ApkWorkload, ReventWorkload):
                   after setup and run"""),
         Parameter('assets_push_timeout', kind=int, default=500,
                   description='Timeout used during deployment of the assets package (if there is one).'),
-        Parameter('clear_data_on_reset', kind=bool, default=True,
+        Parameter('clear_data_on_reset', kind=bool, default=True, override=True,
```
setrofim: Since this override does not change anything, it should be removed.

wlauto/common/android/workload.py
```diff
@@ -181,6 +187,11 @@ class ApkWorkload(Workload):
                   If ``True``, workload will check that the APK matches the target
                   device ABI, otherwise any APK found will be used.
                   '''),
+        Parameter('clear_data_on_reset', kind=bool, default=True,
```
ep1cman: Why was this added? It doesn't seem to be used by your workload

wlauto/workloads/skype/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+
+import time
+from wlauto import AndroidUxPerfWorkload, Parameter
+
+
+class Skype(AndroidUxPerfWorkload):
+
+    name = 'skype'
+    description = '''
+    A workload to perform standard productivity tasks within Skype.  The workload logs in to the
```
ep1cman: similarly to the old `skypevideo` workload this should have instructions covering the fact that a skype acocunt needs to be setup. For voice you can just use the skype echo service but for video a second skype account that auto accepts calls is required. 

In light of this I think the workload should default to voice not video too and default the contact_name to `Echo / Sound Test Service`

ep1cman:
This PR should replace the old `skypevideo` workload

PR #242, opened 2016-09-06
Add Google slides workload

muendelezaji:


wlauto/workloads/googleslides/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+        self.deployable_assets += [self.test_image]
+        if self.use_test_file and self.test_file:
+            self.deployable_assets += [self.test_file]
+
+    def validate(self):
+        super(GoogleSlides, self).validate()
+        if self.use_test_file and not self.test_file:
+            raise WorkloadError('Parameter use_test_file is "True" but test_file was not specified')
+
+        self.uiauto_params['workdir_name'] = os.path.basename(os.path.normpath(self.device.working_directory))
```
setrofim: Should be using `self.device.path` instead of `os.path` for this.

muendelezaji: Updated

PR #247, opened 2016-09-19
GooglePlayBooks: Modified ActionLogger tags and capture points

mcgeagh:


wlauto/workloads/googleplaybooks/uiauto/src/com/arm/wlauto/uiauto/UiAutomation.java
```diff
@@ -307,9 +311,6 @@ private void gesturesTest() throws Exception {
                 case UIDEVICE_SWIPE:
                     uiDeviceSwipe(dir, steps);
                     break;
-                case UIOBJECT_SWIPE:
-                    uiObjectSwipe(pageView, dir, steps);
-                    break;
```
ep1cman: Why was this removed?

mcgeagh: Cody Tidy. Case is never hit.

PR #263, opened 2016-09-28
GooglePlayBooks: Code tidy to conform.

mcgeagh:
Additional parameter added for when the search name differs to the library name.
Default values changed due to Hamlet disappearing from the store.


mcgeagh:
Works on S5, S7, Mate8, MX4Pro

PR #264, opened 2016-09-29
Reader Workload: The Adobe Reader workflow carries out typical…

mcgeagh:
… productivity tasks.

Conform to code style of other workloads.

Changed search strings to be a list. Allows user the ability to search for 1 or more search terms.


wlauto/workloads/reader/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+import os
+
+from wlauto import AndroidUxPerfWorkload, Parameter
+from wlauto.exceptions import ValidationError
+from wlauto.utils.types import list_of_strings
+
+
+class Reader(AndroidUxPerfWorkload):
+
+    name = 'reader'
```
setrofim: The name is too general. Change to "adobereader" or "pdfreader"

wlauto/workloads/reader/uiauto/src/com/arm/wlauto/uiauto/UiAutomation.java
```diff
@@ -0,0 +1,238 @@
...
+        String[] searchStrings =
+            parameters.getString("search_string_list").replace("0space0", " ").split("0newline0");
+
+        setScreenOrientation(ScreenOrientation.NATURAL);
+
+        dismissWelcomeView();
+        openFile(filename);
+        gesturesTest();
+        // Wait one second then tap the centre to bring up the menu gui
+        sleep(1);
```
setrofim: The comment should explain why the wait is necessary. Also, is 1 second going to be enough for slower devices?

mcgeagh: Changed. Now detects if searchicon still hasnt appeared, will retry

wlauto/workloads/reader/__init__.py
setrofim: All was plugin names are lower case, so `adobereader`, `adobe_reader` or `adobe-reader`

PR #271, opened 2016-10-26
ReventWorkload

marcbonnici:
ReventWorkload: Added argument to take screen capture with revent record command
ReventWorkload: State detection fixes

### DIFF
--- a/wlauto/commands/record.py
+++ b/wlauto/commands/record.py
@@ -101,6 +101,8 @@ class RecordCommand(ReventCommand):
         self.parser.add_argument('-p', '--package', help='Package to launch before recording')
         self.parser.add_argument('-C', '--clear', help='Clear app cache before launching it',
                                  action="store_true")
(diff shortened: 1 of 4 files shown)


wlauto/commands/record.py
```diff
@@ -101,6 +101,8 @@ def initialize(self, context):
         self.parser.add_argument('-p', '--package', help='Package to launch before recording')
         self.parser.add_argument('-C', '--clear', help='Clear app cache before launching it',
                                  action="store_true")
+        self.parser.add_argument('-sc', '--screencapture', help='Record a screen capture after recording',
```
setrofim: The '-' name must be a single letter (see http://www.faqs.org/docs/artu/ch10s05.html). I would change the long name to `--capture-screen` -- separating words with dashes adds to readability, and the word ordering more natural (an alternative may by `--enable-screen-capture`, but that's longer).

PR #278, opened 2016-11-04
ReventWorkload: Modified statedection to vary scale of images

marcbonnici:
Due to inaccuracies in revent playback, the resultant state does not
always match the templates precisely causing state detection to fail.
To help this images are now scaled to different sizes before being
matched to the templates to compensate for slight variations in size.

### DIFF
--- a/wlauto/utils/statedetect.py
+++ b/wlauto/utils/statedetect.py
@@ -91,14 +91,24 @@ def match_state(screenshot_file, defpath, state_definitions):  # pylint: disable
     for template_png in template_list:
         template = cv2.imread(os.path.join(defpath, 'templates', template_png + '.png'), 0)
         template_edge = auto_canny(template)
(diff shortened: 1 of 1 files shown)


wlauto/utils/statedetect.py
```diff
@@ -91,14 +91,24 @@ def match_state(screenshot_file, defpath, state_definitions):  # pylint: disable
     for template_png in template_list:
         template = cv2.imread(os.path.join(defpath, 'templates', template_png + '.png'), 0)
         template_edge = auto_canny(template)
+        (tH, tW) = template_edge.shape[:2]
```
setrofim: parens unnecessary

setrofim: Names should be more descriptive and all lower case: `template_height` and `template_width`

marcbonnici:
Updated with your corrections.

PR #294, opened 2016-11-24
geekbench: Improve UiAutomator timeouts for slower devices

jimboatarm:
Tested by running Mate 8 on the lowest cluster at the lowest available frequency


wlauto/workloads/geekbench/__init__.py
```diff
@@ -109,9 +109,13 @@ def __init__(self, device, **kwargs):
 
     def update_result(self, context):
         super(Geekbench, self).update_result(context)
-        major_version = versiontuple(self.version)[0]
-        update_method = getattr(self, 'update_result_{}'.format(major_version))
-        update_method(context)
+        # It is only possible to pull files from /data/data on rooted devices.
+        # Skip this step on devices without priveleges (e.g. Chromebooks) as the logs can be
+        # retrieved manually afterwards
+        if self.device.is_rooted:
```
setrofim: This will result in a "silent" failure on unrooted devices, and the users has to somehow know to pull the files manually afterwards. This manual step makes this workload useless for a large number of applications.

Instead, the workload should check if the device is rooted inside `initialize()` and fail with a `WorkloadError` if not. If it is still desirable to be able to run this on unrooted devices without generating results, this should be done via a parameter that the user has to explicitly enable.

wlauto/workloads/geekbench/__init__.py
```diff
@@ -101,7 +103,8 @@ def __init__(self, device, **kwargs):
         super(Geekbench, self).__init__(device, **kwargs)
         self.uiauto_params['version'] = self.version
         self.uiauto_params['times'] = self.times
-        self.run_timeout = 10 * 60 * self.times
+        if self.times > 1:
+            self.run_timeout = 15 * 60 * self.times
```
setrofim: This silently overwrites user-specified `run_timeout` when `times` is greater than one. Ether

a) replace this with `self.run_timeout = self.run_timeout * self.times` and make it clear in the documentation for the `run_timeout` parameter that it will be multiplied by the number of times, or
b) respect the user-provided timeout regardless of the the value of `times`.

wlauto/workloads/geekbench/uiauto/src/com/arm/wlauto/uiauto/UiAutomation.java
```diff
@@ -160,7 +156,7 @@ public void shareResults() throws Exception {
         getUiDevice().pressMenu();
         UiObject shareButton = new UiObject(selector.text("Share")
                                                     .className("android.widget.TextView"));
-        shareButton.waitForExists(WAIT_TIMEOUT_1SEC);
+        shareButton.waitForExists(WAIT_TIMEOUT_20MIN);
```
setrofim: Why does the timeout need to be this long here? This is basically waiting for the app to generate the json data dump?

wlauto/workloads/geekbench/__init__.py
```diff
@@ -87,6 +87,14 @@ class Geekbench(AndroidUiAutoBenchmark):
         Parameter('times', kind=int, default=1,
                   description=('Specfies the number of times the benchmark will be run in a "tight '
                                'loop", i.e. without performaing setup/teardown inbetween.')),
+        Parameter('iteration_timeout', kind=int, default=900,
+                  description='Timeout for a single iteration of the benchmark. This value is '
+                              ' multiplied by ''times'' to calculate the overall run timeout. '),
```
setrofim: Don't need leading space on this line since there is a trailing space at the end of the previous line.

wlauto/workloads/geekbench/__init__.py
```diff
@@ -87,6 +87,14 @@ class Geekbench(AndroidUiAutoBenchmark):
         Parameter('times', kind=int, default=1,
                   description=('Specfies the number of times the benchmark will be run in a "tight '
                                'loop", i.e. without performaing setup/teardown inbetween.')),
+        Parameter('iteration_timeout', kind=int, default=900,
```
setrofim: Rename to just `timeout` -- for consistency with other workloads; also the term "iteration" should not be applied to the "times" parameter as it means something else in the context of a WA run.

PR #297, opened 2016-11-25
ResourceGetters: Updated executable discovery to use http_getter

marcbonnici:
Executables were not previously automatically downloaded from a
provided remote_assets_url. The http_getter is now called to attempt to
find and download executable files.

### DIFF
--- a/wlauto/resource_getters/standard.py
+++ b/wlauto/resource_getters/standard.py
@@ -345,7 +345,7 @@ class HttpGetter(ResourceGetter):
 
     """
     priority = GetterPriority.remote
-    resource_type = ['apk', 'file', 'jar', 'revent']
+    resource_type = ['apk', 'file', 'jar', 'revent', 'executable']
 
     parameters = [
         Parameter('url', global_alias='remote_assets_url',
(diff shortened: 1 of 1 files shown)


wlauto/resource_getters/standard.py
```diff
@@ -441,6 +441,12 @@ def resolve_resource(self, resource):
                             return asset
                         except ValueError as e:
                             self.logger.warning(e.message)
+        elif resource.name == 'executable':
+            device_abi = resource.owner.device.abi
+            path = '/'.join((['bin', device_abi, resource.filename]))
```
setrofim: The second set of parens is unnecessary.

wlauto/resource_getters/standard.py
```diff
@@ -441,6 +441,12 @@ def resolve_resource(self, resource):
                             return asset
                         except ValueError as e:
                             self.logger.warning(e.message)
+        elif resource.name == 'executable':
+            device_abi = resource.owner.device.abi
```
setrofim: Should use `resource.platform` instead of the device's ABI. The owner may not have a device (and the resource requested may not be for the device).

marcbonnici:
Updated.

PR #298, opened 2016-11-28
Revent: Fixes

marcbonnici:
Changed termination signal to interrupt signal to prevent code exiting too early.
Added exit handler to ensure revent exits correctly as previously was
crashing and therefore not running final code.
Fixed error in writing input event where half of timestamp seconds was missing.
Fixed typo in documentation for revent file structure.

### DIFF
--- a/wlauto/external/revent/revent.c
+++ b/wlauto/external/revent/revent.c
@@ -226,25 +226,25 @@ void destroy_replay_device(int fd)
 		die("Could not destroy replay device");
 }
 
-inline void set_evbit(fd, bit)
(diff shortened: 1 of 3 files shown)


wlauto/external/revent/revent.c
```diff
@@ -1086,11 +1100,18 @@ void record(const char *filepath, int delay, recording_mode_t mode)
...
+
+        if (EXIT){
+        	break;
+        }
+		if (errno == EINTR){
+            break;
+        }
+		if (!ret){
+            break;
+        }
```
setrofim: indentation.

PR #299, opened 2016-11-28
Decorators: Added initial version of decorators and environment manager

marcbonnici:
Added 3 decorators to enabled methods to be executed conditionally:
- Once for each method instance.
- Once for each class.
- Only once per environment including any derived classes.
Added execution control allowing for different environments to
be used in order to determine how often decorated commands should be ran.
Added relevant unittests for the above decorators.

### DIFF
--- a/wlauto/tests/test_exec_control.py
+++ b/wlauto/tests/test_exec_control.py
@@ -1,0 +1,269 @@
+#    Copyright 2013-2015 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+# pylint: disable=W0231,W0613,E0611,W0603,R0201
+from unittest import TestCase
+
+from nose.tools import assert_equal, assert_raises
+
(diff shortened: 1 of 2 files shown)


wlauto/tests/test_environment.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def test_switch_environment(self):
+        activate_environment('ENVIRONMENT1')
+        t1 = TestClass()
+        t1.initilise_once()
+        assert_equal(t1.count, 1)
+
+        activate_environment('ENVIRONMENT2 ')
+        t1.initilise_once()
+        assert_equal(t1.count, 2)
```
setrofim: To complete this test, I would re-activate ENVIRONMENT1, call `intialise_once()` again and make sure that `t1.count` does not increment

wlauto/tests/test_environment.py
```diff
@@ -0,0 +1,289 @@
...
+    @once_per_class
+    def initilise_once_per_class(self):
+        self.count += 1
+
+    @once_per_instance
+    def initilise_once_per_instance(self):
+        self.count += 1
+
+
+class SubSubClass(SubClass):
```
setrofim: `SubClass` and `SubSubClass` are overriding all of the methods of the `TestClass`, which kinda defeats the point of testing the inheritance hierarchy. You want to subclass without re-implementing, so that derived instances will call the base methods.

wlauto/tests/test_environment.py
```diff
@@ -0,0 +1,289 @@
...
+        t1.initilise_once()
+        assert_equal(t1.count, 1)
+
+        sc.initilise_once()
+        sc.initilise_once()
+        assert_equal(sc.count, 0)
+
+        ss.initilise_once()
+        ss.initilise_once()
+        assert_equal(ss.count, 0)
```
setrofim: To complete this test, I would make sure that calling `t1.initialise_once()` again does not increment the count.

PR #307, opened 2016-12-08
Modified how apks containing native code are matched to a target devices supported abi(s).

marcbonnici:
Modified how apks containing native code are matched to a target devices supported abi(s).

### DIFF
--- a/wlauto/common/android/device.py
+++ b/wlauto/common/android/device.py
@@ -115,26 +115,26 @@ class AndroidDevice(BaseLinuxDevice):  # pylint: disable=W0223
         return val
 
     @property
-    def supported_eabi(self):
+    def supported_abi(self):
         props = self.getprop()
         result = [props['ro.product.cpu.abi']]
         if 'ro.product.cpu.abi2' in props:
             result.append(props['ro.product.cpu.abi2'])
         if 'ro.product.cpu.abilist' in props:
-            for eabi in props['ro.product.cpu.abilist'].split(','):
-                if eabi not in result:
-                    result.append(eabi)
+            for abi in props['ro.product.cpu.abilist'].split(','):
+                if abi not in result:
+                    result.append(abi)
 
         mapped_result = []
-        for supported_eabi in result:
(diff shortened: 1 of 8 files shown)


wlauto/common/linux/device.py
```diff
@@ -137,6 +137,10 @@ def abi(self):
         return self._abi
 
     @property
+    def supported_eabi(self):
```
setrofim: should be `supported_abi` to be consistent with the other property

wlauto/utils/android.py
```diff
@@ -183,6 +184,19 @@ def parse(self, apk_path):
             elif line.startswith('launchable-activity:'):
                 match = self.name_regex.search(line)
                 self.activity = match.group('name')
+            elif line.startswith('native-code'):
+                apk_abis = [entry for entry in [entry.strip() for entry in line.split(':')[1].split("'")] if entry]
```
setrofim: could be replaced with

         apk_abis = [entry.strip() for entry in line.split(':')[1].split("'") if entry.strip()]

wlauto/common/linux/device.py
setrofim: Yes, it should be changed. This is a WA `Device` property -- it is not android specific (it should be implemented for both Android and Linux devices). It just so happens that for Android it is implemented by querying the similarly-named Android property, but that is an implementation detail.

wlauto/common/android/device.py
```diff
@@ -108,7 +108,11 @@ def is_rooted(self):
...
     def abi(self):
-        return self.getprop()['ro.product.cpu.abi'].split('-')[0]
+        val = self.getprop()['ro.product.cpu.abi'].split('-')[0]
+        for abi, architectures in ABI_MAP.iteritems():
+            if val in architectures:
+                return abi
+        return val
 
     @property
     def supported_eabi(self):
```
setrofim: Should be updated to ``supported_abi`` as well, while we're here.

PR #330, opened 2017-01-12
Splits ApkWorkload setup() into short methods.

jumana-arm:
Apkworkload setup phase performs many functionalities in
a single method that is broken down into short methods.
The split short methods can be called individually when
relevant use cases arise.

### DIFF
--- a/wlauto/common/android/workload.py
+++ b/wlauto/common/android/workload.py
@@ -204,7 +204,12 @@ class ApkWorkload(Workload):
 
     def setup(self, context):  # pylint: disable=too-many-branches
         Workload.setup(self, context)
+        self.setup_workload_apk(context)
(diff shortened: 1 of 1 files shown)


wlauto/common/android/workload.py
```diff
@@ -146,7 +146,7 @@ class ApkWorkload(Workload):
                   is provided at the run phase).
     :install_timeout: Timeout for the installation of the APK. This may vary wildly based on
                       the size and nature of a specific APK, and so should be defined on
-                      per-workload basis.
+                        per-workload basis.
```
setrofim: Why the extra two spaces here?

jumana-arm: Fixed it

PR #334, opened 2017-01-16
Uxperf applaunch

jumana-arm:
This is the new design of the previous pull request of branch uxperfapplaunch.

I would like to get comments on the design before I test on all workloads and multiple devices. 

I have also not done code tidying.


wlauto/external/uiauto/src/com/arm/wlauto/uiauto/UxPerfUiAutomation.java
```diff
@@ -17,7 +17,7 @@
 
 import java.util.logging.Logger;
 
-public class UxPerfUiAutomation extends BaseUiAutomation {
+public class UxPerfUiAutomation extends UxPerfApplaunch{
```
marcbonnici: Why does this now extend `UxPerfApplaunch`?

marcbonnici:
Just to say I tried testing this with the adobe reader workload and it crashed for me using `applaunch_type: launch_from_long-idle` with the error `java.io.IOException: Error running exec(). Command: [sync;, su, echo, 3, >, /proc/sys/vm/drop_caches] Working Directory: null Environment: null`

wlauto/external/uiauto/src/com/arm/wlauto/uiauto/UxPerfUiAutomation.java
jumana-arm: To avoid, multiple inheritance in the UiAutomation class of every workload.

jumana-arm:
I shall fix that, are you okay with the code style so that I can proceed?

wlauto/external/uiauto/src/com/arm/wlauto/uiauto/UxPerfUiAutomation.java
setrofim: But this inheritance hierarchy makes no sense. If anything, Shouldn't `UxPerfApplaunch` inherit from `UxPerfUiAutomation`? Applaunch is a specialisation of  UxPerf workloads, not the other way round.

Honestly, I think the may needs to be refactored to use composition rather than inheritance, otherwise, as new workload types are added, we're going to end up with an increasingly convoluted and unmanageable inheritance hierarchy.

wlauto/workloads/adobereader/uiauto/src/com/arm/wlauto/uiauto/UiAutomation.java
```diff
@@ -36,32 +37,40 @@
...
     }
+	
+    
+	//Clear the initial run dialogues of the application launch.
+    public void clearDialogues() throws Exception {
+        getParameters();
+        dismissWelcomeView();
+    }
+    
+	//Sets the UiObject that marsk teh end of the application launch.
```
setrofim: typos "marsk" --> "marks", "teh" --> "the"

setrofim:
superseded by https://github.com/ARM-software/workload-automation/pull/338

PR #342, opened 2017-02-03
Applaunch Workload

jumana-arm:


jumana-arm:
This is the same as the uxperfapplaunch-interface branch, created a new branch as the pull request there shows commits reordered.

PR #359, opened 2017-02-24
Allow user to override the method of collecting FPS data,

mcgeagh:
to fallback to SurfaceFlinger method when run on Android M and above


setrofim:
I don't understand why this change is necessary. Please could you amend the commit message to

1. explain why this change is being made
2. adhere to git conventions -- https://chris.beams.io/posts/git-commit/#seven-rules
    In addition to the above, prefix the commit heading with the part of the code being changed.

e.g.

    fps: added parameter to override FPS collection method

     <longer explanation of what is being done and why>

PR #365, opened 2017-03-08
Parameter update

scojac01:
Changed the kind to be string instead of integer. 

This is because the page numbers can occasionally be in roman numerals and we need to handle this accordingly.

### DIFF
--- a/wlauto/workloads/googleplaybooks/__init__.py
+++ b/wlauto/workloads/googleplaybooks/__init__.py
@@ -67,7 +67,7 @@ class Googleplaybooks(AndroidUxPerfWorkload):
                   The Library name can differ (usually shorter) to the Store name.
                   If left blank, the ``search_book_title`` will be used.
                   """),
-        Parameter('select_chapter_page_number', kind=int, default=4,
+        Parameter('select_chapter_page_number', kind=str, default="4",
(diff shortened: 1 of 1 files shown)


marcbonnici:
Would you be able to remove your first commit e2f212f from the pull request which introduces a typo (that the second fixes again)?

setrofim:
1. Your second commit negates the change in the first commit. Please remove the first commit
2. Please follow best practice for commit messages: https://chris.beams.io/posts/git-commit/
3. additionally, please prefix the summary line with the name of the workload, e.g.
  
       googleplaybooks: do stuff

       explain more about what is being done and why

    In this case especially, as the change is counter intuitive (why does a page *number* need to be a string?), it is important to give explanation for the change in the commit body.
4. The change of the `kind` of the parameter is wrong. This change means that `"banana"` is a valid value for the parameter. Even if the value passed to UIAutomator, for whatever reason, needs to be a string, the `kind` of the workload parameter should be an integer. If it cannot be an integer (e.g. you want to support roman numerals for pages), then I would add a new type for this in `wlauto/utils/types.py` and specify that as the `kind`; it would then be up to the type implementation to handle appropriate validation/conversion.

PR #376, opened 2017-03-28
Implementing UIAutomator-based Android workloads

setrofim:


wa/framework/workload.py
```diff
@@ -85,53 +93,117 @@ def __str__(self):
...
-    def __init__(self, target, package='', klass='UiAutomation',
-                 method='runUiAutoamtion'):
-        self.target = target
+    stages = ['setup', 'runWorkload', 'extractResults', 'teardown']
+
+    def __init__(self, owner, package=None, klass='UiAutomation', timeout=600):
+        self.owner = owner
+        self.target = self.owner.target
         self.uiauto_package = package
         self.uiauto_class = klass
```
setrofim: Yes. `class` is a reversed word in Python. So if you want to use it as an identifier, you use `klass` instead (some prefer `class_` instead, but I'm not a fan of trailing underscores).

PR #379, opened 2017-04-06
Device: Updated to set and retrieve binary data from sysfiles

marcbonnici:
`set_sysfile_values` now accepts a `^` symbol prefixed to the
file path to indicated that the value should be written as binary data.
To accommodate this an extra `binary` flag has been added to  the `set_sysfile_value`and `get_sysfile_value` methods to write and retrieve the binary data respectively.

### DIFF
--- a/wlauto/common/linux/device.py
+++ b/wlauto/common/linux/device.py
@@ -17,6 +17,7 @@
 import os
 import re
 import time
+import base64
 import socket
 from collections import namedtuple
 from subprocess import CalledProcessError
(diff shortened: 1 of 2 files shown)


wlauto/common/linux/device.py
```diff
@@ -281,21 +304,24 @@ def get_sysfile_values(self):
 
         """
         values = {}
-        for sysfile in self._written_sysfiles:
-            values[sysfile] = self.get_sysfile_value(sysfile)
+        for (sysfile, binary) in self._written_sysfiles:
```
setrofim: Don't need parens here

marcbonnici:
Updated

PR #392, opened 2017-05-30
Upgrading to UiAutomator2

marcbonnici:


wlauto/commands/create.py
```diff
@@ -392,7 +418,7 @@ def render_template(name, params):
     with open(filepath) as fh:
         text = fh.read()
         template = string.Template(text)
-        return template.substitute(params)
+        return template.safe_substitute(params)
```
setrofim: Why this change?

marcbonnici: It was originally because I used parameter substitution in some of the files themselves during the build stage so when writing the template it would leave them intact, however it wouldn't be difficult to do all of the substitution here instead if that makes more sense.

marcbonnici: Updated to remove this change.

PR #397, opened 2017-06-05
Uiautomator2: Fix for newer adb versions. 

marcbonnici:
In later versions of adb only files with the extension `.apk` are allowed to be installed. This set of changes removes the newly added uiautoapk file type and now uses an apks package name to determine its type during resource discovery.


wlauto/common/android/workload.py
```diff
@@ -84,7 +84,7 @@ def __init__(self, device, _call_super=True, **kwargs):  # pylint: disable=W0613
         self.uiauto_params = ParameterDict()
 
     def init_resources(self, context):
-        self.uiauto_file = context.resolver.get(wlauto.common.android.resources.uiautoApkFile(self))
+        self.uiauto_file = context.resolver.get(wlauto.common.android.resources.ApkFile(self, uiauto_apk=True))
```
setrofim: Rather than using the fully-qualified name here, import `ApkFile` into the current namespace.

marcbonnici:
Updated.

scojac01:
This fixed my ADB version issues. Please merge.

mcgeagh:
Several workloads' java files still are looking for '.uiautoapk' rather than '.apk'
E.G. `wlauto/workloads/applaunch/uiauto/app/src/main/java/com/arm/wlauto/uiauto/applaunch/UiAutomation.java`
Line 57: "com.arm.wlauto.uiauto.%1s.uiautoapk" should be "com.arm.wlauto.uiauto.%1s.apk"

marcbonnici:
Thanks for catching that. The remaining references should be removed in commit b64c615e8ec0378f9c962a3a78faad244a4a60dd

PR #403, opened 2017-06-08
Uiauto2 - Fixes

marcbonnici:


mcgeagh:
Can confirm that on my Huawei Mate 9 - Android N device, this has resolved the problem I had being unable to run applaunch before. Thanks!

PR #406, opened 2017-06-14
Replaced active_cpus by online_cpus in common/linux/device.py

AnthonyARM:
I believe `active_cpus` was removed and replaced by `online_cpus` but linux/device.py still seems to reference `active_cpus`

I'm creating the pull request with the fix now, so that our test infrastructure can pick it up and test it.
I'll confirm in a later message whether or not it fixes the issue.


AnthonyARM:
I confirm this fixes the issue

PR #407, opened 2017-06-14
[RFC] instrumentation: Add initial Instrument for BayLibre ACME

bjackman:
I hacked this up hastily by copy-pasting from LISA. I have since heard that someone else was already working on this? If so we can throw this away.

Missing features:

1. Output format same as DAQ and ARM Energy Probe instruments
2. Allow specifying path to iio-capture
3. Proper doc/description strings

The main reason I'm posting this now is to discuss point 1. It looks like WA currently doesn't have a standardised format for energy measurements, rather each instrument uses its own. Is that something that will change in WA3, with devlib's richer `Instrument` API?

### DIFF
--- a/wlauto/instrumentation/acme/__init__.py
+++ b/wlauto/instrumentation/acme/__init__.py
@@ -1,0 +1,121 @@
+import json
+import os
+import psutil
+from subprocess import Popen, PIPE, STDOUT
+import shutil
+from time import sleep
+
+from wlauto import Instrument, Parameter
+from wlauto.exceptions import HostError
+from wlauto.utils.types import list_of_strs
+from wlauto.utils.misc import which
+
+class ACME(Instrument):
+    name = 'acme'
+
+    parameters = [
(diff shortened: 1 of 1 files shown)


bjackman:
Another thought - this probably works with anything that uses IIO. Probably still prefer to call this "ACME" rather than "IIO" though since I can't even name anoth

bjackman:
Oops, accidentally closed

setrofim:
> The main reason I'm posting this now is to discuss point 1. It looks like WA currently doesn't have a standardised format for energy measurements, rather each instrument uses its own. Is that something that will change in WA3, with devlib's richer Instrument API?

Yes, that is correct -- standardized output formant is something that was added in devlib, and sol will be in WA3; WA2 doesn't have that.

setrofim:
FYI: created a  pull for the alternative implementation: https://github.com/ARM-software/workload-automation/pull/410

bjackman:
Closing, prefer #410.

PR #408, opened 2017-06-14
Updated apk workload mechanism and support for uiautomator 2 

marcbonnici:


wa/framework/workload.py
```diff
@@ -351,49 +354,76 @@ def __init__(self, owner, install_timeout=300, version=None, variant=None,
...
 
-    def init_resources(self, resolver):
-        self.apk_file = resolver.get(ApkFile(self.owner,
-                                             variant=self.variant,
-                                             version=self.version),
-                                             strict=self.strict)
-        self.apk_info = ApkInfo(self.apk_file)
-
     def setup(self, context):
+        self.resolve_package(context)
```
setrofim: This should b edone as part of `initialize` rather than `setup` as this only needs to happen once per instance per run.

wa/framework/workload.py
```diff
@@ -142,9 +143,57 @@ class ApkUiautoWorkload(ApkUIWorkload):
...
+                  description="""
+                  If ``True``, will uninstall workload\'s APK as part of teardown.'
+                  """),
+    ]
+
     def __init__(self, target, **kwargs):
         super(ApkUiautoWorkload, self).__init__(target, **kwargs)
-        self.apk = PackageHandler(self)
+        self.apk = PackageHandler(self,
+                                  package=kwargs.get('package'),
```
setrofim: these are better done as  `package=self.package` and similar, as the super's `__init__` will populate the parameters from kwargs into instance members, and handle the setting of defaults, etc.

wa/framework/workload.py
```diff
@@ -194,6 +156,76 @@ def __init__(self, target, **kwargs):
                                   force_install=kwargs.get('force_install'),
                                   install_timeout=kwargs.get('install_timeout'),
                                   uninstall=kwargs.get('uninstall'))
+
+    def init_resources(self, context):
+        pass
+
+    @once
```
setrofim: `initialize` should be used to resolve the APK/package (as only needs to happen once per run, rather than for each iteration) and so should not be done only `once` per run, but for each instance of the workload (as may specify different versions/variants).

marcbonnici:
Updated

PR #410, opened 2017-06-15
instrumetnation: add support for BayLibre ACME cape

setrofim:
Add an instrument for collecting power and energy from BayLibre
ACME cape probe.

This is an alternative implementation to:
https://github.com/ARM-software/workload-automation/pull/407 and should be
reviewed in concert.

### DIFF
--- a/wlauto/instrumentation/acmecape/__init__.py
+++ b/wlauto/instrumentation/acmecape/__init__.py
@@ -1,0 +1,130 @@
+#pylint: disable=attribute-defined-outside-init
+from __future__ import division
+import csv
+import os
+import signal
+import time
+from fcntl import fcntl, F_GETFL, F_SETFL
+from string import Template
+from subprocess import Popen, PIPE, STDOUT
+
+from wlauto import Instrument, Parameter
+from wlauto.exceptions import HostError
+from wlauto.utils.misc import which
+
+
+IIOCAP_CMD_TEMPLATE = Template("""
+${iio_capture} -n ${host} -b ${buffer_size} -c -f ${outfile} ${iio_device}
+""")
+
+
+def _read_nonblock(pipe, size=1024):
+    fd = pipe.fileno()
+    flags = fcntl(fd, F_GETFL)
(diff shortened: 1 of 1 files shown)


setrofim:
> The other one supports multiple channels but it's not actually a super useful feature and the mechanism is a bit awkward. Could always modify this to add that feature later.

I did not add that because I currently have access to only one probe, so a) I would not be able to test it, and b) I'm not sure what would be the most useful way to expose that. Unless you already have an existing use case to support multiple iio devices, lets add support for that when there is one.

I've made changes to address the other comments.

bjackman:
Sweet, this looks good to me FWIW and I'll let you know when I've tested it. Agree on the multi-channel thing.

bjackman:
Oh, also typo "instrumetnation" in the commit msg

setrofim:
fixed

bjackman:
Tested seems good. One point - for my own use case I'd like this to be the last instrument started and the first stopped, regardless of the accuracy cost to other instruments. I suppose the only way to achieve that is the edit the code to make `start` `fast_start` right? Is `fast_start` supposed to represent that this instrument starts quickly or that it's high priority?

setrofim:
>  I suppose the only way to achieve that is the edit the code to make start fast_start right?

That's the simplest way, yes. You can also manually register a method with a particular priority, e.g.
https://github.com/ARM-software/workload-automation/blob/master/wlauto/instrumentation/trace_cmd/__init__.py#L176. Which may be useful if you want a finer grained control over when something runs.

For this I'll just make them `very_fast`.

> Is fast_start supposed to represent that this instrument starts quickly or that it's high priority?

It means that the method is fast, and therefore will have smaller impact on instrumentation that is already running, and therefore WA will start it closer to the invocation of the workload i.e. if it's before the workload, e.g. start, it will be invoked later, and if it's after the workload, e.g. stop, it will be invoked sooner than the other callbacks for the same signal. Internally, this translates to a higher priority.

PR #414, opened 2017-06-20
AndroidWorkload: Uninstall uiauto apks before reinstalling

marcbonnici:
Due to the fact that uiauto apk files built on different machines will
be signed with different keys, adb will fail to overwrite a previous
version even when set to replace. This commit now will uninstall the
previous uiauto apk file if present before attempting to install the new
version.


setrofim:
Shouldn't 'replace=True' take care of that?

marcbonnici:
Apparently 'replace=True' only allows replacement of apk files if they were signed by the same key and you need to completely uninstall the apk if you want to install another version.

PR #418, opened 2017-06-21
cpustates: workaround for disabled cpuidle

setrofim:
Add "no_idle" parameter to work around the problem where cores are
reported to be in "unknown" state if there are no cpuidle transitions in
the trace. If this parameter is set set, cpustates will assume cores are
active without waiting for an idle transition.


setrofim:
Typos fixed.

PR #421, opened 2017-06-23
types: make enum levels serializable

setrofim:
Add to/from_pod implementations to level, so that it may be serialized
properly. In order to make sure that serialization is a reversible
operation, ensure that the level value is numeric.


setrofim:
Added deserialization via enum and tests

PR #422, opened 2017-06-23
Prevent the same source from being added twice (Once for 'config', on…

AnthonyARM:
…ce for 'global')

I'm not too familiar with the code so I'm not sure how the plugin_cache is supposed to work
But at the moment add_source is called once for each iteration on line 105 which triggers an exception during the second iteration.
This seems to fix the problem but I don't understand the code well enough to say if that's the proper thing to do.

### DIFF
--- a/wa/framework/configuration/parsers.py
+++ b/wa/framework/configuration/parsers.py
@@ -32,7 +32,6 @@ class ConfigParser(object):
 
     def load(self, state, raw, source, wrap_exceptions=True):  # pylint: disable=too-many-branches
         try:
-            state.plugin_cache.add_source(source)
             if 'run_name' in raw:
                 msg = '"run_name" can only be specified in the config '\
                       'section of an agenda'
@@ -102,6 +101,7 @@ class AgendaParser(object):
             raise ConfigError('Error in "{}":\n\t{}'.format(source, str(e)))
 
     def _populate_and_validate_config(self, state, raw, source):
+        state.plugin_cache.add_source(source)
         for name in ['config', 'global']:
             entry = raw.pop(name, None)
             if entry is None:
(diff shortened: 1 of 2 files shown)


setrofim:
Could you share your agenda/config? I have tried running multiple iterations however have not hit this issue.

AnthonyARM:
Here is my agenda:

```yaml
config: {device: mate9, run_name: computer_vision_validation}
global: {iterations: 1}
workloads:
- name: computer_vision
  params: {border_mode: 2, data_directory: /data/data/home,
    display_output: false, image: 640x480, num_threads: 4, options: validate_neon
      validate_cl, suffix: '', tensors: true, working_subdirectory: antbar01}
- name: computer_vision
  params: {border_mode: 0, data_directory: /data/data/home,
    display_output: false, image: 640x480, num_threads: 4, options: validate_neon
      validate_cl, suffix: '', tensors: false, working_subdirectory: antbar01}
- name: computer_vision
  params: {border_mode: 0, data_directory: /data/data/home,
    display_output: false, image: '', num_threads: 4, options: validate_neon validate_cl,
    suffix: '', tensors: false, working_subdirectory: antbar01}
- name: computer_vision
  params: {border_mode: 1, data_directory: /data/data/home,
    display_output: false, image: opticalflow, num_threads: 4, options: old_new validate_neon
      validate_cl, suffix: '', tensors: false, working_subdirectory: antbar01}
```
I believe the problem happens if you both have a `config` and `global` section in your agenda

setrofim:
I've submitted an alternative fix for this issue:

https://github.com/ARM-software/workload-automation/pull/424

AnthonyARM:
Thanks !

PR #431, opened 2017-07-04
trace-cmd: documentation fixes.

setrofim:
Following from this discussion:

https://github.com/ARM-software/workload-automation/commit/4d7ef408c97099658a340af4c9ed04154833ee69

- Remove reference to default events from the overall workload
  documentation. It was, as of recently, outdated, and was also
  redundant, as the actual defaults will be in the parameter-specific
  documentation.
- Remove reference to Android-specific trace-cmd binary -- this was not
  true for a long time.
- Clarify that the on-host trace-cmd binary is now optional due to the
  report_on_target config.
- Fix a few misc typos.


bjackman:
LGTM, FWIW

PR #454, opened 2017-08-03
AdobeReader: Updated to handle latest version

scojac01:
The workload has been updated to handle the latest version
which no longer has the welcome view page. This has been
done by adding a check and return in the dismissWelcomePage
method.


wlauto/workloads/adobereader/uiauto/app/src/main/java/com/arm/wlauto/uiauto/adobereader/UiAutomation.java
```diff
@@ -102,6 +102,10 @@ private void dismissWelcomeView() throws Exception {
         if (!onboarding_finish_button.exists()) {
             welcomeView.swipeLeft(10);
         }
+        
+        if (!onboarding_finish_button.exists()) {
+        return;
```
marcbonnici: Missing indentation.

wlauto/workloads/adobereader/uiauto/app/src/main/java/com/arm/wlauto/uiauto/adobereader/UiAutomation.java
```diff
@@ -102,6 +102,10 @@ private void dismissWelcomeView() throws Exception {
...
+        }
```
marcbonnici: I would suggest changing this to make the clicking of the UiButton conditional on its existence instead, as it would still allow the following checks to execute rather than returning early.
E.g something like:

```
if (onboarding_finish_button.exists()) {
    onboarding_finish_button.click()
}
```

scojac01: Not sure I agree on this. The method performs a lot of actions based on the presence of the onboarding button but none of the checks after this need to be executed if the button is not present?

scojac01:
I was running against 17.3.1.176959 on an Acer R11 Chromebook.

wlauto/workloads/adobereader/uiauto/app/src/main/java/com/arm/wlauto/uiauto/adobereader/UiAutomation.java
marcbonnici: From what I can tell the additional actions are for after the on-boarding button has been pressed which would be upon initial loading of the main screen and therefore may still be applicable. 
In particular the last check of the function is to ensure that the main screen has finished loading and is ready to continue with the workload which without, could cause issues on slower devices.

scojac01: That's just to make sure were on the correct screen for the rest of the workload is it not? 

Nothing in this method is key, if the welcome view is not visible. 

The OpenFile function will just fire up.  

You're the boss though.

marcbonnici: That is true. My thinking behind the alternative way was that the OpenFile function starts with the assumption that the workload is on the correct screen and therefore the purpose of the dismissWelcomeView function should be to ensure that the workload is in the known state before continuing?
And also, potentially in the future there may be additional popups which may require new checks and the early return could be misleading.

scojac01: Makes sense. Will change my implementation, drop this commit and have another to you shortly.

PR #458, opened 2017-08-07
Revent integrated

sdpenguin:
Moved the ReventWorkload class to the Linux folder with new features (see 'revent' workload below) and it has now instead been sub-classed in the Android folder. There have been no functional changes as of yet to the Android version, but the possibility exists for their implementation and is detailed in the comments.

Introduced a new 'revent' workload to playback simple revent recordings, for use with instrumentation. The possibility to specify a .close file (which runs during teardown) as well as the existent .setup revent file has been added. It is also now possible to run the workload with 'idle' set to True in order to simply wait for a specified amount of time whilst still recording results instead of running the .run revent file, whilst still being able to run .setup and .close revent files.

Finally - a fix for a bug which meant the Chromebook Plus's (and possibly other unknown devices?) device_model had a trailing null character - this is now automatically removed.

### DIFF
--- a/wlauto/common/android/workload.py
+++ b/wlauto/common/android/workload.py
@@ -16,22 +16,19 @@
 import os
 import sys
 import time
-from math import ceil
 
 from distutils.version import LooseVersion
 
 from wlauto.core.extension import Parameter, ExtensionMeta, ListCollection
 from wlauto.core.workload import Workload
-from wlauto.core.resource import NO_ONE
-from wlauto.common.android.resources import ApkFile, ReventFile
-from wlauto.common.resources import ExtensionAsset, Executable, File
+from wlauto.common.android.resources import ApkFile
+from wlauto.common.resources import ExtensionAsset, File
 from wlauto.exceptions import WorkloadError, ResourceError, DeviceError
 from wlauto.utils.android import (ApkInfo, ANDROID_NORMAL_PERMISSIONS,
                                   ANDROID_UNCHANGEABLE_PERMISSIONS, UNSUPPORTED_PACKAGES)
 from wlauto.utils.types import boolean, ParameterDict
-from wlauto.utils.revent import ReventRecording
 import wlauto.utils.statedetect as state_detector
-import wlauto.common.android.resources
+from wlauto.common.linux.workload import ReventWorkload
 
 
 DELAY = 5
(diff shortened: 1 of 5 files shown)


wlauto/common/android/workload.py
```diff
@@ -32,6 +32,7 @@
 from wlauto.utils.revent import ReventRecording
 import wlauto.utils.statedetect as state_detector
 import wlauto.common.android.resources
+from wlauto.common.linux.workload import ReventWorkload as ReventWorkloadMaster #Need to have a version of this class for Android, with fixed settings
```
setrofim: Don't do this. Instead...

wlauto/common/android/workload.py
```diff
@@ -481,83 +482,19 @@ def teardown(self, context):
...
-        self.on_device_setup_revent = None
-        self.on_device_run_revent = None
-        self.statedefs_dir = None
-
+        super(ReventWorkload, self).__init__(device, _call_super, **kwargs)
         if self.check_states:
             state_detector.check_match_state_dependencies()
+        self.idle = False #Idle mode is not currently implemented for Android
+        self.no_close = True #Currently support for .close files is not implemented for Android workloads
+        self.default_naming = True #Currently only default-named revent files generated via wa record are implemented for Android workloads
```
setrofim: erm, why? (to all three)

wlauto/common/android/workload.py
```diff
@@ -481,83 +482,19 @@ def teardown(self, context):
...
-            # pylint: disable=too-few-format-args
-            message = '{0}.run.revent file does not exist, Please provide one for your device, {0}'.format(self.device.name)
-            raise WorkloadError(message)
-
-        self.on_device_revent_binary = self.device.install_executable(revent_binary)
-        self.device.push_file(self.revent_run_file, self.on_device_run_revent)
-        self.device.push_file(self.revent_setup_file, self.on_device_setup_revent)
-
-    def _check_statedetection_files(self, context):
+    def _check_statedetection_files(self, context): #Currently Android only
```
setrofim: ...this can be moved down into `GameWorkload`

wlauto/common/linux/workload.py
```diff
@@ -0,0 +1,146 @@
...
+from wlauto.utils.revent import ReventRecording
+
+class ReventWorkload(Workload):
+    # pylint: disable=attribute-defined-outside-init
+
+    parameters = [
+        Parameter('default_naming', kind=bool, default=False,
+                  description='Use the default naming. Set this option to true if the revent file you wish to use was '
+                              'generated using the wa record command.'),
+        Parameter('idle', kind=bool, default=False,
```
setrofim: You don't need this. Just remove the default for `idle_time`. Then if it is not set (i.e. is `None`), you can assume that `idle == False`.

wlauto/common/linux/workload.py
```diff
@@ -0,0 +1,146 @@
...
+                  description='Say whether or not you want this workload to be run in idle mode. In idle mode the revent '
+                              'will only run .setup and .close files and in between will be idle for however many seconds '
+                              'you specify.'),
+        Parameter('idle_time', kind=int, default=10,
+                  description='The time you wish the device to remain idle for (idle must be True for this to be relevant).'),
+        Parameter('file_name', kind=str, default=None,
+                  description='The name of the revent file to look for. Will also look for file_name.setup and file_name.close'),
+        Parameter('no_setup', kind=bool, default=False,
+                  description='Set to True to force disable the setup revent file from running.'),
+        Parameter('no_close', kind=bool, default=False,
```
setrofim: All references to "close" should be changed to "teardown" to be consistent with existing nomenclature.

setrofim: Also, I'm not sure these should be parameters. I would assume that the requirement for setup and teardown would depend on the workload, not the user; i.e. they should be configured on per-workload, rather than per-invocation, basis.

I would create class  attributes `setup_required` and `teardown_required` which default to `True` in the base and may be overridden by individual workload definitions as needed.

wlauto/resource_getters/standard.py
```diff
@@ -89,8 +89,21 @@ def get_base_location(self, resource):
     def register(self):
         self.resolver.register(self, 'revent', GetterPriority.package)
 
-    def get(self, resource, **kwargs):
-        device_model = resource.owner.device.get_device_model()
+    def get(self, resource, **kwargs): 
+        if 'custom_name' in kwargs:
```
setrofim: Why allow custom names?

wlauto/workloads/revent/__init__.py
```diff
@@ -0,0 +1,28 @@
+from wlauto.common.linux.workload import ReventWorkload
+
+class Revent(ReventWorkload): #Skeleton class for implementation of ReventWorkload. This is needed because the Android 'ReventWorkload' also inherits from the Linux ReventWorkload
```
setrofim: Why?

wlauto/common/android/workload.py
sdpenguin: Its to ensure that the Android version of the Workload is essentially the same as it was before since the Idle mode and .close file features as well as non-default names haven't been done for Android yet, so for now I have just forced the options.

wlauto/common/android/workload.py
sdpenguin: Moved

wlauto/common/linux/workload.py
sdpenguin: OK, have now implemented so that supplying an idle_time overrides .run files with no need for an idle parameter

wlauto/workloads/revent/__init__.py
sdpenguin: Have removed the class

wlauto/common/linux/workload.py
```diff
@@ -0,0 +1,148 @@
...
+from wlauto.common.resources import Executable
+from wlauto.common.android.resources import ReventFile
+from wlauto.exceptions import WorkloadError
+from wlauto.utils.revent import ReventRecording
+
+class ReventWorkload(Workload):
+    # pylint: disable=attribute-defined-outside-init
+
+    parameters = [
+        Parameter('default_naming', kind=bool, default=False,
```
sdpenguin: I have removed it as well as support for non-default naming

wlauto/common/linux/workload.py
```diff
@@ -0,0 +1,148 @@
...
+        self.statedefs_dir = None
+
+    def initialize(self, context):
+        devpath = self.device.path
+        self.on_device_revent_binary = devpath.join(self.device.binaries_directory, 'revent')
+
+    def setup(self, context):
+        devpath = self.device.path
+        if not self.no_setup:
+            self.logger.info('Trying to get .setup')
```
sdpenguin: Have removed this from the ReventWorkload class and instead added a debug line for ReventGetter

wlauto/common/linux/workload.py
```diff
@@ -0,0 +1,148 @@
...
+    def _check_revent_files(self, context):
+        # check the revent binary
+        revent_binary = context.resolver.get(Executable(NO_ONE, self.device.abi, 'revent'))
+        if not os.path.isfile(revent_binary):
+            message = '{} does not exist. '.format(revent_binary)
+            message += 'Please build revent for your system and place it in that location'
+            raise WorkloadError(message)
+        if not self.revent_run_file and not self.idle:
+            # pylint: disable=too-few-format-args
+            message = 'It seems a {0}.run.revent or custom_name.run.revent file does not exist, Please provide one for your device: {0}.'.format(self.device.name)
```
sdpenguin: Since I've removed support for custom file names, the line has been shortened.

wlauto/resource_getters/standard.py
```diff
@@ -90,7 +90,20 @@ def register(self):
...
+                        path = os.path.join(location, candidate)
+                        try:
+                            ReventRecording(path).close()
+                            return path
+                        except ValueError as e:
+                            self.logger.warning(e.message)
+                            continue
+        device_model = resource.owner.device.get_device_model() #name format: [model/device_name.stage.revent]
         wa_device_name = resource.owner.device.name
         for name in [device_model, wa_device_name]:
```
sdpenguin: Redundant now that custom_name is removed, but thanks :1st_place_medal:

wlauto/common/android/workload.py
sdpenguin: Now only setup_required (the replacement for no_setup) is set to True and other variables are either defaulted correctly or removed

wlauto/common/linux/workload.py
```diff
@@ -0,0 +1,148 @@
...
+    def run(self, context):
+        if not self.idle:
+            self.run_timeout = self.run_timeout or self.default_run_timeout
+            command = '{} replay {}'.format(self.on_device_revent_binary, self.on_device_run_revent)
+            self.logger.debug('Replaying {}'.format(os.path.basename(self.on_device_run_revent)))
+            self.device.execute(command, timeout=self.run_timeout)
+            self.logger.debug('Replay completed.')
+        else:
+            self.logger.info('Idling for ' + str(self.idle_time) + ' seconds.')
+            time.sleep(self.idle_time)
```
sdpenguin: Have changed it to self.device.sleep()

sdpenguin:
Ready for review

wlauto/common/linux/workload.py
```diff
@@ -0,0 +1,151 @@
...
+        self.teardown_timeout = kwargs.get('teardown_timeout', None)
+        self.revent_setup_file = None
+        self.revent_run_file = None
+        self.revent_teardown_file = None
+        self.on_device_setup_revent = None
+        self.on_device_run_revent = None
+        self.on_device_teardown_revent = None
+        self.statedefs_dir = None
+        self.setup_required = False
+        self.teardown_required = False
```
setrofim: These should be class, not instance, attributes.

PR #463, opened 2017-08-11
Add Jankbench workload

bjackman:
Jankbench simulates user interaction with Android UI components and records
frame rendering times and 'jank' (rendering discontinuity) in an SQLite
database. This  is believed to be a good proxy for the smoothness of user
experience.

This is essentially implemented by copy-pasting from the Jankbench support in
LISA[1] and the GlbCorb benchmark already in WA.

Some notes:

 - Jankbench is not currently publicly available, but it is in use by multiple
   organisations.
 - This doesn't include code for analysing the result.

### DIFF
--- a/wlauto/workloads/jankbench/__init__.py
+++ b/wlauto/workloads/jankbench/__init__.py
@@ -1,0 +1,200 @@
+#    Copyright 2017 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


setrofim:
Given the two bullet points in your notes, this seems of limited use. Is this worth merging at this stage?

bjackman:
OK, closing for now

PR #464, opened 2017-08-11
revent.py: Fix handling of zero-event files

sdpenguin:
Previously the try clause worked to catch StopIteration exceptions correctly,
however upon catching the exception, another statment which set self._duration
to (last.time - first.time) was being run regardless, which defeated the point
of the try clause.

This has been fixed by introducing an else clause to contain said statement.


setrofim:
:+1:

PR #478, opened 2017-09-15
FPS instrument

setrofim:
Implement FPS instrument based on devlib, replicating the corresponding
functionality in WA2.

NOTE: This depends on devlib pull request https://github.com/ARM-software/devlib/pull/170

PR #482, opened 2017-09-19
Uibench

bjackman:
This was mainly an experiment to get started with WA3 but the result does work. Required cherry-picking two commits from https://github.com/ARM-software/workload-automation/pull/478 so can probably wait until that's merged at which point I'll rebase (UiBench is useless without gfxinfo anyway, it doesn't have any metrics of its own).

### DIFF
--- a/wa/workloads/uibench/__init__.py
+++ b/wa/workloads/uibench/__init__.py
@@ -1,0 +1,60 @@
+#    Copyright 2017 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from time import sleep
+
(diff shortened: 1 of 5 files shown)


wa/workloads/uibench/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+        'TrivialListActivity',
+        'TrivialRecyclerViewActivity',
+    ]
+
+    parameters = [
+        Parameter('activity',
+                  mandatory=True,
+                  allowed_values=test_activities,
+                  description='ID of the UiBench test activity to be run.'),
+        Parameter('duration_s', kind=int, default=10,
```
setrofim: Should be just `"duration"` -- do not suffix with units (should be documented in the description).

wa/workloads/uibench/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+        'ActivityTransition',
+        'ActivityTransitionDetails',
+        'TrivialAnimationActivity',
+        'TrivialListActivity',
+        'TrivialRecyclerViewActivity',
+    ]
+
+    parameters = [
+        Parameter('activity',
+                  mandatory=True,
```
setrofim: Mandatory parameters should be avoided unless absolutely necessary -- pick on of the available activities as the default.

bjackman:
Thanks for the review. After looking at what the various activities do, in order to pick a useful-looking one as the default, I discovered that UiBench doesn't actually do anything for most of them, it just opens the activity and then sits there (the activity I was using to test this Workload is an exception, for some reason). So I guess this is useless. IIUC there's an auxiliary package that does the actual testing, so I'll reopen this when I've figured that out...

PR #483, opened 2017-09-20
Exoplayer

bjackman:
The aim of this workload is basically to test YouTube on devices like hikey960 where we can't really get real Youtube working.

All of the stuff to get videos is basically stolen from the old 'video' workload, so if/when that gets ported I suppose we'll want to factor that out into some sort of base class? 

One thing I'm not sure about is the way I've overridden Parameters from the ApkWorkload class (package_name, and force_install) that aren't really user-definable. Not sure how else to achieve that.

There's also the problem with runtime permissions as noted in the comment  - this only works with force_install because 'pm clear' clears runtime permissions but 'pm grant' doesn't seem to give them back.

### DIFF
--- a/wa/workloads/exoplayer/__init__.py
+++ b/wa/workloads/exoplayer/__init__.py
@@ -29,9 +29,10 @@ from devlib.utils.android import grant_app_permissions
 
 # Regexps for benchmark synchronization
 REGEXPS = {
-    'start'    : '.*Displayed com.google.android.exoplayer2.demo/.PlayerActivity',
-    'duration' : '.*period \[(?P<duration>[0-9]+.*)\]',
-    'end'      : '.*state \[.+, .+, E\]'
+    'start'         : '.*Displayed com.google.android.exoplayer2.demo/.PlayerActivity',
+    'duration'      : '.*period \[(?P<duration>[0-9]+.*)\]',
+    'end'           : '.*state \[.+, .+, E\]',
+    'dropped_frames': '.*droppedFrames \[(?P<session_time>[0-9]+\.[0-9]+), (?P<count>[0-9]+)\]'
 }
 
 
@@ -57,6 +58,10 @@ class ExoPlayer(ApkWorkload):
     loading the ExoPlayer sources into Android Studio.
 
     Version r2.4.0 built from commit d979469 is known to work
+
+    Produces a metric 'exoplayer_dropped_frames' - this is the count of frames
+    that Exoplayer itself reports as dropped. This is not the same thing as the
+    dropped frames reported by gfxinfo.
     """
 
     name = 'exoplayer'
@@ -181,6 +186,18 @@ class ExoPlayer(ApkWorkload):
             self.logger.info('Waiting for playback completion ({} seconds)'
                            .format(media_duration_s))
             self.monitor.wait_for(REGEXPS['end'], timeout = media_duration_s + 30)
+
+    def update_output(self, context):
+        regex = re.compile(REGEXPS['dropped_frames'])
+
+        dropped_frames = 0
+        for line in self.monitor.get_log():
+            match = regex.match(line)
+            if match:
+                dropped_frames += int(match.group('count'))
+
+        context.add_metric('exoplayer_dropped_frames', dropped_frames,
+                           lower_is_better=True)
 
     def teardown(self, context):
         super(ExoPlayer, self).teardown(context)
--- a/wa/workloads/exoplayer/__init__.py
+++ b/wa/workloads/exoplayer/__init__.py
@@ -141,6 +141,10 @@ class ExoPlayer(ApkWorkload):
                 return files[0]
 
     def init_resources(self, context):
+        # Needs to happen first, as it sets self.format, which is required by
+        # _find_host_video_file
+        self.validate()
+
         ensure_directory_exists(self.video_directory)
         self.host_video_file = self._find_host_video_file()
 
--- a/wa/workloads/exoplayer/__init__.py
+++ b/wa/workloads/exoplayer/__init__.py
@@ -40,6 +40,7 @@ DOWNLOAD_URLS = {
     'mp4_1080p': 'http://distribution.bbb3d.renderfarming.net/video/mp4/bbb_sunflower_1080p_30fps_normal.mp4',
     'mov_720p':  'http://download.blender.org/peach/bigbuckbunny_movies/big_buck_bunny_720p_h264.mov',
     'mov_480p':  'http://download.blender.org/peach/bigbuckbunny_movies/big_buck_bunny_480p_h264.mov',
+    'ogg_128kbps': 'http://upload.wikimedia.org/wikipedia/commons/c/ca/Tchaikovsky_-_Romeo_and_Juliet_Ouverture_-_Antal_Dorati_(1959).ogg',
 }
 
 
--- a/wa/workloads/exoplayer/__init__.py
+++ b/wa/workloads/exoplayer/__init__.py
@@ -1,0 +1,187 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2017, Arm Limited and contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from collections import defaultdict
+import re
+import os
+import time
+import urllib
+
+from wa import ApkWorkload, Parameter, ConfigError, WorkloadError
+from wa.framework.configuration.core import settings
+from wa.utils.types import boolean
+from wa.utils.misc import ensure_directory_exists
+from devlib.utils.android import grant_app_permissions
+
+# Regexps for benchmark synchronization
+REGEXPS = {
+    'start'    : '.*Displayed com.google.android.exoplayer2.demo/.PlayerActivity',
+    'duration' : '.*period \[(?P<duration>[0-9]+.*)\]',
+    'end'      : '.*state \[.+, .+, E\]'
+}
+
+
+DOWNLOAD_URLS = {
+    'mp4_1080p': 'http://distribution.bbb3d.renderfarming.net/video/mp4/bbb_sunflower_1080p_30fps_normal.mp4',
+    'mov_720p':  'http://download.blender.org/peach/bigbuckbunny_movies/big_buck_bunny_720p_h264.mov',
+    'mov_480p':  'http://download.blender.org/peach/bigbuckbunny_movies/big_buck_bunny_480p_h264.mov',
+}
+
+
+class ExoPlayer(ApkWorkload):
+    """
+    Android ExoPlayer
+
+    ExoPlayer is the basic video player library that is used by the YouTube
+    android app. The aim of this workload is to test a proxy for YouTube
+    performance on targets where running the real YouTube app is not possible
+    due its dependencies.
+
+    ExoPlayer sources: https://github.com/google/ExoPlayer
+
+    The 'demo' application is used by this workload.  It can easily be built by
+    loading the ExoPlayer sources into Android Studio.
+
+    Version r2.4.0 built from commit d979469 is known to work
+    """
+
+    name = 'exoplayer'
+
+    video_directory = os.path.join(settings.dependencies_directory, name)
+
+    package_names = ['com.google.android.exoplayer2.demo']
+    versions = ['2.4.0']
+    action = 'com.google.android.exoplayer.demo.action.VIEW'
+    default_format = 'mov_720p'
+
+    parameters = [
+        Parameter('version', allowed_values=versions, default=versions[-1], override=True),
+        Parameter('duration', kind=int, default=20,
+                  description="""
+                  Playback duration of the video file. This becomes the duration of the workload.
+                  If provided must be shorter than the length of the media.
+                  """),
+        Parameter('format', allowed_values=DOWNLOAD_URLS.keys(),
+                  description="""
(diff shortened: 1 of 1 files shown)


bjackman:
cc @valschneider who wrote the LISA worlkload this is stolen from.

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+                  """.format(video_directory)),
+        Parameter('force_dependency_push', kind=boolean, default=False,
+                  description="""
+                  If true, video will always be pushed to device, regardless
+                  of whether the file is already on the device.  Default is ``False``.
+                  """),
+    ]
+
+    @property
+    def host_video_file(self):
```
setrofim: This should not be a property.  `selected_file ` shoud probably be resolved as part of `init_resource`.

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+
+    name = 'exoplayer'
+
+    video_directory = os.path.join(settings.dependencies_directory, name)
+
+    versions = ["2.4.0"]
+    action = 'com.google.android.exoplayer.demo.action.VIEW'
+
+    parameters = [
+        Parameter('package_name', default='com.google.android.exoplayer2.demo', override=True),
```
setrofim: This should be set as `package_names` class attribute rather than overriding the `package_name` parameter.

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+        Parameter('version', allowed_values=versions, default=versions[-1], override=True),
+        # The app requests runtime permissions, which AFAICT you can't grant via
+        # "pm grant", only "adb install -g". Therefore we'll just have to
+        # reinstall every time.
```
setrofim: I don't understand this. If it's already installed the permissions should already be granted, no?

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+DOWNLOAD_URLS = {
+    'mp4_1080p': 'http://distribution.bbb3d.renderfarming.net/video/mp4/bbb_sunflower_1080p_30fps_normal.mp4',
+    'mov_720p':  'http://download.blender.org/peach/bigbuckbunny_movies/big_buck_bunny_720p_h264.mov',
+    'mov_480p':  'http://download.blender.org/peach/bigbuckbunny_movies/big_buck_bunny_480p_h264.mov',
+}
+
+
+class ExoPlayer(ApkWorkload):
+    """
+    Android ExoPlayer (YouTube)
```
setrofim: Erm, why "YouTube"? This seems like a stand-alone player, playing local video files?

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+        self._selected_file = None
+
+    def enum_video_files(self):
+        for filename in os.listdir(self.video_directory):
+            for format in self.parameters['format'].allowed_values:
+                if format in filename:
+                    self.video_files[format].append(os.path.join(self.video_directory, filename))
+
+    def validate(self):
+        if (self.format and self.filename) and (self.format != self.parameters['format'].default):
```
setrofim: What if `format` explicitly set to the default value along side a filename in the config? Should default it to `None` and then set it to default here after the first check.

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+        grant_app_permissions(self.target, self.package)
+
+    def setup(self, context):
+        self.device_video_file = self.target.path.join(self.target.working_directory,
+                                                       os.path.basename(self.host_video_file))
+        if self.force_dependency_push or not self.target.file_exists(self.device_video_file):
+            self.logger.debug('Copying {} to device.'.format(self.host_video_file))
+            self.target.push(self.host_video_file, self.device_video_file, timeout=120)
+
+        super(ExoPlayer, self).setup(context)
```
setrofim: Is there a reason this needs to happen after pushing the video file? If not, the super call generally be the first thing in a method.

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # Prepare logcat monitor
+        self.monitor = self.target.get_logcat_monitor(REGEXPS.values())
+        self.monitor.start()
+
+    def run(self, context):
```
setrofim: `run` should be as small as possible (instrumentation is already running by this point. The command construction should happen in the setup and result processing /extraction in the corresponding methods.

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+        play_cmd = 'am start -a {} -d "file://{}"'.format(self.action, self.device_video_file)
+        self.logger.info(play_cmd)
```
setrofim: This should not be an info-level log; probably does not need to be logged at all, as this will be logged as part of `execute()`

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+        self.monitor.wait_for(REGEXPS['start'])
+        self.logger.info('Playing media file')
+
+        line = self.monitor.wait_for(REGEXPS['duration'])[0]
+        media_duration_s = int(round(float(re.search(REGEXPS['duration'], line)
+                                           .group('duration'))))
+
+        self.logger.info('Media duration is {} seconds'.format(media_duration_s))
+
+        if self.play_duration and self.play_duration < media_duration_s:
```
setrofim: I don't understand this. Why is `play_duration` necessary at all if you're actually waiting for the video to finish?

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+            self.logger.info('Waiting {} seconds before ending playback'
+                           .format(self.play_duration))
+            time.sleep(self.play_duration)
+        else:
+            self.logger.info('Waiting for playback completion ({} seconds)'
+                           .format(media_duration_s))
+            self.monitor.wait_for(REGEXPS['end'], timeout = media_duration_s + 30)
+
+        self.logger.info('Media file playback completed')
```
setrofim: This is unnecessary -- workload completion would already be indicated in the log. This is also misleading, since if you've hit the `if` condition above, you may have ended up sleeping for some time by this point, after media playback has completed

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def teardown(self, context):
+        self.monitor.stop()
+        super(ExoPlayer, self).teardown(context)
```
setrofim: Again, the `super` should generally be the first thing called, unless there is a specific reason something needs ot happen first. I would also stop the monitor in `extract_results` as that happens much sooner.

wa/workloads/exoplayer/__init__.py
bjackman: IIUC Android has a new 'runtime permissions' thing which is separate from the permissions listed in the .apk manifest - those are the ones where the user gets a dialog box at the moment where the app actually needs the permission. I can't find a way to automatically grant them except with `adb install -g`. It probably does bear some more investigation actually.

wa/workloads/exoplayer/__init__.py
bjackman: Ah yeah I'll reword this. It's the player that the Youtube app uses internally, so we're hoping it's a good proxy for youtube performance.

wa/workloads/exoplayer/__init__.py
bjackman: No reason, will move it

wa/workloads/exoplayer/__init__.py
bjackman: Oops, yep this is debug leakage.

wa/workloads/exoplayer/__init__.py
bjackman: So then we'd start the instrumentation _after_ the video has begun to play? That means depending on how long the instrumentation takes to set up we'd be measuring playback of a different section of video.

wa/workloads/exoplayer/__init__.py
bjackman: Hmm true. I just copied this bit from the 'video' workload, will change it.

wa/workloads/exoplayer/__init__.py
setrofim: Hm IIRC, @marcbonnici  ran into this in the past. Can't remember what the outcome was..

wa/workloads/exoplayer/__init__.py
setrofim: I see. It's worth explaining that in the description, but I would remove the YouTube reference from the first line.

wa/workloads/exoplayer/__init__.py
setrofim: I meant the construction of the `play_cmd`. The invocation would still happen here. I realize that the string form will not have a measure able impact, it's more of a conceptual thing...

wa/workloads/exoplayer/__init__.py
bjackman: We only wait for the video to finish if `play_duration` is unspecified or the video is longer than that.

wa/workloads/exoplayer/__init__.py
bjackman: OK. Does extract_results still happen if the workload fails? We need to be sure to kill the monitor or WA won't die.

wa/workloads/exoplayer/__init__.py
setrofim: Right, but I don't  understand why. If `play_duration`  is specified, there are three scenarios:

- `play_duration` == playback duration. `play_duration` is irrelevant.
- `play_duration` < playback duration. `play_duration` is ignored;  `run` returns only actual playback is finished. In this case, the configuration is actively misleading.
- `play_duration` > playback duration. `run` sleeps doing nothing for an arbitrary period of time. Collected measurements are now off  since what was being measured is not actually video playback but the app just sitting there (essentially idle). Here again the setting is misleading as it does not specify the duration of play.

Am I missing something?

wa/workloads/exoplayer/__init__.py
bjackman: Oh OK, will change that.

wa/workloads/exoplayer/__init__.py
setrofim: Good point. It might not. You are right, `teardown` is the correct place for this.

wa/workloads/exoplayer/__init__.py
bjackman: Oops, when I said 

> We only wait for the video to finish if play_duration is unspecified or the video is longer than that.

I meant if the video is _shorter_ than that. So, the cases are

- `play_duration` == playback duration: irrelevant, yes
- `play_duration` < playback duration: Then we record playing `play_duration` seconds of video (the video is playing while we `time.sleep`).
- `play_duration` > playback duration: Then we record playing the whole video. You are correct that the Parameter is misleading in this case - I think I'll just add a note about it to the description

setrofim: OK, that makes more sense. I also misunderstood how the code worked. From your updated description, the matching of duration returns near the beginning of the workload and does not indicate that playback has finished?

In that case it would make sense; however I would raise an error to force user to adjust the config. (even better, can we query the video file on the host as part of `init_resources`, and raise an error *then* if the configured duration exceeds the video?

Also, please rename to just `duration` to be consistent with other workloads that specify the same parameter.

bjackman:
I've updated the commit but there are a couple more issues to iron out - I'll get back to you once it's ready for a second round of review.

> Given the similarity between workloads, would it make more sense fore the player to be a configurable parameter? (note: I'm not suggesting you add support for the built-in player as part of this pull request)

Hmm, perhaps, although I can imagine that the different players could end up with different sets of tunable parameters...

setrofim:
> Hmm, perhaps, although I can imagine that the different players could end up with different sets of tunable parameters... 

Yes, true; also, given that in `run` you're parsing output specific to this player... OK, let's keep this `exoplayer` for now. I'll revisit this when I get to porting `video` to WA3; maybe it'd make more sense to keep  them separate and split out common functionality into a utility class...

wa/workloads/exoplayer/__init__.py
bjackman: OK, will rename and have it print an error if the video is shorter than `duration` seconds. I won't bother looking at the file on the host - I don't think it would be worth the complexity.

setrofim: I would raise a `ConfigError` rather than just log an error-level message. This is something that needs to be rectified by the user; and we don't want to generate results if the config is wrong.

bjackman: OK

bjackman:
OK I think I've ironed out the issues now. Two points remain

- the `validate` vs `init_resources` thing we discussed. I was thinking the solution might be to say `validate` comes before `init_resources`, but if you need the resources in `validate` you are free to call `self.init_resources` yourself?

- I've put a `grant_app_permissions` call in `setup`. Do you think that belongs in the superclass?

wa/workloads/exoplayer/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+from wa.utils.types import boolean
+from wa.utils.misc import ensure_directory_exists
+from devlib.utils.android import grant_app_permissions
+
+# Regexps for benchmark synchronization
+REGEXPS = {
+    'start'         : '.*Displayed com.google.android.exoplayer2.demo/.PlayerActivity',
+    'duration'      : '.*period \[(?P<duration>[0-9]+.*)\]',
+    'end'           : '.*state \[.+, .+, E\]',
+    'dropped_frames': '.*droppedFrames \[(?P<session_time>[0-9]+.[0-9]+), (?P<count>[0-9]+)\]'
```
bjackman: > `[0-9]+.[0-9]`

should be `[0-9]+\.[0-9]`

bjackman:
Forgot to mention - I worked around the `validate` vs `init_resources` thing but just having the latter call the former, with a comment. I think this is actually a pretty acceptable solution - the workload knows about dependencies and idempotency of those methods.

PR #491, opened 2017-09-29
BaseUiAutomation: add Direction and PinchType enums

bjackman:
These are required by the Geekbench workload which is to be ported from WA2

---

This workload is currently out-of-tree but I'm hoping to submit it in the next few weeks, and having this merged will allow colleagues to use it in the interim.


bjackman:
Hmm, I submitted this without thinking about it but just realised I have no idea why the enums are needed - better hold off on this for a mo

bjackman:
Hmm yeah I can still build the Geekbench workload with this patch reverted.

PR #503, opened 2017-10-06
energy instrument backends parameter validation fix

setrofim:


wa/instrumentation/energy_measurement.py
```diff
@@ -157,6 +157,7 @@ class AcmeCapeBackend(EnergyInstrumentBackend):
                   Host name (or IP address) of the ACME cape board.
                   """),
         Parameter('iio-device', default='iio:device0',
+                  kind=list_or_string,
```
bjackman: I don't think this belongs in this commit - my 'multiple channels' commit will add it.

setrofim: Ah crap. yeah, I was just using that to test. now removed.

PR #504, opened 2017-10-06
framework/target: fix runtime param name resolution

setrofim:
Runtime param names from config were matched against names available on
for the target using an overly-loose regex that resulted in the wrong
config point being returned.

Use caseless_string matching instead.


bjackman:
Not equipped to provide input on the code, but I've tested this and it looks good

!(thumbs up)[https://media.giphy.com/media/XreQmk7ETCak0/giphy.gif]

PR #509, opened 2017-10-10
Fixes

setrofim:
Energy instrument:

- Convert parameter keys to identifiers before applying
- Raise an error if passed a param invalid for a backend.

Other:

- Fix APK resolution with HTTP getter


bjackman:
Not sure if I mentioned, I've tested this and it LGTM.

PR #510, opened 2017-10-10
trace-cmd: Disable creation of trace.txt by default

bjackman:
I don't know of any use for the trace.txt produced by this option,
and it is responsible for most of WA's disk usage. Furthermore it
happens to clash with other tools that have specific requirements for
the format of trace.txt. So I propose disabling it by default.

### DIFF
--- a/wa/instrumentation/trace-cmd.py
+++ b/wa/instrumentation/trace-cmd.py
@@ -136,7 +136,7 @@ class TraceCmdInstrument(Instrument):
                   form the buffer size until set succeeds or size is reduced to
                   1MB.
                   """),
-        Parameter('report', kind=bool, default=True,
+        Parameter('report', kind=bool, default=False,
                   description="""
                   Specifies whether reporting should be performed once the
                   binary trace has been generated.


setrofim:
Most of the analysis scripts and notebooks (e.g. WA's cpustates processor, among other things ) will parse trace.txt. That's what we want most of the time. The only use I know of the binary trace, on the other hand, is kernelshark.

Could you elaborate on what other tools trace.txt clashes with?

(Also note that rather than modifying the default here, you can always set "report" to false in your main user config, so that it will be automatically picked up for your runs).

derkling:
Main points for me are:
1. WA (at least version 3) should be considered mainly as a data collection tool
2. post-processing is eventually done by other tools

From what you are saying, it seems that we are not aligned on that, and that WA3 still aims at be the only tool used to collect and post-process everything.

Apart from that observation,  a more pragmatic observation is that a trace.txt is not a standard format.
The standard format is trace.dat and you can generate many different output from that (raw, formatted, etc...).
If you collect a trace from a target and want to share your data with someone else to use whatever tools they wanna use to post-process these trace data, you don't need a trace.txt. You definitively want the trace.dat binary format. Since you cannot avoid to share the trace.dat, which is the only true standard format, then you will always end up with wa_output folders which contains duplicated trace data.
Consider that a reported trace.txt can be also way bigger then the corresponding trace.dat.

All that considered, I would say that, if WA3 for whatever reasons it needs a trace.txt to support some prepossessing (which should likely not be there) then it can certainly generate a trace.txt, do its processing and then remove the file.

If the user explicitly ask for a trace report, then WA3 should also provide a configuration parameter which allows the user to specify how the trace.txt should be generated, i.e. which trace-cmd parameters should be used.

To conclude, as it is right now, WA is just hardcoding the generation of a trace.txt report which is not standard at all just for some (most of the times not required) possible post-processing (which should not be part of WA).

setrofim:
> WA (at least version 3) should be considered mainly as a data collection tool

WA version 3 will supersede WA version 2, and is intended as a complete replacement for it. As such, it will broadly match the functionality available in WA version 2 (there a a few plugins in v2 that are now stale in are probably not worth porting).

WA *is* intended as a data gathering tool; and is it *not* intended as full-on analysis suite. However, WA supports, and will continue to support way of doing routine processing on the raw data it gathers. Things like derivation (e.g. calculating fps from frame data), compression, format conversion, summary statistics generation, database export, etc. have been and will continue to be available as configurable options.

Some of WA's users are not programmers (or not that comfortable with high-level scripting) and just want something the can paste directly into a spreadsheet without having to write additional scripts. 

> post-processing is eventually done by other tools

Yes, that's true. My point was that, in my experience, most of those tools will be parsing the text rather than binary trace. So 8/10 times, you want to report the binary trace before doing anything else; hence it is the default.

>   WA3 still aims at be the only tool used to collect and post-process everything.

Not at all. Just because WA provides an option to post-process something, does not mean you *have* to use. The option can can always be disabled, and WA will always make raw output available.

> Apart from that observation, a more pragmatic observation is that a trace.txt is not a standard format.
The standard format is trace.dat and you can generate many different output from that (raw, formatted, etc...).

This is the opposite of what I have been told before. When I was implementing a generic trace parser to be used in scripts/notebooks,  I have investigated whether I should try to parse the trace.dat or parse trace.txt. My thinking at the time was exactly that of avoiding generating trace.txt's unnecessarily.

What I was told at the time was that I should definitely stick to parsing text trace. Apparently, the binary format is not well defined, is not compatible between versions, and is not intended for "public  consumption". Different versions of `trace-cmd`, even, may not necessarily be able to report each other's traces. Where as the text format, on the other hand, is a lot more stable between versions, and scripts relying on it, rather than the binary trace, are a lot more likely to continue to work without needing to be constantly updated.

Are you now saying the above isn't true, and the binary format is the standard that should parsed directly, rather than the text trace? If so, please could you point me to the format spec, or a parser implementation I could use as guide? I would love to remove the need to generate  text trace all the time.

> If the user explicitly ask for a trace report, then WA3 should also provide a configuration parameter which allows the user to specify how the trace.txt should be generated, i.e. which trace-cmd parameters should be used.

This is exactly the case now.

>as it is right now, WA is just hardcoding the generation of a trace.txt report 

This is not true. It is a configurable parameter, not hard-coded.

The issue, and what is being addressed in this pull, is that the default for that parameter is currently to generate the trace. The reason is is currently that way is that, my experience is, text trace is wanted by most users most of the time. If you don't want it, it's a single configuration entry in your environment.

bjackman:
I didn't realise trace.txt was parsed by tools in WA. So, closing this.

PR #511, opened 2017-10-10
energy_measurement: Ensure 'standard' energy metrics are produced

bjackman:
Where backends have the capability to collect from multiple devlib
Instruments, EnergyMeasurement currently appends 'device' keys to the
names of metrics reported by those Instruments. Where multiple
Instruments are indeed used, it then sums up equivalent metrics to
produce the 'normal' metric without the 'device' keys
appended.

E.g. If you have two ACME 'devices' enabled you will get something
like
'device_total_energy_iio:device0' and
'device_total_energy_iio:device1', which will be summed to produce
'device_total_energy'.

However when only one Instrument is actually instantiated, this is
not needed and not done. Therefore just directly produce the metric
with the name originally given by devlib (i.e. in the example,
'device_total_energy').

### DIFF
--- a/wa/instrumentation/energy_measurement.py
+++ b/wa/instrumentation/energy_measurement.py
@@ -290,7 +290,7 @@ class EnergyMeasurement(Instrument):
             # Append the device key to the filename and artifact name, unless
             # it's None (as it will be for backends with only 1
             # devce/instrument)
-            if device is not None:
+            len(self.instruments) > 1:
                 name = 'energy_instrument_output_{}'.format(device)
             else:
                 name = 'energy_instrument_output'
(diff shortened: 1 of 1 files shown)


setrofim:
Perhaps this is an argument for specifying the device as a classifier, rather than formatting it into the name?

bjackman:
Yeah that may be true - I think classifiers are probably a cleaner theoretical representation for this, I just didn't go with relying on them because I'm worried about the usability cost. I'll do some fiddling and see how it ends up looking for the user

setrofim:
Usability cost?  Would they not be more usable, as they will already appear a separate entry in the results, so you don't need to manually split them out afterwards?

bjackman:
Well, they appear as a separate entry (one per device + one summed one) with the current implementation, but because the classifiers don't appear by default in results.csv, you don't see which is which without checking the JSON (or using the proper WA parsing lib, which I haven't got round to...)

setrofim:
Fair enough. I guess there is no harm in merging this for now.

PR #512, opened 2017-10-11
WA3 dmesg instrument

bjackman:
Maybe this belongs in devlib? It's pretty trivial so fine here IMO.


setrofim:
> Maybe this belongs in devlib? It's pretty trivial so fine here IMO.

Yup agreed, should be trivial to port to devlib if necessary.

Since the second commit is modifying a new file being added by the first, it should probably be collapsed into it. Apart from that, LGTM.

bjackman:
OK, squashing

PR #515, opened 2017-10-12
hwmon: Add hwmon instrument

bjackman:
For some reason I didn't look at the WA2 instrument until afer I
wrote this, so it's different in how it handles 'before/after'
quantities (which I called 'momentary' in the code) - instead of
adding a 'mean' it just adds two separate metrics: one for before and
one for after. Personally I think this is more useful (esp. for
temperature) but if necessary I can revert to the old behaviour.

This also doesn't add the helpful classifier telling you the hwmon
device, because we don't have immediate access to that info from the
devlib Instrument. If necessary I'm sure it could be added later.

### DIFF
--- a/wa/instrumentation/hwmon.py
+++ b/wa/instrumentation/hwmon.py
@@ -1,0 +1,88 @@
+#    Copyright 2017 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from devlib import HwmonInstrument as _Instrument
+
+from wa import Instrument
+from wa.framework.instrumentation import fast
+
+MOMENTARY_QUANTITIES = ['temperature', 'power', 'voltage', 'current', 'fps']
+CUMULATIVE_QUANTITIES = ['energy', 'tx', 'tx/rx', 'frames']
+
+class HwmonInstrument(Instrument):
+    name = 'hwmon'
+
+    description = """
+    Hardware Monitor (hwmon) is a generic Linux kernel subsystem,
+    providing access to hardware monitoring components like temperature or
+    voltage/current sensors.
+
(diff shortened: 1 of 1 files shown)


wa/instrumentation/hwmon.py
```diff
@@ -0,0 +1,77 @@
...
+CUMULATIVE_QUANTITIES = ['energy', 'tx', 'tx/rx', 'frames']
+
+class HwmonInstrument(Instrument):
+    name = 'hwmon'
+
+    description = """
+    Hardware Monitor (hwmon) is a generic Linux kernel subsystem,
+    providing access to hardware monitoring components like temperature or
+    voltage/current sensors.
+    """
```
setrofim: Would be good to expand the description with a second paragraph outlining what this instrument actually reports (i.e. before/after for "momentary" and diff for the "continuous" metrics.

wa/instrumentation/hwmon.py
```diff
@@ -0,0 +1,77 @@
...
+        self.instrument = _Instrument(self.target)
+
+    def setup(self, context):
+        self.instrument.reset()
+
+    @fast
+    def start(self, context):
+        self.before = self.instrument.take_measurement()
+
+    def stop(self, context):
```
setrofim: Should be `@fast` as well?

bjackman: Oooo yes, could this also explain why we've been seeing ftrace stopped before the energy instrumentation?

I was planning to investigate that, I assumed that instrumentation should be `stop`ped in the reverse of the order it was `start`ed in...

setrofim: Yes, for the stages after `run()`, the order should be reversed; the priority decorators will take care of of that. `@fast` applied to `stop()` will mean it will before slower stop methods (i.e. closer to the stage it's wrapping).

wa/instrumentation/hwmon.py
```diff
@@ -0,0 +1,88 @@
...
+    hwmon data will be reported.
+    """
+
+    def initialize(self, context):
+        self.instrument = _Instrument(self.target)
+
+    def setup(self, context):
+        self.instrument.reset()
+
+    @fast
```
derkling: Since Energy measurements start/stop should be the really first thing we start/stop across the execution of a workload... should not be safer to set these callbacks as @very_fast level?

setrofim: "@fast" already means this will run closer to the measured region than most other instrumentation. Bear in mind, that the decorator is called that for a reason -- it implies that the method it decorates is actually fast to execute and will not interfere much with the instrumentation that is already running. Given that take_measurement() for hwmon will result in a call to the target for each active channel, this definitely does not qualify as "@very_fast" (if anything, I would argue that "@fast" is a stretch...).

PR #518, opened 2017-10-17
Improve implementation of 'instrumentation' job spec field

bjackman:
This is basically to allow this:
```
workloads: 
- name: foo     # Run a few iterations with ftrace enabled, for deep debugging
  iterations: 3
- name: foo     # Run a load more iterations with ftrace disabled, so we don't fill up our filers
  iterations: 30
  instrumentation: [~trace-cmd]
```

I'm not sure if this implementation is quite right, in particular it doesn't allow _enabling_ an instrument specifically for a job spec (that currently ends up enabling it for them all) and I'm not sure the cleanest way to do that.. so feedback welcome.

### DIFF
--- a/wa/framework/configuration/core.py
+++ b/wa/framework/configuration/core.py
@@ -1023,7 +1023,7 @@ class JobGenerator(object):
         self.ids_to_run = []
         self.sections = []
         self.workloads = []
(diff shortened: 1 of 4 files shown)


setrofim:
Closing as this functionality is now supported on `next`

PR #520, opened 2017-10-19
framework/execution: Add event when skipping job due to phones_home

bjackman:


setrofim:
This is an alternative that ensures that all errors and warnings from some loggers (Runner's, workloads', insturments', and processors') are added as events.

bjackman:
Cool, that seems better.

PR #521, opened 2017-10-19
 framework: Add bail_on_init_failure run configuration 

bjackman:
If a workload fails in `setup` or `run`, we continue with the run. This is an
attempt to have similar behaviour for failure in `initialize`.

If a workload fails in `initialize`, we SKIP all the jobs with the same
ID (i.e. same job spec) under the assumption that their `initialize` will fail
too.

-----

Marked this [RFC] because I don't really know if this is desirable behaviour (seems helpful to me, but maybe I've misunderstood the model), and I'm not sure if I've implemented it right (because Runner has to mutate ExecutionState by overwriting its job_queue)

### DIFF
--- a/wa/framework/configuration/core.py
+++ b/wa/framework/configuration/core.py
@@ -780,6 +780,19 @@ class RunConfiguration(Configuration):
             ''',
         ),
         ConfigurationPoint(
+            'bail_on_init_failure',
+            kind=bool,
+            default=True,
+            description='''
+            When jobs fail during their main setup and run phases, WA will
+            continue attempting to run the remaining jobs. However, by default,
+            if they fail during their early initialization phase, the entire run
+            will end without continuing to run jobs. Setting this to ``False``
+            means that WA will instead skip all the jobs from the job spec that
+            failed, but continue attempting to run others.
+            '''
+        ),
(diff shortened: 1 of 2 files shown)


setrofim:
The fact that a failed `initialize()` kills the run is intentional. The reasoning being, since all initialization happens up front, before the run actually starts, it is better to fail early and let the user deal with the issue (either by addressing the cause of the failure, or by explicitly excluding the offending workload from the agenda).

I can see this change being useful in non-interactive environments though (where WA is invoked as part of some infrastructure, rather than by the user directly). Perhaps this behavior should be controlled by a configuration setting.

In terms of implementation, I agree that having the Runner mutate the ExecutionState is messy. I think a way around that would be to move the job initialization loop inside the ExecutionContext, so that instead of

```
       for job in self.context.job_queue:
             # initialize the job
```

the Runner would just  do

```
      self.context.initialize_jobs()
```

bjackman:
That all sounds pretty reasonable, I'll see if I can implement it.

bjackman:
OK I pushed a version with a config setting and a bit of a cleaner implementation (thanks for the suggestion, much simpler than the direction I was going!)

PR #523, opened 2017-10-23
Geekbench: Updating workload to work with GB4 onwards irrespective of…

scojac01:
… minor version. Previous version support has also been removed


setrofim:
From https://github.com/ARM-software/workload-automation/issues/531, it seems version 3 is still actively used, so support for it cannot be removed.

PR #524, opened 2017-10-23
Logger events

setrofim:
This adds error- and warning-level messages from selected loggers to the output events.


bjackman:
This looks good to me. I had a look to see if there were any explicit calls to `add_event ` that could be dropped now, but they all seem to be next to calls to the `log_error` function rather than the `logging.error`, so seems good.

PR #526, opened 2017-10-23
Fix logging error when exception raised

bjackman:


bjackman:
Ummm... should the `str` call actually just be in the `add_event`? Or the caller of `add_event`? Hopefully the fact that I don't know the answer means it doesn't really matter.

PR #537, opened 2017-11-06
Csvproc fix

bjackman:
We currently populate results_so_far with a JobOutput for each Job and then a
Result for the RunOutput. This results in a bug when trying to access the
id/label/iteration.

This is fixed by always ensuring the we store Output objects and not
Results (results_so_far is renamed to outputs_so_far to reflect this), and
treating the RunOutput specially in _write_outputs.

### DIFF
--- a/wa/processors/csvproc.py
+++ b/wa/processors/csvproc.py
@@ -42,28 +42,28 @@ class CsvReportProcessor(ResultProcessor):
             raise ConfigError(msg)
 
     def initialize(self):
-        self.results_so_far = []  # pylint: disable=attribute-defined-outside-init
+        self.outputs_so_far = []  # pylint: disable=attribute-defined-outside-init
         self.artifact_added = False
 
     def process_job_output(self, output, target_info, run_output):
-        self.results_so_far.append(output)
-        self._write_results(self.results_so_far, run_output)
+        self.outputs_so_far.append(output)
+        self._write_outputs(self.outputs_so_far, run_output)
         if not self.artifact_added:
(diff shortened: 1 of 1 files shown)


wa/processors/csvproc.py
```diff
@@ -77,8 +78,15 @@ def _write_results(self, results, output):
             writer.writerow(['id', 'workload', 'iteration', 'metric', ] +
                             extra_columns + ['value', 'units'])
 
-            for o in results:
-                header = [o.id, o.label, o.iteration]
+            for o in outputs:
+                if isinstance(o, JobOutput):
```
setrofim: this could just be `if o.kind == 'job':`

bjackman: Updated this. I also added an explicit check for kind=='job', just because it feels more rigorous.

wa/processors/csvproc.py
```diff
@@ -77,8 +77,18 @@ def _write_results(self, results, output):
...
+                if o.kind == 'job':
+                    header = [o.id, o.label, o.iteration]
+                elif o.kind == 'run':
+                    # Should be a RunOutput. Run-level metrics aren't attached
+                    # to any job so we leave 'id' and 'iteration' blank, and use
+                    # the run name for the 'label' field.
+                    header = [None, o.info.run_name, None]
+                else:
+                    raise RuntimeError(
+                        'Output of kind "{}" unrecognised by csvproc'.format(kind))
```
marcbonnici: `kind` should be `o.kind`

bjackman: Oh, good spot, thanks.

PR #539, opened 2017-11-14
Porting UxPerf Workloads

marcbonnici:


wa/framework/uiauto/app/src/main/java/com/arm/wa/uiauto/BaseUiAutomation.java
```diff
@@ -173,6 +215,108 @@ public int getDisplayCentreHeight() {
...
+                view.swipeRight(steps);
+                break;
+            case NULL:
+                throw new Exception("No direction specified");
+            default:
+                break;
+        }
+    }
+
+        public void uiObjectVertPinchIn(UiObject view, int steps, int percent) throws Exception {
```
setrofim: Indentation level seems to change from this line onwards.

wa/framework/workload.py
```diff
@@ -218,7 +229,25 @@ def teardown(self, context):
...
-        pass
+        self.remove_assets(context)
+
+    def deploy_assets(self, context, target_directory=None):
+        if not self.deployable_assets:
+            return
+
+        if not target_directory:
+            target_directory = self.target.working_directory
+        self.target.execute('mkdir -p {}'.format(target_directory))
```
setrofim: Why is this necessary? Shouldn't the working_directory already exist by the time we get to initializing workloads?

wa/framework/workload.py
```diff
@@ -218,7 +229,25 @@ def teardown(self, context):
...
+
+        for i, asset in enumerate(self.deployable_assets):
+            self.target.push(self.asset_files[i], target_directory)
+            self.deployed_assets.append(self.target.path.join(target_directory,
```
setrofim: The join could be replaced with `self.target.get_workpath(asset)`

wa/framework/uiauto/app/src/main/java/com/arm/wa/uiauto/BaseUiAutomation.java
marcbonnici: I can't seem to see the change in indention. Has this been fixed?

wa/framework/workload.py
marcbonnici: The idea here was that you could set a custom directory for the assets to be deployed to instead of the working directory so the custom folder may not have existed. However I've re-done this bit to be part of the base Workload Class and to use an attribute to store the directory path instead.

wa/framework/workload.py
marcbonnici: This was left as join to allow for custom directories.

wa/framework/uiauto/app/src/main/java/com/arm/wa/uiauto/BaseUiAutomation.java
setrofim: yup, seems to be fixed now. :+1:

PR #541, opened 2017-11-16
framework/target: wait 20 sec for network to show up

qperret:
Some targets need more than 5 seconds to re-connect to a wifi network
after disabling airplane mode. Change the delay to 20 seconds to make
that more reliable.


qperret:
Agreed. That was just me being lazy ...
That should be better now.

PR #542, opened 2017-11-21
Fixes for uxperf workloads

setrofim:


wa/framework/workload.py
```diff
@@ -66,8 +66,11 @@ class Workload(TargetedPlugin):
 
     # Used to store information about workload assets.
     deployable_assets = []
```
marcbonnici: `deployable_assets` should also be moved into `__init__.py`

marcbonnici: Ignore me.

PR #543, opened 2017-11-23
Fix package resolution

marcbonnici:
Mandate that either a package name is specified or the owner defines
supported package names for use during package resolution.

### DIFF
--- a/wa/framework/workload.py
+++ b/wa/framework/workload.py
@@ -562,13 +562,29 @@ class PackageHandler(object):
             msg = 'Cannot Resolve package; No package name(s) specified'
             raise WorkloadError(msg)
 
-        self.apk_file = context.resolver.get(ApkFile(self.owner,
-                                                     variant=self.variant,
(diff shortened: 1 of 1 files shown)


wa/framework/workload.py
```diff
@@ -562,13 +562,29 @@ def resolve_package(self, context):
...
+                                                                       version=self.version,
+                                                                       package=package,
+                                                                       exact_abi=self.exact_abi,
+                                                                       supported_abi=self.supported_abi),
+                                                               strict=self.strict))
+            if len(available_packages) == 1:
+                self.apk_file = available_packages[0]
+            elif len(available_packages) > 1:
+                msg = 'Multiple matching packages found on host'
+                raise WorkloadError(msg.format(self.owner))
```
setrofim: Would be good to include paths to the found APKs in the error message.

marcbonnici: Updated.

PR #548, opened 2017-11-27
util/android, get_apk_versions: try to find 'aapt' in $PATH as well

koukaipan:
Some Linux distro provide android build-tools in packages, we should also try
to find 'aapt' in $PATH if it cannot be found in $ANDROID_HOME.

### DIFF
--- a/dev_scripts/get_apk_versions
+++ b/dev_scripts/get_apk_versions
@@ -10,22 +10,29 @@ sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
 
 from wlauto.exceptions import WAError, ToolError
 from wlauto.utils.doc import format_simple_table
+from wlauto.utils.misc import check_output, get_null, which
 
 
 def get_aapt_path():
     """Return the full path to aapt tool."""
     sdk_path = os.getenv('ANDROID_HOME')
-    if not sdk_path:
-        raise ToolError('Please make sure you have Android SDK installed and have ANDROID_HOME set.')
-    build_tools_directory = os.path.join(sdk_path, 'build-tools')
-    versions = os.listdir(build_tools_directory)
-    for version in reversed(sorted(versions)):
-        aapt_path = os.path.join(build_tools_directory, version, 'aapt')
-        if os.path.isfile(aapt_path):
-            logging.debug('Found aapt for version {}'.format(version))
-            return aapt_path
+    if sdk_path:
+        build_tools_directory = os.path.join(sdk_path, 'build-tools')
+        versions = os.listdir(build_tools_directory)
+        for version in reversed(sorted(versions)):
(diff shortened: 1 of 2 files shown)


dev_scripts/get_apk_versions
```diff
@@ -10,22 +10,31 @@ sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
 
 from wlauto.exceptions import WAError, ToolError
 from wlauto.utils.doc import format_simple_table
+from wlauto.utils.misc import (check_output, get_null)
```
setrofim: Parens are unnecessary here.

dev_scripts/get_apk_versions
```diff
@@ -10,22 +10,31 @@ sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
...
+        else:
+            raise ToolError('aapt not found. Please make sure at least one Android platform is installed.')
     else:
-        raise ToolError('aapt not found. Please make sure at least one Android platform is installed.')
+        logging.debug("ANDROID_HOME is not set, try to find in $PATH.")
+        if subprocess.call('aapt version >{} 2>&1'.format(get_null()), shell=True) == 0:
+            aapt_path = os.path.dirname(subprocess.check_output('which aapt', shell=True))
+            aapt_path = os.path.join(aapt_path, 'aapt')
+            logging.debug('Using aapt from {}'.format(aapt_path))
+            return aapt_path
```
setrofim: You should use 'which()' here, rather than using OS ultils directly:

https://github.com/ARM-software/workload-automation/blob/master/wlauto/utils/misc.py#L605

This will also eliminate the need to call `aapt version` beforehand. This if statement can be replaced with
```
aapt_path = which('aapt')
if aapt_path:
    return aapt_path
raise ToolError('Please make sure you have Android SDK installed and have ANDROID_HOME set.')
```

wlauto/utils/android.py
```diff
@@ -458,19 +458,26 @@ def _initialize_without_android_home(env):
...
+                env.aapt = aapt_path
+                break
+        else:
+            raise HostError('aapt not found. Please make sure at least one Android platform is installed.')
+    elif not subprocess.call('aapt version >{} 2>&1'.format(get_null()), shell=True):
+        # User may already installed 'aapt' from package provided by distro's
+        # Try finding in $PATH
+        aapt_path = os.path.dirname(subprocess.check_output('which aapt', shell=True))
+        aapt_path = os.path.join(aapt_path, 'aapt')
+        logger.debug('Using aapt from {}'.format(aapt_path))
```
setrofim: Again, use WA's `which()` here (see comment above).

dev_scripts/get_apk_versions
koukaipan: Oh!!! Thank you for pointing it out, I did not notice that there is a useful wrapper!!

PR #552, opened 2017-12-01
Add Create Workload Command

marcbonnici:


wa/commands/templates/uiauto/UiAutomation.java
```diff
@@ -0,0 +1,57 @@
...
+import com.arm.wa.uiauto.BaseUiAutomation;
+
+@RunWith(AndroidJUnit4.class)
+public class UiAutomation extends BaseUiAutomation {
+
+    protected Bundle parameters;
+
+    public static String TAG = "${name}";
+
+@Before
```
bjackman: Is it normal to have these decorators (sorry if wrong jargon) de-indented like this?

marcbonnici: No they shouldn't have been, thanks for pointing that out.

setrofim:
Typo in commit title: "framwork/entrypoint:" --> "framework/entrypoint:"

wa/commands/create.py
```diff
@@ -68,6 +78,75 @@ def execute(self, state, args):
...
+        elif args.uiautomation:
+            kind = 'uiauto'
+        elif args.apk_uiautomation:
+            kind = 'apk_uiauto'
+        else:
+            kind = 'basic'
+
+        try:
+            create_workload(args.name, kind, where, check_name)
+        except CommandError, e:
```
setrofim: Should be `except CommandError as e`

wa/commands/create.py
```diff
@@ -68,6 +78,75 @@ def execute(self, state, args):
...
+            print "ERROR:", e
```
setrofim: Should log this rather than print.

wa/commands/create.py
```diff
@@ -68,6 +78,75 @@ def execute(self, state, args):
...
+                                    help='Use UI automation template. This template generates a UI automation ' +
+                                         'Android project as well as the Python class. This template that makes ' +
+                                         'no assumptions about the nature of your workload, apart from the fact ' +
+                                         'that you need UI automation.')
+        template_group.add_argument('-U', '--apk-uiautomation', action='store_true',
+                                    help='Use Apk UI automation template. This template generates a UI automation ' +
+                                         'Android project as well as the Python class. It also allows you to specify' +
+                                         'an APK file that will be installed and ran on the device. You should place ' +
+                                         ' the APK file into the workload\'s directory at the same level as the ' +
+                                         ' __init__.py.')
```
setrofim: Given an increasing number of different workload configurations (and it is entirely possible that we will be adding more in the future), perhaps it might make sense to have a single '--kind'  option that takes one of a set of values as an argument, rather than a different option for each kind.

wa/commands/create.py
```diff
@@ -77,7 +156,121 @@ class CreateCommand(ComplexCommand):
...
+        workload_dir = _d(os.path.join(where, name))
+
+    if kind == 'basic':
+        create_basic_workload(workload_dir, name, class_name, **kwargs)
+    elif kind == 'apk':
+        create_apk_workload(workload_dir, name, class_name, **kwargs)
+    elif kind == 'revent':
+        create_revent_workload(workload_dir, name, class_name, **kwargs)
+    elif kind == 'apk_revent':
+        create_apk_revent_workload()(workload_dir, name, class_name, **kwargs)
```
setrofim: A stray set of parens, `()`, here?

wa/commands/create.py
```diff
@@ -77,7 +156,121 @@ class CreateCommand(ComplexCommand):
...
+    elif kind == 'uiauto':
+        create_uiautomator_workload(workload_dir, name, class_name, **kwargs)
+    elif kind == 'apk_uiauto':
+        create_apk_uiautomator_workload(workload_dir, name, class_name, **kwargs)
+    else:
```
setrofim: Could simplify this a little by having a `dict` mapping the kind name onto the corresponding function and then just doing

```
try:
    create_funcs[kind](worload_dir, name, class_name, **kwargs)
except KeyError:
     raise CommandError('Unknown workload type: {}'.format(kind))
```

marcbonnici: Addressed comments.

wa/commands/create.py
```diff
@@ -68,6 +78,44 @@ def execute(self, state, args):
...
+        self.parser.add_argument('name', metavar='NAME',
+                                 help='Name of the workload to be created')
+        self.parser.add_argument('-p', '--path', metavar='PATH', default=None,
+                                 help='The location at which the workload will be created. If not specified, ' +
+                                      'this defaults to "~/.workload_automation/plugins".')
+        self.parser.add_argument('-f', '--force', action='store_true',
+                                 help='Create the new workload even if a workload with the specified ' +
+                                      'name already exists.')
+        self.parser.add_argument('-k', '--kind', metavar='KIND', default='basic',
+                                 help='The type of workload to be created. The available options ' +
```
setrofim: Should set `allowed_values` to `workload_types`

wa/commands/create.py
```diff
@@ -77,7 +125,91 @@ class CreateCommand(ComplexCommand):
...
-        #CreatePackageSubcommand,
+        CreateWorkloadSubcommand,
         CreateAgendaSubcommand,
+        #CreatePackageSubcommand,
     ]
+
+
+def create_workload(name, kind='basic', where='local', check_name=True, **kwargs):
+
+    create_funcs = {
```
setrofim: Rather than listing the workloads in `workload_types` class attribute, and then again here, I would move this dict into global level and the use its keys instead of `workload_types`

marcbonnici: Thanks, updated.

wa/commands/create.py
```diff
@@ -77,7 +116,90 @@ class CreateCommand(ComplexCommand):
...
+    os.chmod(build_script, stat.S_IRWXU | stat.S_IRWXG | stat.S_IRWXO)
+
+    source_file = _f(os.path.join(path, 'app', 'src', 'main', 'java',
+                                  os.sep.join(package_name.split('.')[:-1]),
+                                  'UiAutomation.java'))
+    with open(source_file, 'w') as wfh:
+        wfh.write(render_template(os.path.join('uiauto', 'UiAutomation.java'),
+                                  {'name': name, 'package_name': package_name}))
+
+create_funcs = {
```
setrofim: Move to the top (below imports) and update indentation should be 
```

create_funcs = {
     'basic': create_template_workload,
      ....
}
```

marcbonnici: Updated

PR #554, opened 2017-12-04
Fix some pylint errors

bjackman:
Nothing very interesting to see here but might as well not throw it away.

### DIFF
--- a/wa/framework/execution.py
+++ b/wa/framework/execution.py
@@ -417,7 +417,7 @@ class Runner(object):
         except KeyboardInterrupt:
             job.set_status(Status.ABORTED)
             raise
-        except Exception as e:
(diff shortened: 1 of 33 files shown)


wa/framework/execution.py
```diff
@@ -277,7 +266,6 @@ def __init__(self):
         self.logger = logging.getLogger('executor')
         self.error_logged = False
         self.warning_logged = False
-        pluginloader = None
```
bjackman: Hmm this looks suspect actually

bjackman:
New commits include one actual bug fix (framework/output: Fix checking artifact kind) and a tests fix (tests: Update use of devlib cpufreq API)

PR #564, opened 2017-12-11
Juno fixes.

setrofim:
NOTE: this pull request depends on https://github.com/ARM-software/devlib/pull/220

### DIFF
--- a/wa/instrumentation/energy_measurement.py
+++ b/wa/instrumentation/energy_measurement.py
@@ -25,6 +25,7 @@ from devlib.instrument.energy_probe import EnergyProbeInstrument
 from devlib.instrument.daq import DaqInstrument
 from devlib.instrument.acmecape import AcmeCapeInstrument
 from devlib.instrument.monsoon import MonsoonInstrument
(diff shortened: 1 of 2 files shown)


wa/instrumentation/energy_measurement.py
```diff
@@ -284,9 +292,10 @@ def initialize(self, context):
         self.instruments = self.backend.get_instruments(self.target, **self.params)
 
         for instrument in self.instruments.itervalues():
-            if instrument.mode != CONTINUOUS:
+            if not (instrument.mode & CONTINUOUS):
                 msg = '{} instrument does not support continuous measurement collection'
                 raise ConfigError(msg.format(self.instrument))
+            instrument.setup()
```
marcbonnici: Why does setup get called here?

setrofim: Because we should make sure this happens at least once per run. This is where, e.g., binaries get installed.  See

http://devlib.readthedocs.io/en/latest/overview.html#one-time-setup

PR #565, opened 2017-12-12
configuration/core.py: fix run of result processors

ionela-voinescu:
Fix typo %s/result_processor/result_processors that results in
result processors to not be run.

Signed-off-by: Ionela Voinescu <ionela.voinescu@arm.com>


setrofim:
This typo needs to be fixed in the opposite direction -- the kind needs to be `"result_processor"` (singular).

setrofim:
... Which it already seems to be. Are you running into an error due to this? What are the steps to reproduce?

ionela-voinescu:
Okay, let me take another look. Yes, in my latest tests none of the result processors (default:csv, status) are run.

ionela-voinescu:
Closing this as it's obviously not the right fix.

PR #568, opened 2017-12-13
Show Command

marcbonnici:
Add show command to WA3

### DIFF
--- a/wa/commands/show.py
+++ b/wa/commands/show.py
@@ -1,0 +1,94 @@
+#    Copyright 2014-2015 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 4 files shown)


wa/commands/show.py
```diff
@@ -0,0 +1,93 @@
...
+        self.parser.add_argument('plugin', metavar='PLUGIN',
+                                 help='The name of the plugin to display documentation for.')
+
+    def execute(self, state, args):
+        name = args.plugin
+        rst_output = None
+
+        plugin = get_plugin(name)
+        if plugin:
+            rst_output = get_rst_from_extension(plugin)
```
setrofim: Should rename this function to `get_rst_from_plugin`

wa/commands/show.py
```diff
@@ -0,0 +1,93 @@
...
+            # Correctly format the tile of the man page
+            output = output.replace('.SH', '.TH', 1)
+            call('echo "{}" | man -l -'.format(escape_double_quotes(output)), shell=True)
+        else:
+            print rst_output
+
+def get_plugin(name):
+    for kind in pluginloader.kinds:
+        results = pluginloader.list_plugins(kind)
+        for result in results:
```
setrofim: Why iterate by `kind` here? Why not just `for result in pluginloader.list_plugins():` ?

wa/commands/show.py
```diff
@@ -0,0 +1,93 @@
...
+            if name == result.name:
+                return result
+            if hasattr(result, 'alias'):
+                for alias in result.alias:
+                    if name == alias:
+                        return result
+
+def get_target(name):
```
setrofim: For the sake of clarity, I would rename this to something like `get_target_description`

marcbonnici:
Comments Addressed

PR #581, opened 2018-01-12
Exec control fixes

marcbonnici:


wa/utils/exec_control.py
```diff
@@ -96,10 +94,7 @@ def wrapper(*args, **kwargs):
         if __active_environment is None:
             activate_environment('default')
 
-        func_id = repr(method.func_name)
-        # Store the least derived class, which isn't object, to account
-        # for subclasses.
-        func_id += repr(getmro(args[0].__class__)[-2])
+        func_id = repr(method.__code__)
```
setrofim: This means  `func_id` can be arbitrarily large, depending on the function size. While large functions should be avoided anyway, I'd still be more comfortable is this had a bounded maximum size. One way to ensure  that would be to compute a hash. A computationally lighter solution might be be just to take the first N (e.g.) bytes  of the `__code__` object.

setrofim: OK, never mind, I was being silly. `__repr__` only returns a string with the function address and the file. That is both unique and short enough.

PR #583, opened 2018-01-12
Exoplayer: Add support for Landscape mode

derkling:
To improve reproducibility of tests, it could help to ensure that the
same screen orientation is used on media playback.
This adds a new workload parameter which allows to ensore a LANDSCAPE
mode, which can be useful to better represent video playback scenarios.
When not enforced by the agenda, we enforce a PORTRAIT orientation.

Signed-off-by: Patrick Bellasi <patrick.bellasi@arm.com>


derkling:
Fixed typo in commit message. Thanks.

PR #584, opened 2018-01-15
[RFC] Add Initial Support for ChromeOs

marcbonnici:
Dependant on PR-https://github.com/ARM-software/devlib/pull/228

### DIFF
--- a/wa/framework/target/descriptor.py
+++ b/wa/framework/target/descriptor.py
@@ -2,7 +2,7 @@ from collections import OrderedDict
 from copy import copy
 
 from devlib import (LinuxTarget, AndroidTarget, LocalLinuxTarget,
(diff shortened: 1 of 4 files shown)


wa/framework/target/descriptor.py
```diff
@@ -4,14 +4,14 @@
 from devlib import (LinuxTarget, AndroidTarget, LocalLinuxTarget,
                     Platform, Juno, TC2, Gem5SimulationPlatform,
                     AdbConnection, SshConnection, LocalConnection,
-                    Gem5Connection)
+                    Gem5Connection, ChromeOsTarget)
```
setrofim: Really knit-picky, but I would stick the new Target import just after the existing Target imports.

marcbonnici:
Updated to support the conditional presence of the android container.

PR #589, opened 2018-01-25
Timestamps alignment between poller and ftrace

setrofim:
Add support for aligning timestamps in poller csv output with those that appear
in ftrace.


wa/instruments/poller/poller.c
```diff
@@ -40,10 +57,12 @@ int main(int argc, char ** argv) {
...
     memset(buf, 0, sizeof(buf));
-    struct timeval current_time;
+    struct timespec current_time;
     double time_float;
     char *labels;
     int labelCount = 0;
+    int should_write_marker = 0;
+    int ret;
 
     static char usage[] = "usage: %s [-h] [-t INTERVAL] FILE [FILE ...]\n"
```
marcbonnici: Update usage message with new parameter?

setrofim: done.

PR #594, opened 2018-01-31
Disable augmentations fix

marcbonnici:


wa/framework/plugin.py
```diff
@@ -528,6 +528,8 @@ def resolve_alias(self, alias_name):
         is already an plugin name, then the result is ``(alias_name, {})``.
 
         """
+        if alias_name.startswith('~'):
```
setrofim: This should not be done here.  The `'~'`  is part of configuration notation, `PluginLoader` should not be aware of this...

marcbonnici: Updated to deal with this in `configuration/core`.

PR #601, opened 2018-02-06
Enabling support in the geekbench workload for the latest versions of…

scojac01:
… the standard and corporate applications


wa/workloads/geekbench/__init__.py
```diff
@@ -407,11 +388,16 @@ class GeekbenchCorproate(Geekbench):
...
+    versions = {
+        '5.0.0': {
+            'package': 'com.primatelabs.geekbench4.corporate',
+            'activity': 'com.primatelabs.geekbench.HomeActivity',
+        },
+        '4.1.0': {
+            'package': 'com.primatelabs.geekbench4.corporate',
+            'activity': 'com.primatelabs.geekbench.HomeActivity',
+        },
+        }
```
marcbonnici: If the package name and activity do not differ then the extra version could just be appended to the versions list, however this is going to be changed to dictionary then the `version` parameter will also need to be updated so that the `allowed_values` and `default` parameters are populated correctly.

wa/workloads/geekbench/__init__.py
```diff
@@ -118,6 +113,7 @@ def __init__(self, *args, **kwargs):
         self.gui.uiauto_params['version'] = self.version
         self.gui.uiauto_params['times'] = self.times
         self.gui.uiauto_params['is_corporate'] = self.is_corporate
+        self.gui.timeout = 3600
```
marcbonnici: The gui timeout should probably be configurable and related to the `timeout` parameter.

wa/workloads/geekbench/uiauto/app/src/main/java/com/arm/wa/uiauto/geekbench/UiAutomation.java
```diff
@@ -95,6 +96,10 @@ public void runWorkload() throws Exception {
                     runCpuBenchmarks(isCorporate);
                     waitForResultsv3onwards();
                     break;
+                case 5:
+                    runCpuBenchmarks(isCorporate);
+                    waitForResultsv3onwards();
+                    break;
```
setrofim: This is identical to case 4. Just fall through and use the same block for both?

PR #619, opened 2018-02-07
Updating Geekbench workload to support the latest versions of both th…

scojac01:
…e standard and corporate application


wa/workloads/geekbench/uiauto/app/src/main/java/com/arm/wa/uiauto/geekbench/UiAutomation.java
```diff
@@ -91,7 +92,7 @@ public void runWorkload() throws Exception {
                         shareResults();
                     }
                     break;
-                case 4:
+                case 4: case 5:
```
setrofim: Should be on the next line

PR #656, opened 2018-02-20
Googleslides: Updating the workload to work with version 1.18 and sup…

scojac01:
…port the latest Huawei devices


wa/workloads/googleslides/__init__.py
```diff
@@ -121,3 +121,19 @@ def init_resources(self, context):
...
+
+    def deploy_assets(self, context):
+        super(GoogleSlides, self).deploy_assets(context)
+        #Move the test files to the download folder
+        d = self.target.working_directory
+        e = self.target.external_storage
+
+        orig_file_path = self.target.path.join(d, self.test_file)
+        new_file_path = self.target.path.join(e, 'Download')
+        self.target.execute('cp {} {}'.format(orig_file_path, new_file_path))
```
marcbonnici: Might be worth adding a `self.target.refresh_files(self.deployed_assets)` here?

wa/workloads/googleslides/__init__.py
```diff
@@ -121,3 +121,19 @@ def init_resources(self, context):
...
             raise ValidationError('{} must be a PPTX file'.format(self.test_file))
+
+    def deploy_assets(self, context):
+        super(GoogleSlides, self).deploy_assets(context)
+        #Move the test files to the download folder
+        d = self.target.working_directory
+        e = self.target.external_storage
+
+        orig_file_path = self.target.path.join(d, self.test_file)
+        new_file_path = self.target.path.join(e, 'Download')
```
marcbonnici: Need to add the `new_file_path` to `self.deployed_assets` for the clean up to work correctly.

scojac01:
Updated.

wa/workloads/googleslides/__init__.py
```diff
@@ -121,3 +121,21 @@ def init_resources(self, context):
...
```
marcbonnici: Just checking that you meant the 'Download' folder as opposed to the 'Downloads' folder? 
And also will this always be present on the device or should there be a check / creation of the folder?

wa/workloads/googleslides/__init__.py
```diff
@@ -121,3 +121,21 @@ def init_resources(self, context):
...
+        self.deployed_assets.append(new_file_path)
+        self.target.execute('cp {} {}'.format(orig_file_path, new_file_path))
+        self.target.refresh_files(self.deployed_assets)
+
+    def remove_assets(self, context):
+        for asset in self.deployed_assets:
+            self.target.remove(os.path.dirname(asset))
+        self.target.refresh_files(self.deployed_assets)
```
marcbonnici: Is your intention to to delete the entire 'Download' folder here? 
If not then this method can be removed and the built-in functionality will remove the individual files.

wa/workloads/googleslides/__init__.py
scojac01: Downloads is correct. As far as I can tell, this is the default folder on all devices.

wa/workloads/googleslides/__init__.py
scojac01: That is not my intention. I will remove.

wa/workloads/googleslides/__init__.py
scojac01: Isn't this happening on line 133? 

self.deployed_assets.append(new_file_path)

PR #677, opened 2018-03-13
Add arm energy probe instrument

vingu-linaro:
Arm Energy Probe with arm_probe has been recently added in devlib's instrument.
Add the arm_energy_probe in the WA list of Energy measurement instruments

Signed-off-by: Vincent Guittot <vincent.guittot@linaro.org>

### DIFF
--- a/wa/instruments/energy_measurement.py
+++ b/wa/instruments/energy_measurement.py
@@ -18,10 +18,12 @@
 from __future__ import division
 from collections import defaultdict
 import os
+import shutil
 
 from devlib import DerivedEnergyMeasurements
 from devlib.instrument import CONTINUOUS
 from devlib.instrument.energy_probe import EnergyProbeInstrument
+from devlib.instrument.arm_energy_probe import ArmEnergyProbeInstrument
 from devlib.instrument.daq import DaqInstrument
 from devlib.instrument.acmecape import AcmeCapeInstrument
 from devlib.instrument.monsoon import MonsoonInstrument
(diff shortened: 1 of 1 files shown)


wa/instruments/energy_measurement.py
```diff
@@ -153,6 +154,18 @@ def validate_parameters(self, params):
                 msg = 'Number of Energy Probe port labels does not match the number of resistor values.'
                 raise ConfigError(msg)
 
+class ArmEnergyProbeBackend(EnergyInstrumentBackend):
+
+    name = 'arm_energy_probe'
+
+    parameters = [
+        Parameter('config_file', kind=str, default='./config',
```
setrofim: Avoid relative paths in parameter values. 

The config files should by preserved as part of the output (probably in `__meta`) as it is essential for reproducing results.

vingu-linaro: So I'm going to remove default path for config
ok. I'm going to copy the file in __meta

vingu-linaro: How can I get path to __meta from EnergyInstrumentBackend ?
I have planned to copy the file from validate_parameters() but can't get the path where to copy the file

setrofim: The path can be obtained via `context.run_output.metadir` I would add that as an argument to `EnergyInstrumentBackend`'s  `__init__`, and it can then be passed inside `EnergyMeasurement.initialize`  via the `get_plugin` call. 

The actual copying can than happen inside your backend's `get_instruments` since this is also where it is used to initialize the `devlib` instrument.

vingu-linaro:
I have updated the patch and tried to implement what you proposed

setrofim:
This is not what I suggesting. Do not pass the `context` around everywhere -- just pass in the one path that you need. Also I suggested adding the parameter to the`__init__` not `get_instruments`, though this is probably not that important, and I'm fine with you leaving it there.

vingu-linaro:
I have updated the patchset to pass only meta directory path. Sorry if I misunderstood you request in the previous version

setrofim:
Please update the title of your first commit

"Add context parameter" --> "Add meta directory"

PR #684, opened 2018-03-13
workloads/antutu: Remove assertion that all scores are found

marcbonnici:


scojac01:
I've just noticed that the AndroidManifest.xml in the antutu workload is in uiauto/app/src when it should be in uiauto/app/src/main - could we roll this up into your change?

PR #685, opened 2018-03-14
docker: Add note on how to use TTY

Sticklyman1936:
If one wishes to use the docker container with a TTY, then one must
explicitly mount it into the container when starting it. A note
conveying this has been added to the Dockerfile.

### DIFF
--- a/extras/Dockerfile
+++ b/extras/Dockerfile
@@ -24,6 +24,12 @@
 # such as required assets, can be mounted into the container using a
 # second --volume command.
 #
+# If you require access to a TTY from the Docker container, please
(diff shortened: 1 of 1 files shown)


extras/Dockerfile
```diff
@@ -24,6 +24,10 @@
 # such as required assets, can be mounted into the container using a
 # second --volume command.
 #
+# If you require access to a TTY from the Docker container, please
+# also mount this into the container in the same style as is used to
+# mount USB.
```
setrofim: I think an example would be useful here.

PR #688, opened 2018-03-14
commands/show: Adds support for showing global settings

marcbonnici:


wa/commands/show.py
```diff
@@ -98,3 +105,9 @@ def get_rst_from_target(target):
     text += get_params_rst(target.assistant_params)
     text += '*Note:* For available runtime parameters please see the documentation'
     return text + '\n'
+
+def get_rst_for_global_config():
+    text = underline('Global Configuration')
```
setrofim: I think it might be useful to add a short paragraph here explaining what this. I.e. That this is controls the behaviour of WA/run as a whole and that it should go into a config file (either in `$WA_USER_DIRECTORY/config.yaml` or one specified with `-c`), or into  `config`/`global` section of the agenda.

wa/commands/show.py
```diff
@@ -37,17 +39,22 @@ def execute(self, state, args):
         name = args.plugin
         rst_output = None
 
-        plugin = get_plugin(name)
-        if plugin:
-            rst_output = get_rst_from_plugin(plugin)
-            plugin_name = plugin.name
-            kind = '{}:'.format(plugin.kind)
+        if name == caseless_string('settings'):
```
setrofim: In addition to documenting global settings here, I would also add a separate section (still under `settings` keyword) for environment variables.

marcbonnici:
Comments Addressed.

PR #691, opened 2018-03-15
Agenda conversion

marcbonnici:


wa/framework/host.py
```diff
@@ -36,3 +44,75 @@ def init_user_directory(overwrite_existing=False):  # pylint: disable=R0914
...
+            write_param_yaml(entry, param, output)
+
+        # Write any additional aliased paramters into new config
+        plugin_cache = PluginCache()
+        output.write("# Additional global aliases\n")
+        for param in orig_agenda.keys():
+            if plugin_cache.is_global_alias(param):
+                entry = {param: orig_agenda[param]}
+                yaml.dump(format_parameter(entry), output, default_flow_style=False)
+                output.write("\n")
```
setrofim: What about converting plugin-specific configuration? Also need a check here to see if `param` is a plugin name.

marcbonnici: Updated.

PR #694, opened 2018-03-17
Add "process" command

Rahlubenru:


wa/commands/process.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self, subparsers):
+        self.loader = pluginloader
+        super(ProcessCommand, self).__init__(subparsers)
+
+    def initialize(self, context):
+        self.parser.add_argument('-d', '--process-directory', metavar='DIR', default=None,
+                                 help="""
+                                 Specify a directory to walk to find the data
+                                 from previous runs to process .If this option is
+                                 not specified, then {} will be used instead.
```
setrofim: Add double grave quotes (` `` `) around  the path to make sure it displays as monospace.

wa/commands/process.py
```diff
@@ -0,0 +1,84 @@
...
+                                 """.format(settings.default_output_directory))
+
+    def execute(self, config, args):
+        self.logger.debug('Version: {}'.format(get_wa_version()))
+        self.logger.debug('Command Line: {}'.format(' '.join(sys.argv)))
```
setrofim: Remove these. These are present in the `run` command to ensure this information becomes part of `run.log` in the output. There is no purpose for these here.

wa/commands/process.py
```diff
@@ -0,0 +1,84 @@
...
+
+    name = 'process'
+    description = 'Process the output from previously run workloads.'
+
+    def __init__(self, subparsers):
+        self.loader = pluginloader
+        super(ProcessCommand, self).__init__(subparsers)
+
+    def initialize(self, context):
+        self.parser.add_argument('-d', '--process-directory', metavar='DIR', default=None,
```
setrofim: It makes no sense to run process command without specifying an existing output directory. Make this an argument, not an option

wa/commands/process.py
```diff
@@ -0,0 +1,84 @@
...
+                                 """.format(settings.default_output_directory))
+
+    def execute(self, config, args):
+        self.logger.debug('Version: {}'.format(get_wa_version()))
+        self.logger.debug('Command Line: {}'.format(' '.join(sys.argv)))
+
+        if args.process_directory:
+            process_directory = args.process_directory
+        else:
+            process_directory = settings.default_output_directory
```
setrofim: Do not do this -- force the user to specify the directory they want to post-process explicitly.

wa/commands/process.py
```diff
@@ -0,0 +1,84 @@
...
+        process_directory = os.path.expandvars(process_directory)
+        self.logger.debug('Using process directory: {}'.format(process_directory))
+        if not os.path.exists(process_directory):
+            self.logger.error('Path "{}" does not exist'.format(process_directory))
+            return
```
setrofim: Rather than logging and returning, raise a `CommandError` here instead.

wa/commands/process.py
```diff
@@ -0,0 +1,84 @@
...
+        for run_output in discover_wa_outputs(process_directory):
```
setrofim: Don't do this, at least not by default. The directory should be a WA output directory not contain it. Use `RunOutput` directly here instead.

If you want to add the ability to recursively process a directory tree with multiple outputs, make the user explicitly specify that that is their intent, e.g. with `-r` opition.

wa/commands/process.py
```diff
@@ -0,0 +1,84 @@
...
+            for proc in config.get_processors():
+                self.run_processor(run_output, proc)
```
setrofim: Instead of implementing your own `run_processor`, use a `ProcessorManager` here:

https://github.com/ARM-software/workload-automation/blob/next/wa/framework/output_processor.py#L35

This will ensure that the various stages for different processors get executed in the correct order.

Rahlubenru: It appears to be the `Executor` and `Runner` in
https://github.com/ARM-software/workload-automation/blob/next/wa/framework/execution.py
that are governing the sequence rather than the `ProcessorManager`. 

Should I be trying to get this to work with the <Executor> or stick with just using the `ProcessorManager`?

The `ProcesorManager` is still useful, except that `process_job_output` requires a context to get the `job_output`, `target_info`, and `run_output` from. When called from the `Runner` this is an `ExecutionContext`.

I can see two reasonable ways around this if I'm to continue using the `ProcessorManager`, create a new class for the process command to use that contains these values, or don't use the `ProcessorManager` for that call. Do you have a preference for either of these or object to both?

setrofim: Hm, good point. It maybe worth creating a version of the `ProcessorManager` that directly takes `job_output`, `target_info`, and `run_output`; or maybe that's overkill. In any case, the important thing is to ensure that all stages are executed in the correct order, you should execute each staged for all enabled processors before moving on to the next (rather than executing all stages for each processor in turn).

Rahlubenru:
If this should skip previously used processors by default would I be correct in assuming that it will need to update the records with any additional processors that were used during the run?

setrofim:
> I did try to reduce it to the one commit but I'm pretty sure I failed at that so apologies there. 

Please try this:

http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Rahlubenru:
That is decidedly similar to what I did do, though since they were all the same file I dropped the older two. On my fork I do now only have that one commit but pushing here it doesn't appear to have removed the previous two, is that expected?

Sticklyman1936:
Hi! You don't want to be "dropping" the older commits, but instead you want to be squashing your commits into a single commit. By dropping commits you are losing the information in those commits. Looking at your fork, it appears that all of your commits are still there.

My advice would be do to the following:

* `git checkout` the changes you have on your fork - these are likely on `origin/next`, but depending on your naming scheme this can be different.
* Next, again going by your fork, you will want to do a `git rebase HEAD~3`. The simplest solution here would be to leave the first (as represented in when rebasing) commit as is, squash the second commit and the third commit into the first. For each squash you have the ability to update the commit message, so be sure to modify the commit message appropriately.
* Check your work with `git log` and `git show` to be sure you still have all your changes
* Finally, push your changes to your fork with `git push --force origin next` (again, assuming your fork is called 'origin'). Note that the force is required as you are re-writing history, and otherwise the push will fail.

marcbonnici:
I'm not entirely sure what has happened here so I think the easiest way to fix this might be to just backup your `process.py` file outside of your git repository, checkout the latest version of the next branch from WA and create your commit again.

So first thing to check, do you have the main WA repository set up as a remote? And what is the output if you run:  `git remote -v`?

Rahlubenru:
I already took a manual backup of the process.py file as soon as things started going wrong.

I do believe I have the main WA set up as remote the output of the requested command is below.

> matthew@linux-drzu:~/PycharmProjects/workload-automation> git remote -v
> origin  https://github.com/Rahlubenru/workload-automation.git (fetch)
> origin  https://github.com/Rahlubenru/workload-automation.git (push)
> upstream        https://github.com/ARM-software/workload-automation.git (fetch)
> upstream        https://github.com/ARM-software/workload-automation.git (push)

marcbonnici:
Ah ok that's good.
So in that case you should be able to do `git reset --hard upstream/next` which will remove any changes you've done locally and take your branch back to the upstream version.

After this you should be able to restore your file, create a new commit with your changes and then force push that back to your `origin/next`.

Rahlubenru:
The reset has reset the commits, `git log` shows the commit history for the upstream now. It has not removed the file from my fork or my local folder though and so I can't recommit the file.

I tried using `git checkout upstream/next` and `git fetch upstream` but neither of these has aligned my files with the upstream.

marcbonnici:
Was the reset command with the `--hard` argument while you were on your `origin/next` branch?

`git fetch upstream` should have retrieved the latest version of upstream from github and your `git checkout upstream/next` would have taken you into a detached state so you will need to run a `git checkout next` to get back to your branch before continuing.

And just to check, when you're back on your `next` branch, what is the last commit that `git log` shows you?

Rahlubenru:
I think I have my fork sorted now, I just ran `git push origin next --force` even though my local copy looks wrong and that worked. My local is still wrong though somehow. 

After the `checkout next` it was my most recent commit but I did the reset again after and so the following is the last entry.

> commit 72f2f8259480717c31d036047caf4380d1916121 (HEAD -> next, upstream/next, origin/next, 
> origin/HEAD)
> Author: Sergei Trofimov <sergei.trofimov@arm.com>
> Date:   Thu Mar 22 12:19:47 2018 +0000
> 
>     fw/config: better error when merging augs
>     
>     If one or more entries for augmentations in configuration contains an
>     invalid value, raise ConfigError with the entry name.

marcbonnici:
Ok, so you appear to have a clean branch that is up to date with the latest upstream version now. 

In which case you should now be able to restore your `process.py` file and re-commit it?

Rahlubenru:
Yes, I found the problem with my local copy so I'll get this redone shortly. Thank you for your patience.

it appears to have gone to #701 so I guess this one needs closing or is there a way I can get it to pull to this one?

Rahlubenru:
It appears I closed it myself, not quite sure how.

marcbonnici:
Ok cool, no problem. 

It may have closed itself when all the commits were removed and I'm not aware of a way to combine the PR's again so continuing on #701 is fine.

PR #701, opened 2018-03-23
@Rahlubenru

Rahlubenru:
commands/process create initial version

Add process command to run output processors over data produced by
a previous run.

Default processing is to skip the processors that were used during
the run and to process the specified directory only. Options to
use additional processors, force reprocessing and allow a directory
walk to discover additional output directories.

### DIFF
--- a/wa/commands/process.py
+++ b/wa/commands/process.py
@@ -1,0 +1,116 @@
+#    Copyright 2014-2018 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+
+from wa import Command
+from wa import discover_wa_outputs
+from wa.framework.exception import CommandError
+from wa.framework.output import RunOutput
+from wa.framework.output_processor import ProcessorManager
+
+class ProcessContext(object):
+
+    def __init__(self):
+        self.run_output = None
+        self.target_info = None
+        self.job_output = None
+
+class ProcessCommand(Command):
+
+    name = 'process'
+    description = 'Process the output from previously run workloads.'
+
+    def initialize(self, context):
+        self.parser.add_argument('directory', metavar='DIR',
+                                 help="""
+                                 Specify a directory containing the data
+                                 from a previous run to be processed.
+                                 """)
+        self.parser.add_argument('-p', '--processor', action='append',
+                                 dest='additional_processors', metavar='OutputProcessor',
+                                 help="""
+                                 Specify an output processor to add from the
+                                 command line. This can be used to run a
+                                 processor that is not normally used without
+                                 introducing permanent change to the config 
+                                 (which one might then forget to revert). This
+                                 option may be specified multiple times.
+                                 """)
+        self.parser.add_argument('-f', '--force', action='store_true',
+                                 help="""
+                                 Run processors that have already been
+                                 run. By default these will be skipped.
+                                 """)
+        self.parser.add_argument('-r', '--recursive', action='store_true',
+                                 help="""
+                                 Walk the specified directory to process
+                                 all of the previous runs contained within
+                                 instead of just processing the root.
+                                 """)
+
+    def execute(self, config, args):
+        process_directory = os.path.expandvars(args.directory)
+        self.logger.debug('Using process directory: {}'.format(process_directory))
+        if not os.path.exists(process_directory):
+            msg = 'Path `{}` does not exist, please specify a valid path.'
+            raise CommandError(msg.format(process_directory))
+        if not args.recursive:
+            output_list = [RunOutput(process_directory)]
+        else:
+            output_list = [output for output in discover_wa_outputs(process_directory)]
+        pc = ProcessContext()
+        for run_output in output_list:
+            if not args.recursive:
+                self.logger.info('Installing output processors')
+            else:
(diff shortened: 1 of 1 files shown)


Rahlubenru:
Apologies for being late to get back, I was busy the last few days. I made changes to add the missing export stage and the run processing. In testing though I noticed that the sqlite processor requires additional parameters to function, specifically the database to use. That makes the specifying of additional processors in the command more complicated, should I have it accept an agenda file to read processors from instead?

setrofim:
> should I have it accept an agenda file to read processors from instead?

No, this should not take an agenda. It should already be possible to provide configuration for the processors via the config files. When you create `ProcessorManager`, specify `loader=config.plugin_cache`; that will ensure that configuration from config takes effect for the processors you install from command line.

wa/commands/process.py
```diff
@@ -0,0 +1,107 @@
...
+        self.target_info = None
+        self.job_output = None
+
+class ProcessCommand(Command):
+
+    name = 'process'
+    description = 'Process the output from previously run workloads.'
+
+    def __init__(self, subparsers):
+        self.loader = pluginloader
```
setrofim: This is not used and is not necessary; use `config.plugin_cache` instead.

Rahlubenru: I tried that, while correct that the quoted point is not required, using `config.plugin_cache` instead of `pluginloader` when initialising the `ProcessorManager` results in it not being able to perform the `install` operation, happy to try to work around this if you think that is still the correct course of action.

While I did need that config point clarifying regardless, further investigation of the problem I was having revealed it was not the cause. The cause is that in the `Configuration` class in `wa.framework.configuration.core` has a `from_pod` function that loads the line: 

> "boot_parameters": null, 

from jobs.json as an empty dict `{}`, which results in `_update_spec` in the sqlite processor to fail the `json.dumps(spec.boot_parameters)` call because it is not serializable. Is there a reason the pod load works this way? During the run the `boot_parameters` value is `None`

In the meantime I will continue to test what I have plus check whether the config is working, the database is created before the failure point above for the sqlite processor.

Rahlubenru:
Is there a way to discover what processors have already been used on the run rather than the job? Because if there is I haven't been able to find it to know what to disable when processing the run.

setrofim:
> Is there a way to discover what processors have already been used on the run rather than the job? 
Output processors enabled for job can be accessed via the job's output; their names will be listed in `job_output.spec.augmentations` (instruments and output processors are handled in the same way in configuration as "augmentations" -- sorry I should have mentioned this earlier, I realise the mapping is not obvious).

wa/commands/process.py
setrofim: erm, I'll need to look into this; an empty dict should be serializable.

setrofim: ok, I see the problem; it's actually an `obj_dict`, so needs to be `to_pod()`'d. Will fix.

setrofim: Should be fixed by https://github.com/ARM-software/workload-automation/pull/709/files

Rahlubenru:
I was already using `job_output.spec.augmentations` for the jobs, is there an equivalent for the run as a whole is what I was asking? Since I was asked to add `process_run_output` and not rerun previously completed processors.

wa/commands/process.py
Rahlubenru: Thank you, I'll grab that and retest

Rahlubenru: That does seem to have fixed it thanks

setrofim:
Ah, sorry, I misunderstood. Augmentations are tracked at job level, so augmentations for the run would be the union of augmentations for all jobs. There is no property for it in `RunOutput` at the moment, though I've added one here: 

https://github.com/ARM-software/workload-automation/pull/706

(This probably won't get merged until next week though, as there is a bank holiday in UK tomorrow.)

Rahlubenru:
I tried squashing but then it says I'm pushing the squashed commit and 10 other commits that are already in the branch: the ones I pulled to my fork to gain the changes I required. Should I do it that way or just tidy up the descriptions for the three I have? Happy to redo the squash just to show you what I mean if you want to see before deciding.

setrofim:
Erm, you seem to have merge commits in there? You should be rebasing your commits on top of latest `next`, not merging. And yes, please squash commits making changes to earlier commits in the same pull request.

wa/commands/process.py
```diff
@@ -0,0 +1,116 @@
...
+        else:
+            output_list = [output for output in discover_wa_outputs(process_directory)]
+        pc = ProcessContext()
+        for run_output in output_list:
+            if not args.recursive:
+                self.logger.info('Installing output processors')
+            else:
+                self.logger.info('Install output processors for run in path `{}`'
+                                 .format(run_output.basepath))
+            pm = ProcessorManager()
```
setrofim: Should instantiate with `loader=config.plugin_cache` here.

Rahlubenru: Hm, OK, will do. I thought I had tried it and it failed on the install so I reverted but very preliminary testing just now suggests otherwise.

Rahlubenru: Yes, it fails if you specify a processor on the command line that's not in the config because there's no `get_output_processor` function. I can try to work around that if you like?

setrofim: Should be fixed by https://github.com/ARM-software/workload-automation/pull/713

Rahlubenru:
Ok, made that change too. Also I presume I'm doing something slightly wrong in git that's removing the previous code comments or would a squash always do that anyway? Regardless that change is made and I think it's in an acceptable state again.

setrofim:
`squash` will always prompt you to write a new commit message, and it will have messages for all commits being squashed in comments (so they won't appear in the final message, unless you specifically uncomment).

You can specify `fixup` instead of `squash` in which case the messages for the "fixup" commits will be ignored and the base commit's message will be kept.

Rahlubenru:
I meant the ones in the review here where you comment against the source.

setrofim:
Oh that's nothing to do with git squashing. That's just how GitHub does things -- when the line that was commented on changes, it hides the comment (you can click on "show outdated" to view it).

Rahlubenru:
Aha, thank you.

PR #702, opened 2018-03-27
GFXBench: Creating a WA3 workload to execute a subset of GFXBench ben…

scojac01:
…chmarks


setrofim:
Please change the commit message to match best practice: https://chris.beams.io/posts/git-commit/

wa/workloads/gfxbench/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+#
+import re
+
+from wa import ApkUiautoWorkload, WorkloadError
+
+
+class Gfxbench(ApkUiautoWorkload):
+
+    name = 'gfxbench'
+    package_names = ['net.kishonti.gfxbench.gl.v50000.corporate']
```
setrofim: Is this specific to the corporate (in which case, the workload should be renamed to reflect that), or will this work with the free version (in which case, its package should be added here)?

wa/workloads/gfxbench/uiauto/app/src/main/java/com/arm/wa/uiauto/gfxbench/UiAutomation.java
```diff
@@ -0,0 +1,240 @@
...
+                    .className("android.widget.TextView"));
+                if (sync.exists()){
+                    UiObject data =
+                        mDevice.findObject(new UiSelector().resourceId("android:id/button1")
+                            .className("android.widget.Button"));
+                    data.click();
+                }
+
+        UiObject home = 
+            mDevice.findObject(new UiSelector().resourceId("com.glbenchmark.glbenchmark27:id/tabbar_back")
```
setrofim: seems to refer to the to the GLB2.7 package; is this intentional?

wa/workloads/gfxbench/uiauto/app/src/main/java/com/arm/wa/uiauto/gfxbench/UiAutomation.java
```diff
@@ -0,0 +1,240 @@
...
+                .className("android.widget.LinearLayout"));
+            home.waitForExists(300000);
+    }
+
+    public void runBenchmark() throws Exception {
+        UiScrollable list = new UiScrollable(new UiSelector().scrollable(true));
+
+        //Disable high level tests
+        mDevice.click(840,1200);
+        UiObject highlevel =
```
setrofim: Should factor out repetitive code  above into separate disable/enable functions.

wa/workloads/gfxbench/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+    regex_matches = [re.compile(r'Car Chase score \(([\d.]+)'),
+                     re.compile(r'Car Chase Offscreen score \(([\d.]+)'),
+                     re.compile(r'Manhattan 3.1 score \(([\d.]+)'),
+                     re.compile(r'1080p Manhattan 3.1 Offscreen score \(([\d.]+)'),
+                     re.compile(r'1440p Manhattan 3.1 Offscreen score \(([\d.]+)'),
+                     re.compile(r'Tessellation score \(([\d.]+)'),
+                     re.compile(r'Tessellation Offscreen score \(([\d.]+)')]
```
marcbonnici: Might this be similar to the antutu workload in that some tests are not supported on all devices resulting in no score being present? 
You allow for not being able to convert the expected score to a float below however your regexes will only match on numbers.

wa/workloads/gfxbench/uiauto/app/src/main/java/com/arm/wa/uiauto/gfxbench/UiAutomation.java
scojac01: This is correct.

wa/workloads/gfxbench/__init__.py
scojac01: We only get numbers as a result so is that a problem? I thought I had matched the antutu workload behaviour here?

marcbonnici: If you can only get number in that location that will be fine, although for the antutu workload on some older devices not all the tests were supported and therefore produced output like `'Tessellation score Not Supported` causing the workload to fail, therefore in https://github.com/ARM-software/workload-automation/pull/684 we change it to allowing matching of any output following the predefined statements and then handle any invalid conversions rather than only matching on numbers directly.

wa/workloads/gfxbench/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+    name = 'gfxbench-corporate'
+    package_names = ['net.kishonti.gfxbench.gl.v50000.corporate']
+    regex_matches = [re.compile(r'Car Chase score \(.+)'),
+                     re.compile(r'Car Chase Offscreen score \(.+)'),
+                     re.compile(r'Manhattan 3.1 score \(.+)'),
+                     re.compile(r'1080p Manhattan 3.1 Offscreen score \(.+)'),
+                     re.compile(r'1440p Manhattan 3.1 Offscreen score \(.+)'),
+                     re.compile(r'Tessellation score \(.+)'),
+                     re.compile(r'Tessellation Offscreen score \(.+)')]
```
marcbonnici: These regex strings have a `\` before the `(.+)` causing a syntax error.

wa/workloads/gfxbench/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+    parameters = [
+        Parameter('subtest_x', kind=int, default=840,
+                  description="""
+                  The x co-ordinate of the test selection menu required
+                  """),
+        Parameter('subtest_y', kind=int, default=1200,
+                  description="""
+                  The y co-ordinate of the test selection menu required
+                  """)
```
setrofim: Just noticed these. Why are you doing it this way, rather than specifying the names of the sub-tests to run? The coordinates are going to  vary all the time, and it's not something that the user of the test should need to worry about.

scojac01: There are no element names given for this menu. This is the only way to do it as far as I can see.

setrofim: Presumably, the entries in the menu are still distinct elements? So you can still select them based on their relative position to each other within the menu? You'd have to maintain a name --> position index mapping your code, in that case, but that would still be much preferable to expecting the user to calculate the coordinates each time.

scojac01: They are not. 

image

setrofim: OK what about reading the bounds of the object  with id `main_circleControl` and then calculating the x/y coordinates as offsets from its center? You'd need to check how this elem scales in hoziontal mode, but if its in anyway sensible, calculating the offset from the center of control as a proportion of its total size should work.

wa/workloads/gfxbench/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+        logcat_file = context.get_artifact_path('logcat')
+        with open(logcat_file) as fh:
+            for line in fh:
+                for regex in self.regex_matches:
+                    match = regex.search(line)
+                    if match:
+                        try:
+                            result = float(match.group(1))
+                        except ValueError:
+                            result = 'NaaN'
```
marcbonnici: This should be `NaN` to be a valid float

wa/workloads/gfxbench/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+                     re.compile(r'Tessellation Offscreen score (.+)')]
+    description = '''
+    Execute a subset of graphical performance benchmarks
+
+    Test description:
+    1. Open the gfxbench application
+    2. Execute Car Chase, Manhattan and Tessellation benchmarks
+
+    '''
+
```
marcbonnici: The default workload timeout seems to be too short to allow for the tests to complete. 
Add an `__init__`  method and set `self.gui.timeout` to user configurable parameter with a longer default?

wa/workloads/gfxbench/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+    def __init__(self)
+        self.gui.timeout = self.timeout
```
marcbonnici: Also should add in the user parameter to allow configuration of the time-out.
e.g. 
```
parameters = [
  Parameter('timeout', kind=int, default=X, description=X),
]

PR #708, opened 2018-03-29
daq: fix resistor_values parameters list

vingu-linaro:
resistor_values is an array of float and not int as the shunt resistors
are most of the time lower than 1 Ohm.

With kind=list_of_ints, all resistor values are rounded to 0.

Signed-off-by: Vincent Guittot <vincent.guittot@linaro.org>

### DIFF
--- a/wa/instruments/energy_measurement.py
+++ b/wa/instruments/energy_measurement.py
@@ -20,6 +20,8 @@ from collections import defaultdict
 import os
 import shutil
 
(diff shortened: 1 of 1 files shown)


setrofim:
Please change to `list_of_numbers` instead.

vingu-linaro:
I have pushed a new version that uses list_of_numbers.  I have also added another fix for sample_rate_hz that I found in the meantime while testing NI DAQ

PR #712, opened 2018-04-03
workloads: add recentfling

JaviMerino:
Port recentfling from WA2.

### DIFF
--- a/wa/workloads/recentfling/__init__.py
+++ b/wa/workloads/recentfling/__init__.py
@@ -1,0 +1,133 @@
+#    Copyright 2013-2015 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#pylint: disable=E1101,W0201
+
+import os
+import re
+
+from wa import Workload, Parameter, File
+from wa.framework.exception import WorkloadError, TargetError
+from wa.utils.exec_control import once
+
+
+class Recentfling(Workload):
+
+    name = 'recentfling'
+    description = """
+    Tests UI jank on android devices.
+
+    For this workload to work, ``recentfling.sh`` and ``defs.sh`` must be placed
+    in ``~/.workload_automation/dependencies/recentfling/``. These can be found
+    in the `AOSP Git repository <https://android.googlesource.com/platform/system/extras/+/master/tests/workloads>`_.
+
+    To change the apps that are opened at the start of the workload you will need
+    to modify the ``defs.sh`` file. You will need to add your app to ``dfltAppList``
+    and then add a variable called ``{app_name}Activity`` with the name of the
+    activity to launch (where ``{add_name}`` is the name you put into ``dfltAppList``).
+
+    You can get a list of activities available on your device by running
+    ``adb shell pm list packages -f``
+    """
+    supported_platforms = ['android']
+
+    parameters = [
+        Parameter('loops', kind=int, default=3,
+                  description="The number of test iterations."),
+        Parameter('start_apps', kind=bool, default=True,
+                  description="""
+                  If set to ``False``,no apps will be started before flinging
+                  through the recent apps list (in which the assumption is
+                  there are already recently started apps in the list.
+                  """),
+        Parameter('device_name', kind=str, default=None,
+                  description="""
+                  If set, recentfling will use the fling parameters for this
+                  device instead of automatically guessing the device.  This can
+                  also be used if the device is not supported by recentfling,
+                  but its screensize is similar to that of one that is supported.
+
+                  For possible values, check your recentfling.sh.  At the time
+                  of writing, valid values are: 'shamu', 'hammerhead', 'angler',
+                  'ariel', 'mtp8996', 'bullhead' or 'volantis'.
+                  """),
+    ]
+
+    @once
+    def initialize(self, context):  # pylint: disable=no-self-use
+        if self.target.get_sdk_version() < 23:
+            raise WorkloadError("This workload relies on ``dumpsys gfxinfo`` \
+                                 only present in Android M and onwards")
+
+        defs_host = context.resolver.get(File(self, "defs.sh"))
+        Recentfling.defs_target = self.target.install(defs_host)
+        recentfling_host = context.resolver.get(File(self, "recentfling.sh"))
+        Recentfling.recentfling_target = self.target.install(recentfling_host)
+
+    def setup(self, context):
+        args = '-i {} '.format(self.loops)
+        if not self.start_apps:
+            args += '-N '
+        if self.device_name is not None:
+            args += '-d {}'.format(self.device_name)
+
+        self.cmd = "echo $$>{workdir}/pidfile; cd {bindir}; exec ./recentfling.sh {args}; rm {workdir}/pidfile".format(
+            workdir=self.target.working_directory,
+            bindir=self.target.executables_directory, args=args)
+
+        self._kill_recentfling()
+        self.target.ensure_screen_is_on()
+
+    def run(self, context):
+        self.output = ""
+        try:
+            self.output = self.target.execute(self.cmd, timeout=120)
+        except KeyboardInterrupt:
+            self._kill_recentfling()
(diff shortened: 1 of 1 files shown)


wa/workloads/recentfling/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+                  also be used if the device is not supported by recentfling,
+                  but its screensize is similar to that of one that is supported.
+
+                  For possible values, check your recentfling.sh.  At the time
+                  of writing, valid values are: 'shamu', 'hammerhead', 'angler',
+                  'ariel', 'mtp8996', 'bullhead' or 'volantis'.
+                  """),
+    ]
+
+    def initialise(self, context):  # pylint: disable=no-self-use
```
setrofim: Should be the international spelling: `initialize`

wa/workloads/recentfling/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def run(self, context):
+        args = '-i {} '.format(self.loops)
+        if not self.start_apps:
+            args += '-N '
+        if self.device_name is not None:
+            args += '-d {}'.format(self.device_name)
+        cmd = "echo $$>{dir}/pidfile; cd {bindir}; exec ./recentfling.sh {args}; rm {dir}/pidfile"
+        cmd = cmd.format(args=args, dir=self.target.working_directory,
+                         bindir=self.target.executables_directory)
```
setrofim: Everything up to this point should be moved into the `setup`; `run` should be kept as small  as possible.

wa/workloads/recentfling/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+        self.recentfling_target = self.target.install(self.recentfling_host)
+        self._kill_recentfling()
+        self.target.ensure_screen_is_on()
+
+    def run(self, context):
+        args = '-i {} '.format(self.loops)
+        if not self.start_apps:
+            args += '-N '
+        if self.device_name is not None:
+            args += '-d {}'.format(self.device_name)
```
setrofim: `device_name` does not have a default value and was never validated. Is this something that can be `initialize`'d from the target? (IIRC, this should be in the Android props).

wa/workloads/recentfling/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+    def initialise(self, context):  # pylint: disable=no-self-use
+        if context.target.get_sdk_version() < 23:
+            raise WorkloadError("This workload relies on ``dumpsys gfxinfo`` \
+                                 only present in Android M and onwards")
+
+    def setup(self, context):
+        self.defs_host = context.resolver.get(File(self, "defs.sh"))
+        self.defs_target = self.target.install(self.defs_host)
+        self.recentfling_host = context.resolver.get(File(self, "recentfling.sh"))
+        self.recentfling_target = self.target.install(self.recentfling_host)
```
setrofim: The above should be moved into `initialize` as this only needs to be done once per run.

wa/workloads/recentfling/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+                    count = 0
+                for metric in zip(group_names, match.groups()):
+                    context.result.add_metric(metric[0],
+                                              metric[1],
+                                              None,
+                                              classifiers={"loop": count or "Average"})
+
+    def teardown(self, context):
+        self.target.uninstall_executable(self.recentfling_target)
+        self.target.uninstall_executable(self.defs_target)
```
setrofim: If installation is moved into `initialize`, as suggested above, this should correspondingly be done inside `finalize`.

wa/workloads/recentfling/__init__.py
JaviMerino: If you don't pass "-d" to `recentfling.sh`, it guesses the device name from the Android props.  The idea of `device_name` is that you can override this for devices that are not supported by recentfling but their screensize is similar to that of one that is supported.  See the `device_name` description for more info.  I've given it a default value of None.

setrofim: Ah, sorry, I missed that. Thanks for the explanation.

wa/workloads/recentfling/__init__.py
JaviMerino: Why `setup()` and not `initialize()`?  If we are going to move it, we may as well put it in `initialize()`.

JaviMerino:
Latest update (767b15e) addresses all the review comments.

wa/workloads/recentfling/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        if self.device_name is not None:
+            args += '-d {}'.format(self.device_name)
+
+        self.cmd = "echo $$>{workdir}/pidfile; cd {bindir}; exec ./recentfling.sh {args}; rm {workdir}/pidfile".format(
+            workdir=self.target.working_directory,
+            bindir=self.target.executables_directory, args=args)
+        self.defs_host = context.resolver.get(File(self, "defs.sh"))
+        self.defs_target = self.target.install(self.defs_host)
+        self.recentfling_host = context.resolver.get(File(self, "recentfling.sh"))
+        self.recentfling_target = self.target.install(self.recentfling_host)
```
marcbonnici: I might suggest splitting the discovery of assets and installation into a separate method which is decorated with the ``@once`` decorator and called here instead?
This is to prevent the scripts being repeatedly deployed for multiple iterations. (Note: in that case these attributes would need to be set as class attributes instead.)

wa/workloads/recentfling/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+            self.output = self.target.execute(self.cmd, timeout=120)
+        except KeyboardInterrupt:
+            self._kill_recentfling()
+            raise
+
+    def update_output(self, context):
+        loop_group_names = ["90th Percentile", "95th Percentile", "99th Percentile", "Jank", "Jank%"]
+        count = 0
+        p = re.compile("Frames: \d+ latency: (?P<pct90>\d+)/(?P<pct95>\d+)/(?P<pct99>\d+) Janks: (?P<jank>\d+)\((?P<jank_pct>\d+)%\)")
+        for line in self.output.strip().splitlines():
```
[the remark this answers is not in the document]
JaviMerino: Ok, I have added this to the beginning of the method:

```
if not self.output:
    return
```

wa/workloads/recentfling/__init__.py
JaviMerino: What if I decorate `initialize()` and `finalize()` with `@once` instead?  Wouldn't that achieve the same?

marcbonnici: You are also doing your args manipulation and cmd generation in ``initialize`` so if you decorated the entire method you would not be able to have multiple configurations of the workload. As you're not doing anything else in your ``finalize`` method this should be fine to decorate.

wa/workloads/recentfling/__init__.py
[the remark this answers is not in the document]
marcbonnici: Thanks, just checking you have also initialized ``self.output`` to ``None`` somewhere as well?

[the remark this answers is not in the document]
JaviMerino: Yup, at the beginning of `run()`.  Looking at other workloads, shouldn't hwuitest and rt_app do a similar thing?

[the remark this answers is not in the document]
marcbonnici: Cool, and yes that's a good point I'll update them as well thanks.

wa/workloads/recentfling/__init__.py
JaviMerino: True, we are doing arg manipulation in `initialize()`.  @setrofim suggested putting it in `setup()`, but I decided to move it to `initialize()`.  Now I see why I was wrong.  I'll leave the arg manipulation in `setup()` and let `initialize()` be a method that only needs to be called once.

wa/workloads/recentfling/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+    @once
+    def initialize(self, context):  # pylint: disable=no-self-use
+        if self.target.get_sdk_version() < 23:
+            raise WorkloadError("This workload relies on ``dumpsys gfxinfo`` \
+                                 only present in Android M and onwards")
+
+        defs_host = context.resolver.get(File(self, "defs.sh"))
+        self.defs_target = self.target.install(defs_host)
+        recentfling_host = context.resolver.get(File(self, "recentfling.sh"))
+        self.recentfling_target = self.target.install(recentfling_host)
```
marcbonnici: As this is now only executed once for all instances of the workload, ``self.recentfling_target`` and ``self.defs_target`` should be set as class attributes i.e ``Recentfling.recentfling_target = ...``

JaviMerino:
Latest update 5e870a92bdace3d5c23669f5c0c09491b283fafa makes `recentfling_target` and `defs_target` class attributes.

PR #714, opened 2018-04-04
workloads: Add manual workload

setrofim:
Port manual workload from WA2. This workload allows using WA to
configure and instrument the target, while letting the user perform
arbitrary actions in place of running a fixed workload.

### DIFF
--- a/wa/workloads/manual/__init__.py
+++ b/wa/workloads/manual/__init__.py
@@ -1,0 +1,95 @@
+#    Copyright 2013-2015 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


wa/workloads/manual/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+                  on devices API level 23+.
+                  """),
+    ]
+
+    def validate(self):
+        if self.duration is None:
+            if self.user_triggered is None:
+                self.user_triggered = True
+            elif self.user_triggered is False:
+                self.duration = self.default_duration
```
marcbonnici: Is ``self.default_duration`` defined somewhere?

setrofim: erm, it is now :)

good catch!

PR #717, opened 2018-04-05
Fix workload early termination

marcbonnici:


wa/workloads/rt_app/__init__.py
```diff
@@ -172,6 +172,7 @@ def setup(self, context):
         self.timeout = self.duration + time_buffer
 
     def run(self, context):
+        self.output = None
```
setrofim: I would move this into `setup`

marcbonnici: Updated.

PR #718, opened 2018-04-05
framework/getters: Ignore connection errors with HTTP getter

marcbonnici:
If the http getter is used but a connection is not avalible, when
attempting to fetch the resource index an error will be raised. Now we
ignore the error allowing for the remaining getters to be used.


marcbonnici:
Updated

Issue #704, opened 2018-03-29
Port facebook workload to WA3


setrofim:

JaviMerino:
FTR, I'm working on this in the facebook branch of [JaviMerino/workload-automation](https://www.github.com/JaviMerino/workload-automation).

PR #720, opened 2018-04-06
Port Facebook from WA2

JaviMerino:


wa/workloads/facebook/uiauto/app/src/main/java/com/arm/wa/uiauto/facebook/UiAutomation.java
```diff
@@ -0,0 +1,190 @@
...
+    /*
+     * The 'runWorkload' method implements the following activities
+     * Login to facebook account.
+     * Send a message.
+     * Check latest notification.
+     * Search particular user account and visit his/her facebook account.
+     * Update the facebook status
+     */
+    @Test
+    public void runWorkload() throws Exception {
```
setrofim: Could you break this up into several methods each performing a single logical step of execution, and the run method should be just a simple sequence of steps. That will make it a lot easier to maintain this.

wa/workloads/facebook/__init__.py
```diff
@@ -0,0 +1,44 @@
+#    Copyright 2013-2015 ARM Limited
```
marcbonnici: The commit message references the wrong issue, it should be https://github.com/ARM-software/workload-automation/issues/704

JaviMerino:
Thanks for the review. I am afraid I do not have time to work on getting Facebook in WA3 any more, so I do not think I am able to act on it.

PR #725, opened 2018-04-10
wlauto/utils/ssh.py: Fix set terminal window size

alimon:
Ensure that the terminal window size is set to 500x200, if not long
commands (prompt + command > 80 chars) will fail because the pexpect
search to match the command string in order to get the result and the
command is wrapped to first 80 chars.

For example, when check a file and executes:

...
if [ -f '/sys/kernel/debug/sched_features' ]; then echo 1; else echo 0; fi
...
File
\"/usr/local/lib/python2.7/dist-packages/wlauto/common/linux/device.py\",
line 228, in get_properties
if self.is_file(propfile):
File
\"/usr/local/lib/python2.7/dist-packages/wlauto/common/linux/device.py\",
line 215, in is_file
return boolean(output.split()[-1])  # pylint: disable=maybe-no-member

IndexError(list index out of range)
...

In order to fix this scenario enables checkwinsize in the shell and use
stty to set the new terminal window size.

Signed-off-by: Aníbal Limón <anibal.limon@linaro.org>

### DIFF
--- a/wlauto/utils/ssh.py
+++ b/wlauto/utils/ssh.py
@@ -66,8 +66,14 @@ def ssh_get_shell(host, username, password=None, keyfile=None, port=None, timeou
                 raise DeviceError(message.format(host))
             time.sleep(5)
 
+    conn.sendline('shopt -s checkwinsize')
+    conn.prompt()
     conn.setwinsize(500,200)
     conn.sendline('')
+    conn.prompt()
+    conn.sendline('stty rows 500')
+    conn.prompt()
+    conn.sendline('stty cols 200')
     conn.prompt()
     conn.setecho(False)
     return conn


alimon:
@setrofim Ok, so need to i send other PR targeting next?

alimon:
Ok it will  make no difference testing the command result.

setrofim:
> Ok, so need to i send other PR targeting next?

Well, as I was saying, I'm not sure this is the Right Way(tm) of fixing this, and I don't mind it being merged into `master` because that's going away soon. But yes, would need to submit this fix into `devlib` (a different repo) in order for it to be available on `next`.

> About to the assumption of stty, i can make a test to search if is available (which) previous try to use it, how it sounds?

You don't need to do that -- `stty` is available as an applet of `busybox` which is guaranteed to be available because we're deploying it. You would just need to perform this operation after busybox has been deployed (which happens very early on).

There is also a call to `conn.setwinsize()` in the code where you've implemented your fix, which is specifically there to fix this issue. Clearly, it doesn't seem to be working in your case. I'd rather understand why that is the case and fix *that* before pursuing the `stty` approach...

alimon:
@setrofim The issue looks to be at level of pexpect using spawn.setwinsize and in some how it doesn't send the signal to ssh client (SIGWICH) for resize. I'm not an expert on how pexpect works.

If you can merge into master will be great and wait -next for the proper fix.

PR #733, opened 2018-04-26
framework/config: add disable all augmentations

Rahlubenru:
Add glyph `~~` to be used in the configuration to disable all
augmentations specified at a lower configuration level.

Also useable with the `--disable` argument to disable all augmentations.

### DIFF
--- a/wa/framework/configuration/core.py
+++ b/wa/framework/configuration/core.py
@@ -914,6 +914,7 @@ class JobGenerator(object):
         self._enabled_processors = toggle_set()
         self._read_enabled_instruments = False
         self._read_enabled_processors = False
+        self._disable_all_augmentations = False
         self.disabled_augmentations = set()
 
         self.job_spec_template = obj_dict(not_in_dict=['name'])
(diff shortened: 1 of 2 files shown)


setrofim:
There is no mention of `--disable` in the feature request. It's intended to quickly temporarily disable specific instrumentation, e.g. during interactive testing. The use case for it  is orthogonal from the one stated in the feature request (ensure a "clean" starting config for your agenda).

However, since the augmentations specified with `--disable` flag are also managed as a `toggle_set`, implementing the `~~` behaviour for the latter should  make the former "just work" without the need for any additional changes -- specifying `--disable ~`  on the command line should result in `toggle_set(['~~'])` being merged into augmentations for all jobs, effecting the disabling of every augmentation.

Rahlubenru:
I feel I should apologise that you felt that was the case so, sorry.

From my position I can't be too disappointed, it was still an educational experience and that was my main reason for attempting this in the first place so thank you for the opportunity and the feedback. Plus I get to see how you solved it, I've been busy this week but will take a proper look when I get a free moment.

PR #735, opened 2018-04-27
workload/rt-app: Better JSON processing error reporting

Sticklyman1936:
The rt-app workload parses the rt-app JSON file in order to override
some options (specifically duration). Previously, if the JSON was
syntactically incorrect, an uninformative ValueError was raised. Now
we raise a ConfigError with appropriate message prompting the user to
fix the file.

### DIFF
--- a/wa/workloads/rt_app/__init__.py
+++ b/wa/workloads/rt_app/__init__.py
@@ -21,7 +21,7 @@ from collections import OrderedDict
 from subprocess import CalledProcessError
 
 from wa import Workload, Parameter, Executable, File
-from wa.framework.exception import WorkloadError, ResourceError
+from wa.framework.exception import WorkloadError, ResourceError, ConfigError
 from wa.utils.misc import check_output
 from wa.utils.exec_control import once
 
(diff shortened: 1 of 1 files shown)


wa/workloads/rt_app/__init__.py
```diff
@@ -231,7 +231,13 @@ def _load_json_config(self, context):
         config_file = self._generate_workgen_config(user_config_file,
                                                     context.output_directory)
         with open(config_file) as fh:
-            config_data = json.load(fh, object_pairs_hook=OrderedDict)
+            try:
+                config_data = json.load(fh, object_pairs_hook=OrderedDict)
+            except ValueError:
+                # We were not able to parse the JSON file. Raise an informative error.
+                msg = "Failed to parse {}. Please make sure it is valid JSON."
+                raise ConfigError(msg.format(user_config_file))
```
setrofim: Maybe format the `ValueError` into the message as well (if it gives more information about the syntax error)?

Sticklyman1936: The ValueError message was:

`ValueError(No JSON object could be decoded)`

Which does not seem to be overly informative so I personally feel that it is not useful.

setrofim: Agreed. I'd hoped it'd at least give a line number or something. This is fine in that case.

PR #741, opened 2018-04-30
Do not lose early logging

setrofim:
- Buffer early log records before a log file becomes available, so that they are
not lost.
- Add `indentcontext` to remove the need for `try`/`finally` guards for ensuring dedenting.
- Remove superfluous "Skipping..." entries from `pluginloader` logging.

### DIFF
--- a/wa/utils/log.py
+++ b/wa/utils/log.py
@@ -21,6 +21,7 @@ import os
 import string
 import subprocess
 import threading
+from contextlib import contextmanager
 
 import colorama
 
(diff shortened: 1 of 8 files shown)


wa/utils/log.py
```diff
@@ -148,6 +149,13 @@ def dedent():
     _indent_level -= 1
 
 
+@contextmanager
+def indentcontext():
+    indent()
+    yield
+    dedent()
```
marcbonnici: Should this be in a try/finally block in case of errors?

setrofim: No, this is ensured by the context machinery.

setrofim: OK, I was wrong, `contextmanager` decorator *wont* ensure that by default. Will fix.

setrofim:
Rebased onto latest `next` and fixed `contextmanager` issue.

PR #661, opened 2018-02-27
PcMark: Porting the workload to WA3

scojac01:

PR #744, opened 2018-05-09
PCMark: Porting the PCMark workload to WA3

scojac01:


scojac01:
Had to create a new PR as the old one wasn't being updated. Sorry.

wa/workloads/pcmark/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        super(PcMark, self).__init__(target, **kwargs)
+        self.gui.timeout = 1500
+
+    def extract_results(self, context):
+        results_path = self.target.path.join(self.target.external_storage, "PCMark for Android")
+        result_file = self.target.execute('ls -1 \"' + results_path + '\"| tail -n 1')
+        result_file = result_file.rstrip()
+        result = (results_path + "/" + result_file)
+        self.target.pull(result, context.output_directory)
+        context.add_artifact('pcmark-result', result_file, kind='raw')
```
marcbonnici: `result_file` is the file path on the device rather than the host.

wa/workloads/pcmark/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __init__(self, target, **kwargs):
+        super(PcMark, self).__init__(target, **kwargs)
+        self.gui.timeout = 1500
+
+    def extract_results(self, context):
+        results_path = self.target.path.join(self.target.external_storage, "PCMark for Android")
+        result_file = self.target.execute('ls -1 \"' + results_path + '\"| tail -n 1')
+        result_file = result_file.rstrip()
+        result = (results_path + "/" + result_file)
```
marcbonnici: This should be done with `self.target.path.join`

wa/workloads/pcmark/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        self.target.pull(result, context.output_directory)
+        context.add_artifact('pcmark-result', result_file, kind='raw')
+        expected_results = len(self.regex_matches)
+        zf = zipfile.ZipFile(context.output_directory + "/" + result_file, 'r').read('Result.xml')
```
marcbonnici: This host path joining should be done with `os.path.join`

wa/workloads/pcmark/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        self.gui.timeout = 1500
+
+    def extract_results(self, context):
+        results_path = self.target.path.join(self.target.external_storage, "PCMark for Android")
+        result_file = self.target.execute('ls -1 \"' + results_path + '\"| tail -n 1')
+        result_file = result_file.rstrip()
+        result = (results_path + "/" + result_file)
+        self.target.pull(result, context.output_directory)
+        context.add_artifact('pcmark-result', result_file, kind='raw')
+        expected_results = len(self.regex_matches)
```
marcbonnici: The actual processing of the results file should be performed in the `update_output` method.

wa/workloads/pcmark/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    Known working APK version: 2.0.3716
+    '''
+
+    def __init__(self, target, **kwargs):
+        super(PcMark, self).__init__(target, **kwargs)
+        self.gui.timeout = 1500
+
+    def extract_results(self, context):
+        results_path = self.target.path.join(self.target.external_storage, "PCMark for Android")
+        result_file = self.target.execute('ls -1 \"' + results_path + '\"| tail -n 1')
```
marcbonnici: As @setrofim mentioned at https://github.com/ARM-software/workload-automation/pull/661#discussion_r172888210 this can be made more readable.

wa/workloads/pcmark/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        context.add_artifact('pcmark-result', result_file, kind='raw')
+        expected_results = len(self.regex_matches)
+        zf = zipfile.ZipFile(context.output_directory + "/" + result_file, 'r').read('Result.xml')
+        for line in zf.split('\n'):
+            for regex in self.regex_matches:
+                match = regex.search(line)
+                if match:
+                    scores = float(match.group(1))
+                    entry = regex.pattern
+                    entry = entry[:-9]
```
marcbonnici: Are the scores always guaranteed to be 8 digits long? Otherwise this could cause the entry names to be incorrectly extracted.

wa/workloads/pcmark/__init__.py
scojac01: result_file is just the file name (e.g "pcma-result-2018-05-09T16-09-11-699.zip") not the path on the host. 

result is the path and the filename. 

result_path is just the path.

wa/workloads/pcmark/__init__.py
scojac01: I tried using this suggestion but it didn't work.

wa/workloads/pcmark/__init__.py
scojac01: Results are guaranteed to be four digits long. This seems to work correctly?

wa/workloads/pcmark/__init__.py
setrofim: Could you elaborate on "didn't work"?

marcbonnici: I suspect this was due to the path not being escaped, I've submitted a fix for this https://github.com/ARM-software/devlib/pull/265. Would you be able to pull the latest version and try this again?

wa/workloads/pcmark/__init__.py
marcbonnici: Ah apologies, I misread that.

wa/workloads/pcmark/__init__.py
marcbonnici: Ok just wanted to check that older devices wouldn't have resulted in lower scores and therefore altered the name matching.

wa/workloads/pcmark/uiauto/app/src/main/java/com/arm/wa/uiauto/pcmark/UiAutomation.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Test
+    public void teardown() throws Exception{
+        unsetScreenOrientation();
+    }
+
+    //Swipe to benchmarks and back to initialise the app correctly
+    private void loadBenchmarks() throws Exception {
+        UiObject title = 
+            mDevice.findObject(new UiSelector().resourceId("com.futuremark.pcmark.android.benchmark:id/actionBarMenuItemText")
```
marcbonnici: I've just tested the workload with our nexus 10 / chromebook (tablet layout) and this element can't be found. Maybe use the menu bar itself instead? Resource-id: `android:id/action_bar` as this seems to be consistent across the devices although the class name for that element does change.

wa/workloads/pcmark/uiauto/app/src/main/java/com/arm/wa/uiauto/pcmark/UiAutomation.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Test
+    public void teardown() throws Exception{
+        unsetScreenOrientation();
+    }
+
+    //Swipe to benchmarks and back to initialise the app correctly
+    private void loadBenchmarks() throws Exception {
+        UiObject title = 
+            mDevice.findObject(new UiSelector().resourceId("android:id/action_bar_spinner")
```
marcbonnici: This is still failing to be found for me, I've attached the UI dump for comparison. 
[setup-error.tar.gz](https://github.com/ARM-software/workload-automation/files/1995503/setup-error.tar.gz)

scojac01: This looks very different to the application I am running. 

From your dump it looks like you can see Install correctly on first launch. This isn't the same behaviour that we see on mobile devices. I will add in some logic to handle this better.

marcbonnici: Ah ok, I've been testing this with version 2.0.3716 (the latest available the play store) just in case this is different to the version you're using.

PR #751, opened 2018-05-14
Parameter cleanup & add cpu mask type

marcbonnici:


wa/workloads/meabo/__init__.py
```diff
@@ -72,17 +72,19 @@ class Meabo(Workload):
             default=1048576,
         ),
         Parameter(
-            'nrow',
+            'num_row',
```
setrofim: I would pluralize this to `num_rows`; and similar blow.

marcbonnici: Updated

PR #763, opened 2018-05-23
Update Agenda Tests + Misc Fixes

marcbonnici:


wa/framework/execution.py
```diff
@@ -227,7 +227,8 @@ def add_event(self, message):
     def take_screenshot(self, filename):
         filepath = self._get_unique_filepath(filename)
         self.tm.target.capture_screen(filepath)
-        self.add_artifact('screenshot', filepath, kind='log')
```
setrofim: If `capture_screen()` failed, wouldn't that result in a `TargetError` and therefore never get to `add_artifiact` anyway?

wa/framework/execution.py
```diff
@@ -227,7 +227,8 @@ def add_event(self, message):
     def take_screenshot(self, filename):
```
setrofim: In the commit message:

"as metric" --> "as an artifact"

marcbonnici: Fixed

wa/framework/execution.py
marcbonnici: In the case of trying to take a screenshot from a `LinuxTarget` if scot is not installed then `capture_screen()` simply returns with a debug message printed.

PR #765, opened 2018-05-24
Implement metadata recording inside results.

setrofim:
Use this new capability to track android app version and resource hash
tracking.

### DIFF
--- a/wa/framework/workload.py
+++ b/wa/framework/workload.py
@@ -73,7 +73,7 @@ class Workload(TargetedPlugin):
         self.asset_files = []
         self.deployed_assets = []
 
-    def init_resources(self, context):
+    def init_resources(self, resolver):
         """
         This method may be used to perform early resource discovery and
         initialization. This is invoked during the initial loading stage and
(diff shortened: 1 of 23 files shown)


wa/framework/workload.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, target, **kwargs):
         self.asset_files = []
         self.deployed_assets = []
 
-    def init_resources(self, context):
+    def init_resources(self, resolver):
```
marcbonnici: Why does `init_resources` now take a resolver?

setrofim: Because it only needs a resolver, since that is all it uses, it doesn't need a full context (note: in practice, it will still get passed the context, since we're now using that as a resolver, but the name change here is to indicate that the supplied object is only expected to be able to get resources).

marcbonnici: In which case should this be updated in the subclasses of `Workload` as well to reflect the change?

setrofim: Good point. I have made the change.

marcbonnici: And also this probably means that the documentation for creating workloads and the workload templates need to be updated as well.

setrofim: Updated.

marcbonnici: Sorry I meant the create command workload templates and I just realised the resource resolution / deployment documentation will also needs changing.

setrofim: Ah, I didn't realise we had any :).  Will do...

setrofim: All done. Also spotted and fixed a minor inaccuracy elsewhere in resource docs.

PR #767, opened 2018-05-24
Reboot Policy Updates

marcbonnici:


wa/framework/execution.py
```diff
@@ -142,7 +142,6 @@ def start_job(self):
         job_output = init_job_output(self.run_output, self.current_job)
         self.current_job.set_output(job_output)
         self.update_job_state(self.current_job)
-        self.tm.start()
```
setrofim: the `stop()` should also be moved into the runner.

wa/framework/execution.py
```diff
@@ -486,6 +486,9 @@ def run_next_job(self, context):
             if self.context.reboot_policy.reboot_on_each_job:
                 self.logger.debug('Rebooting on new job.')
                 self.context.tm.target.reboot()
+            elif self.context.reboot_policy.reboot_on_each_spec and context.spec_changed:
+                self.logger.debug('Rebooting on new spec.')
```
setrofim: These messages should be upgraded to info, now that I think about it.

marcbonnici: Updated and added logging for initial reboot.

wa/framework/execution.py
marcbonnici: Fixed.

PR #775, opened 2018-05-30
Fixes to instrument priorities.

setrofim:
- Update table header in generated docs
- Use enum for signal priorities
- Use signal priorities for instrument decorators

### DIFF
--- a/doc/build_instrument_method_map.py
+++ b/doc/build_instrument_method_map.py
@@ -18,7 +18,8 @@ import sys
 import string
 from copy import copy
 
-from wa.framework.instrument import SIGNAL_MAP, Priority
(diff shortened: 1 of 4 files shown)


doc/build_instrument_method_map.py
```diff
@@ -37,7 +37,7 @@ def generate_instrument_method_map(outfile):
     signal_table = format_simple_table([(k, v) for k, v in SIGNAL_MAP.iteritems()],
                                        headers=['method name', 'signal'], align='<<')
     priority_table = format_simple_table(zip(Priority.names, Priority.values),
```
marcbonnici: This will also need to be updated to use signal priorities.

setrofim:
Comments addressed.

PR #785, opened 2018-06-05
fw/parsers: Ensure that a section is a valid dict

marcbonnici:
Ensure that all section entries are valid dicts otherwise raise an error
informing the user.


wa/framework/configuration/parsers.py
```diff
@@ -140,6 +140,9 @@ def _pop_sections(self, raw):
         sections = raw.pop("sections", [])
         if not isinstance(sections, list):
             raise ConfigError('Invalid entry "sections" - must be a list')
+        for section in sections:
+            if not isinstance(section, dict):
```
setrofim: A better check her would be `if not hasattr(section, 'items')`.

marcbonnici: Updated.

PR #786, opened 2018-06-06
Add target-params to Platform Tuple

sdpenguin:
Added support for specifying target parameters in the Platform tuple.

### DIFF
--- a/wa/framework/target/descriptor.py
+++ b/wa/framework/target/descriptor.py
@@ -416,26 +416,31 @@ ASSISTANTS = {
     'chromeos': ChromeOsAssistant
 }
 
-# name --> ((platform_class, conn_class), params_list, defaults)
+# name --> (((platform_class, conn_class), params_list, defaults), target_defaults)
 # Note: normally, connection is defined by the Target name, but
 #       platforms may choose to override it
+# Note: the target_defaults allows you to override common target_params for a
+# particular platform. Parameters you can override are in COMMON_TARGET_PARAMS
+# Example of overriding one of the target parameters: Replace last None with:
+# {'shell_prompt': CUSTOM__SHELL_PROMPT}
 PLATFORMS = {
-    'generic': ((Platform, None), COMMON_PLATFORM_PARAMS, None),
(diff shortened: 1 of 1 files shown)


sdpenguin:
Have updated the commit to remove JUNO_SHELL_PROMPT reference and make it compatible with the current tip of next.

sdpenguin:
Updated to stop the overwrite of existing target_params.

wa/framework/target/descriptor.py
```diff
@@ -416,26 +416,30 @@ def get_descriptions(self):
     'chromeos': ChromeOsAssistant
 }
 
-# name --> ((platform_class, conn_class), params_list, defaults)
+# name --> (((platform_class, conn_class), params_list, defaults), target_defaults)
 # Note: normally, connection is defined by the Target name, but
 #       platforms may choose to override it
+# Note: the target_defaults allows you to override common target_params for a
+# particular platform. Parameters you can override are in COMMON_TARGET_PARAMS
+import re
```
marcbonnici: This should not be imported here. Is it needed at all?

wa/framework/target/descriptor.py
```diff
@@ -416,26 +416,30 @@ def get_descriptions(self):
...
+    'generic': ((Platform, None), COMMON_PLATFORM_PARAMS, None, None),
     'juno': ((Juno, None), COMMON_PLATFORM_PARAMS + VEXPRESS_PLATFORM_PARAMS,
             {
                  'vemsd_mount': '/media/JUNO',
                  'baudrate': 115200,
                  'bootloader': 'u-boot',
                  'hard_reset_method': 'dtr',
-            }),
+            },
+            {'shell_prompt': DEFAULT_SHELL_PROMPT},),
```
marcbonnici: Do we need to override ``shell_prompt`` with its default value?

sdpenguin:
Have updated the code to stop the redundant override, and added an example of how to override in the comments.

wa/framework/target/descriptor.py
sdpenguin: Sorry. Whoops! Have removed it.

PR #791, opened 2018-06-11
Youtube: Updating to reflect changes in accessing options menu

scojac01:


setrofim:
This should probably be kept backwards-compatible with earlier versions, so should check for both strings.

wa/workloads/youtube/uiauto/app/src/main/java/com/arm/wa/uiauto/youtube/UiAutomation.java
```diff
@@ -122,7 +122,14 @@ public void clearFirstRunDialogues() throws Exception {
     }
 
     public void disableAutoplay() throws Exception {
-        clickUiObject(BY_DESC, "More options");
+        UiObject moresettings = 
+            mDevice.findObject(new UiSelector().descriptionContains("More Settings"));
```
setrofim: Should the string here be `"More options"`?

PR #806, opened 2018-06-15
utils/serializer: Handle empty file extensions

sdpenguin:
The ValueError does not given enough information in the case of an empty file
extension. Add a special case to handle files with no extension and
prompt the user.

### DIFF
--- a/wa/utils/serializer.py
+++ b/wa/utils/serializer.py
@@ -299,6 +299,12 @@ def load(s, fmt='json', *args, **kwargs):
 def _read_pod(fh, fmt=None):
     if fmt is None:
         fmt = os.path.splitext(fh.name)[1].lower().strip('.')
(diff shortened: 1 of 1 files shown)


wa/utils/serializer.py
```diff
@@ -305,6 +305,9 @@ def _read_pod(fh, fmt=None):
         return json.load(fh)
     elif fmt == 'py':
         return python.load(fh)
+    elif fmt == '':
+        # Special case of no given file extension
+        raise ValueError('Please specify a file format for {} e.g. ".yaml"'.format(getattr(fh, 'name', '<none>')))
```
setrofim: I think this would make more sense to check after the splitext call, because then you can also have a more specific message, e.g. "Could not determine format from file extension; Please specify it with 'fmt' parameter"

PR #807, opened 2018-06-15
Add Gmail offline mode option

joesavage:


wa/workloads/gmail/__init__.py
```diff
@@ -70,6 +75,22 @@ def init_resources(self, context):
...
+        if self.offline_mode and not self.target.is_rooted:
+            raise WorkloadError('This workload requires root to set up Gmail for offline usage.')
+
+    def setup_rerun(self):
+        super(Gmail, self).setup_rerun()
+        database_src = self.target.path.join(self.target.working_directory, 'mailstore.tar')
+        database_dst = self.target.path.join(self.target.package_data_directory, self.package, 'databases')
+        owner = self.target.execute("stat -c '%U' {}".format(database_dst), as_root=True).strip()
+        self.target.execute('tar -C {} -xvf {}'.format(database_dst, database_src), as_root=True)
+        self.target.execute('chown -R {}:{} {}'.format(owner, owner, database_dst), as_root=True)
```
setrofim: `stat`, `tar`, and `chown` should be invoked via `self.target.busybox`

wa/workloads/gmail/__init__.py
```diff
@@ -70,6 +75,22 @@ def init_resources(self, context):
...
+        self.gui.uiauto_params['offline_mode'] = self.offline_mode
         # Only accept certain image formats
         if os.path.splitext(self.test_image.lower())[1] not in ['.jpg', '.jpeg', '.png']:
             raise ValidationError('{} must be a JPEG or PNG file'.format(self.test_image))
+
+    def setup(self, context):
+        self.requires_network = not self.offline_mode
+        self.requires_rerun = self.offline_mode  # Need to restart the application after modifying its data directory
+        super(Gmail, self).setup(context)
+        if self.offline_mode and not self.target.is_rooted:
```
setrofim: This check should be done inside `initialize`

setrofim:
Also please add mailstore.tar to https://github.com/ARM-software/workload-automation-assets

joesavage:
The only thing with the `mailstore.tar` file is that I'm not totally sure to what degree its contents depend on the Google account that the device is configured with. My mailstore files, for instance, all share a base name of `mailstore.<my_test_email>@gmail.com.db` and presumably also have some account-specific data stored within them. As a result, I'm unsure how well my exact set of files will work in other environments.

setrofim:
This can be investigated by attempting to run on an alternative device. Gmail can have multiple accounts registered so, maybe it will be ok, as long as only specific databases are copied. Maybe it would also be possible to adjust the contents of the tarball (e.g. rename the db files) prior to deploying it. Do we know the format of the database? Maybe we can write records into existing database, rather than overwriting them.

Failing all that, the workload description should be expanded with instructions on how the user can provision a `mailstore.tar` on their own.

joesavage:
Having briefly glanced through the contents of the main SQLite mailstore database, writing new records into this doesn't seem out of the question. It would probably take a bit of time to get right, though. I'm not a Gmail or Android user myself, so it's not obvious to me — for instance — quite what the difference is between the "conversations" and "messages" tables. And I'd assume that any little piece of data out of place would probably result in a crash.

For the time being, testing on another device to see if my files will work seems like a good idea. Failing that, providing instructions for tarring up the right files seems like a reasonable thing to do.

wa/workloads/gmail/__init__.py
```diff
@@ -36,6 +36,12 @@ class Gmail(ApkUiautoWorkload):
     6. Enter text in the Compose field
     7. Click the Send mail button
 
+    To run the workload in offline mode, a 'mailstore.tar' file is required. In order to
+    generate such a file, Gmail should first be operated from an Internet-connected environment.
+    After this, the relevant database files can be found in the
+    '/data/data/com.google.android.gm/databases' directory. These files can then be archived to
+    produce a tarball using a command such as ``tar -cvf mailstore.tar -C /path/to/databases .``.
+
```
setrofim: Should also add that the file should be dropped into '~/.workload_automation/dependencies/gmail/` (creating that directory if it doesn't already  exist.

wa/workloads/gmail/__init__.py
```diff
@@ -50,17 +56,29 @@ class Gmail(ApkUiautoWorkload):
...
+                  This mode requires root and makes use of a tarball of email
+                  database files 'mailstore.tar' for the email account to be used.
+                  This file is extracted directly to the application's 'databases'
+                  directory at '/data/data/com.google.android.gm/databases'.
+                  '''),
     ]
 
-    # This workload relies on the internet so check that there is a working
-    # internet connection
-    requires_network = True
```
setrofim: Rather than removing this, should turn it into a property:

```
@property
def requires_network(self):
     return not self.offline_mode
```

PR #810, opened 2018-06-15
Add chrome workload

joesavage:
Relies on the same rerun infrastructure as #807 — these pull requests can be considered to overlap / be co-dependent to some extent.


joesavage:
Agreed that the dependencies should be portable here, will submit a corresponding pull request shortly.

PR #819, opened 2018-06-20
workloads/googlemaps: Add googlemaps workload

joesavage:
Adds a new `googlemaps` workload. In some sense I suppose this can be considered a port of the `googlemap` workload from WA2, but I've modified the workload name (adding the "s"), switched things over from revent to UIAutomator (using a totally new series of actions), and added the `offline_mode` option.

With regards to the dependencies for offline mode, I don't think we can trivially distribute these through `workload-automation-assets`. The `databases.tar` archive looks clean and shouldn't present any issues, but the `files.tar` archive is a bit more complicated. It contains binary files that might contain account-specific stuff that I haven't noticed and also includes map data that could present licensing issues.


wa/workloads/googlemaps/uiauto/app/src/main/java/com/arm/wa/uiauto/UiAutomation.java
```diff
@@ -0,0 +1,149 @@
...
+        directions.clickAndWaitForNewWindow(uiAutoTimeout);
+    }
+
+    @Test
+    public void runWorkload() throws Exception {
+        // Search for and select Cambridge train station
+        search("Cambridge Station CBG, England", "search_omnibox_text_box");
+        selectSearchResultContaining("[CBG]");
+
+        // Dismiss the tutorial dialog for place information if it appears
```
setrofim: Rather than commenting on what each code block does, please break them out into appropriately named functions and call them here. (e.g. see the implementation of googleplaybooks: https://github.com/ARM-software/workload-automation/blob/next/wa/workloads/googleplaybooks/uiauto/app/src/main/java/com/arm/wa/uiauto/googleplaybooks/UiAutomation.java)

This will make it easier to see the overall flow and to find the code for a particular action when it needs to be updated.

wa/workloads/googlemaps/uiauto/app/src/main/java/com/arm/wa/uiauto/UiAutomation.java
```diff
@@ -0,0 +1,149 @@
...
+        match.click();
+    }
+
+    public void getDirectionsFromLocation() throws Exception {
+        UiObject directions = mDevice.findObject(new UiSelector().resourceId(packageID + "placepage_directions_button"));
+        directions.clickAndWaitForNewWindow(uiAutoTimeout);
+    }
+
+    @Test
+    public void runWorkload() throws Exception {
```
setrofim: Please move he high-level methods corresponding to a particular stage above the method implementing specific actions.

wa/workloads/googlemaps/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+    which include Cambridge train station and Corpus Christi college.
+
+    Following this, the files of interest can be found in the 'databases' and 'files'
+    subdirectories of the '/data/data/com.google.android.apps.maps/' directory. The contents of
+    these subdirectories can be archived into tarballs using commands such as
+    ``tar -cvf databases.tar -C /path/to/databases .``. These 'databases.tar' and 'files.tar' archives
+    should then be placed in the '~/.workload_automation/dependencies/googlemaps' directory on your
+    local machine, creating this if it does not already exist.
+
+    Known working APK version: 7.72.2
```
marcbonnici: I can't seem to find this version of google maps. Although I tested with 9.72.2 and it seems to work correctly, was this the version you meant?

wa/workloads/googlemaps/uiauto/app/src/main/java/com/arm/wa/uiauto/UiAutomation.java
joesavage: Personally I'm not convinced that this will make the code easier to read, hence why I didn't write it this way, since in my mind this introduces unnecessary indirection (straight-line code is really easy to reason about!), but happy to make this change if you think it to be worthwhile.

joesavage:
> As a general thought (not necessary for this pull request), we could make using this and similar workloads easier by automating asset extraction. There would still be a manual element involved (e.g. ensuring the appropriate landmarks are present in the cached maps), but we can at least automate the creation of the tarball and the placing of it into the appropriate location on the host.

Good idea. Yeah, that would make a lot of sense.

wa/workloads/googlemaps/uiauto/app/src/main/java/com/arm/wa/uiauto/UiAutomation.java
```diff
@@ -0,0 +1,149 @@
...
+        pressBack();
+    }
+
+    @Test
+    public void teardown() throws Exception {
+        unsetScreenOrientation();
+    }
+
+    public void runApplicationSetup() throws Exception {
+        // Dismiss a dialog regarding real-time traffic updates
```
marcbonnici: Might need to add another check here for a "Getting the most from Google" splash screen which can be shown as the first thing upon a new installation of google maps. I've attached a screenshot and UX dump.
[wk1-googlemaps-1-attempt01.tar.gz](https://github.com/ARM-software/workload-automation/files/2118681/wk1-googlemaps-1-attempt01.tar.gz)

wa/workloads/googlemaps/__init__.py
joesavage: Oops, you're totally right it's 9.72.2.

wa/workloads/googlemaps/uiauto/app/src/main/java/com/arm/wa/uiauto/UiAutomation.java
setrofim: > straight-line code is really easy to reason about 

True, but breaking these blocks out into separate functions does not make the code any less straight-line -- there is still only a single path. And each of the code blocks is logically distinct from the other -- they do not need to be considered together to understand the code, so having them next to each other in the same function does not facilitate clarity.

On the other hand, in practice, maintaining the UI automation code comes down to making small changes to a specific operation in response to some UI change. Finding where in the code said operation takes place, especially for someone not already familiar with the automation flow, is made a lot easier if each operation is "folded" into its own function. 

> and I'm not re-using any of the code blocks here

Re-use is far from the only justification for functions.

joesavage: > Re-use is far from the only justification for functions.

Agreed, but what I was trying to get at is that I find the over-application of abstraction to be much more of a problem than the under-application in cases like this. That is, [I favour "style C" as described by John Carmack here](http://number-none.com/blow/john_carmack_on_inlined_code.html). In this case though we're talking about a tiny change in a tiny file so it doesn't really matter... I'll make some changes to clean it up a little and you can see what you think.

setrofim: He didn't really make a case for "style C" there, IMO. Most of his argument comes down to "long call chains are bad", and I agree with that; I mean, nobody wants [this](https://twitter.com/fioraaeterna/status/689908645476184068?lang=en). But that's not an argument for favouring style C over style B in a particular situation. Both of them have advantages and disadvantages, and in different situations, one might be more appropriate than the other.

In this case, what tends to happen, is:
- Workload fails
- You re-run and observe UI interactions
- You spot where it goes wrong
- You go to the automation source to fix the issue

Now, with style C, you have to scan through several screen-fulls of code to identify the part that needs fixing. With style B, you look at the outline of execution, and can immediately identify the relevant function (and yes, descriptive names are of paramount importance here), and can  jump to the relevant code right away.

wa/workloads/googlemaps/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+    @property
+    def requires_rerun(self):
+        # In offline mode we need to restart the application after modifying its data directory
+        return self.offline_mode
+
+    def __init__(self, target, **kwargs):
+        super(GoogleMaps, self).__init__(target, **kwargs)
+        if self.offline_mode:
+            self.deployable_assets = ['databases.tar', 'files.tar']
+            self.clean_assets = True
```
marcbonnici: This parameter should be `self.cleanup_assets`

joesavage: Oh, really? `clean_assets` is being used in the Gmail and Chrome workloads too (I copied it across from the original Gmail workload file if I remember correctly)

marcbonnici: Yea, it was a mistake in the original workload. I've submitted a PR to correct it in those other workloads.

wa/workloads/googlemaps/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+
+class GoogleMaps(ApkUiautoWorkload):
+
+    name = 'googlemaps'
+    description = '''
+    A workload to perform standard navigation tasks with Google Maps. This workload searches
+    for known locations, pans and zooms around the map, and follows driving directions
+    along a route.
+
+    To run the workload in offline mode, 'databases.tar' and 'files.tar' archives are required.
```
setrofim: Use monospace formatting (` `` `) for file names and paths.

joesavage: Sure, can do. FWIW I've used single quotes for filenames and path in the Gmail and Chrome workloads too so might want to update things there.

setrofim: Yup, I've missed those when reviewing. I've already submitted a fix for `chrome`.

PR #820, opened 2018-06-20
LMBench: Creating a new workload to execute the LMBench workload

scojac01:


wa/workloads/lmbench/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+        # self.test has been pre-validated, so this _should_ only fail if there's an abi mismatch
+        host_exe = context.resolver.get(Executable(self, abi, self.test))
+        self.target_exe = self.target.install(host_exe)
+        self.commands = []
+
+        setup_test = getattr(self, '_setup_{}'.format(self.test))
+        setup_test()
+
+    def run(self, context):
+        self.output = []
```
marcbonnici: This does not appear to be used.

wa/workloads/lmbench/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+        if self.parallelism is not None:
+            parts.append('-P {}'.format(self.parallelism))
+        if self.warmup is not None:
+            parts.append('-W {}'.format(self.warmup))
+        if self.repetitions is not None:
+            parts.append('-N {}'.format(self.repetitions))
+        return ' '.join(parts) + ' '
+
+    def extract_results(self, context):
+        results_path = self.target.path.join(self.target.external_storage, "devlib-target", "lmbench.output")
```
marcbonnici: Should this be using `self.target.working_directory` instead?

wa/workloads/lmbench/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+        self.commands = []
+
+        setup_test = getattr(self, '_setup_{}'.format(self.test))
+        setup_test()
+
+        for _ in range(self.times):
+            for command in self.commands:
+                self.target.execute(command, timeout=self.run_timeout)
+
+    def update_result(self, context):
```
marcbonnici: This method is not used in WA3, valid methods are  `extract_results` and `update_output`. Although it looks like the functionality here is duplicated in your `extract_results` method so this can probably just be removed.

wa/workloads/lmbench/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+        host_file = os.path.join(context.output_directory, 'lmbench.output')
+        target_file = self.target.path.join(self.target.working_directory, 'lmbench.output')
+        self.target.pull_file(target_file, host_file)
+        context.add_artifact('lmbench', host_file, 'data')
+
+    def extract_results(self, context):
+        results_path = self.target.path.join(self.target.external_storage, "devlib-target", "lmbench.output")
```
marcbonnici: Avoid using the hard-coded path `devlib-target`, this should probably be `self.target.path.join(self.target.working_directory, 'lmbench.output')` instead.

setrofim: or, more concisely, `self.target.get_workpath('lmbench.output')`

PR #837, opened 2018-06-27
Misc fixes

setrofim:
- Lengthen timeout for `gemm` and `conv` tests of `deepbench` to be the same as for sparse test. 
- Consider headers when calculating column widths inside `write_table`. This will ensure proper alignment in cases where the column header is wider than any of the column values.


wa/workloads/deepbench/__init__.py
```diff
@@ -104,7 +104,7 @@ def setup(self, context):
     def run(self, context):
```
marcbonnici: The previous commit seems to be included in this PR by accident.

setrofim: Erm it is, but only because I thought it has been committed already. I'll adjust  the PR title/text.

PR #845, opened 2018-06-29
fw/output: expose metadata

setrofim:
Expose result.metadata as a property of the output.

### DIFF
--- a/wa/framework/output.py
+++ b/wa/framework/output.py
@@ -81,6 +81,12 @@ class Output(object):
             return []
         return self.result.events
 
+    @property
+    def metadata(self):
+        if self.result is None:
+            return {}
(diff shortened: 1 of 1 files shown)


wa/framework/output.py
```diff
@@ -81,6 +81,12 @@ def events(self):
             return []
         return self.result.events
 
+    @property
+    def metadata(self):
```
marcbonnici: This should probably be added to the API documentation as well.

setrofim: It's already documented in the API as being there:

https://github.com/ARM-software/workload-automation/blob/next/doc/source/api/output.rst#id5

just was never actually exposed in the output -- whoops!

PR #873, opened 2018-07-11
Augmentation and resource getter config tracking

setrofim:


setrofim:
pep8 issues addressed.

PR #906, opened 2018-09-24
Googlephotos: Updating to work with latest version

scojac01:
Updating the googlephotos workload to work with app version 4.0.0.212659618


wa/workloads/googlephotos/uiauto/app/src/main/java/com/arm/wa/uiauto/googlephotos/UiAutomation.java
```diff
@@ -304,6 +304,9 @@ private void editPhotoColorTest() throws Exception {
         Iterator<Entry<String, PositionPair>> it = testParams.entrySet().iterator();
 
         clickUiObject(BY_ID, packageID + "edit", "android.widget.ImageView");
+        UiObject enhance = 
+            mDevice.findObject(new UiSelector().description("Enhance photo"));
+        enhance.click();
```
setrofim: What happens for the older app versions? Should check if `enhnace` exists and only try to click it if does.

Issue #907, opened 2018-09-28
pcmark fails on android Q


qais-yousef:
Using Linaro's latest Android snapshot images of aosp/master (Android Q), pcmark fails because there's a new pop on application launch that complains about "This app was built for an older version of Android...." which breaks uiAutomation with this splatter

```
ERROR            INSTRUMENTATION_STATUS: class=com.arm.wa.uiauto.pcmark.UiAutomation                                                                                                                        
ERROR            INSTRUMENTATION_STATUS: current=1                                                                                                                                                          
ERROR            INSTRUMENTATION_STATUS: id=AndroidJUnitRunner                                                                                                                                              
ERROR            INSTRUMENTATION_STATUS: numtests=1                                                                                                                                                         
ERROR            INSTRUMENTATION_STATUS: stream=                                                                                                                                                            
ERROR            com.arm.wa.uiauto.pcmark.UiAutomation:                                                                                                                                                     
ERROR            INSTRUMENTATION_STATUS: test=setup                                                                                                                                                         
ERROR            INSTRUMENTATION_STATUS_CODE: 1                                                                                                                                                             
ERROR            INSTRUMENTATION_STATUS: class=com.arm.wa.uiauto.pcmark.UiAutomation                                                                                                                        
ERROR            INSTRUMENTATION_STATUS: current=1                                                                                                                                                          
ERROR            INSTRUMENTATION_STATUS: id=AndroidJUnitRunner                                                                                                                                              
ERROR            INSTRUMENTATION_STATUS: numtests=1                                                                                                                                                         
ERROR            INSTRUMENTATION_STATUS: stack=android.support.test.uiautomator.UiObjectNotFoundException: Application has not loaded within the given time                                                 
ERROR                   at com.arm.wa.uiauto.pcmark.UiAutomation.loadBenchmarks(UiAutomation.java:78)                                                                                                       
ERROR                   at com.arm.wa.uiauto.pcmark.UiAutomation.setup(UiAutomation.java:49)                                                                                                                
ERROR                   at java.lang.reflect.Method.invoke(Native Method)                                                                                                                                   
ERROR                   at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)                                                                                             
ERROR                   at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)                                                                                              
[...]

PR #908, opened 2018-10-01
pcmark: update uiautomation to fix Android-Q breakage

qais-yousef:
A new popup appears when running pcmark on android Q that complains
about the app being built for an older version of android.

Since this popup will be temporary, the fix has to make sure not to
break in the future when this popup disappears or when the test is ran
on a compatible version of android.

To achieve this, we attempt to dismiss the popup and if we timeout we
silently carry on with the test assuming no popup will appear.

Signed-off-by: Qais Yousef <qais.yousef@arm.com>


qais-yousef:
When testing this I noticed that the timeout seemed longer than 5seconds. Not sure if there's another delay before that or there's an API change that makes the 5seconds wait appear longer.

qais-yousef:
This should fix issue #907

PR #919, opened 2018-10-24
Add comon method to dismiss Android Q popup

marcbonnici:
Add a base method to dismiss the Android Q warning message rather than having to implement on a per workload basis e.g. https://github.com/ARM-software/workload-automation/pull/917/ and https://github.com/ARM-software/workload-automation/pull/908/


wa/framework/uiauto/app/src/main/java/com/arm/wa/uiauto/BaseUiAutomation.java
```diff
@@ -547,6 +547,18 @@ public void selectGalleryFolder(String directory) throws Exception {
         }
     }
 
+    // If an an app is not designed for running on the latest version of android
+    // (currently Q) dissmiss the warning popup if present.
+    public void dismissAndroidVersionPopup() throws Exception {
+        UiObject acceptButton =
+            mDevice.findObject(new UiSelector().resourceId("android:id/button1")
```
setrofim: It seems like this might collide a lot. Is there some specific dialog text you could look for?

marcbonnici: Good point, I think the text is also standardised so will add a check for that.

PR #924, opened 2018-11-13
fw/target/info: Allow for missing values when reading from pod

marcbonnici:
Allow for deserializing of a TargetInfo pod when missing information.


setrofim:
Most of that information is not optional, so deserialisation *should* fail if it's missing.

marcbonnici:
Updated to only allow for select missing information.

setrofim:
I still don't understand why those values would be missing from a valid structure?

marcbonnici:
These values have been added in sequentially so when trying to load results generated with an older versions of WA they can be missing and therefore cause deserialization to fail.

PR #925, opened 2018-11-13
fw/workload: Add parameter to control if package data should be cleared.

marcbonnici:
Allow specifying that the packages data should not be cleared
before starting a workload. This is useful for benchmark apps that
otherwise need their assets redeploying each time.


setrofim:
Is this something that needs to be a parameter, or is it more appropriate for this to be a workload attribute?

marcbonnici:
Good point, in this case I think an attribute might work better. 
Updated.

PR #938, opened 2019-01-02
Workloads: Add support for UIBench

pietos01:
Add support for Android's UIBench (AOSP - [`frameworks/base/tests/UIBench`](https://github.com/aosp-mirror/platform_frameworks_base/tree/master/tests/UiBench)) suite of tests as a WA workload.

Please note that this is based on ARM-software/devlib#357.

### DIFF
--- a/wa/workloads/uibench/__init__.py
+++ b/wa/workloads/uibench/__init__.py
@@ -1,0 +1,52 @@
+#    Copyright 2013-2019 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from wa import Parameter, ApkWorkload
+
+
+class Uibench(ApkWorkload):
+
+    name = 'uibench'
+    description ="""
+        Runs a particular activity of the UIBench_ workload suite. The suite
+        is provided by Google as a testbench for the Android UI.
+
+        .. _UIBench: https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/tests/UiBench/
+    """
+    package_names = ['com.android.test.uibench']
+    loading_time = 1
+
+    parameters = [
+        Parameter('activity', kind=str,
+                  description="""
+                  The UIBench activity to be run. Each activity corresponds to
+                  a test. If this parameter is ignored, the application is
+                  launched in its main menu. Please note that the available
+                  activities vary between versions of UIBench (which follow
+                  AOSP versioning) and the availability of the services under
+                  test may depend on the version of the target Android. We
+                  recommend using the APK of UIBench corresponding to the
+                  Android version, enforced through the ``version`` parameter to
+                  this workload.
+                  """),
+        Parameter('duration', kind=int, default=10,
+                  description="""
+                  As activities do not finish, this workload will terminate
+                  UIBench after the given duration.
+                  """),
+    ]
+
+    def run(self, context):
+        super(Uibench, self).run(context)
+        self.target.sleep(self.duration)
--- a/wa/framework/workload.py
+++ b/wa/framework/workload.py
@@ -257,6 +257,12 @@ class ApkWorkload(Workload):
                 raise ConfigError('Target does not appear to support Android')
 
         super(ApkWorkload, self).__init__(target, **kwargs)
+
+        if self.activity is not None and '.' not in self.activity:
+            # If we're receiving just the activity name, it's taken relative to
+            # the package namespace:
+            self.activity = '.' + self.activity
+
         self.apk = PackageHandler(self,
                                   package_name=self.package_name,
                                   variant=self.variant,


wa/workloads/uibench/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+            'default value of this parameter will result in UIBench opening '
+            'in its workload selection menu.'),
+        Parameter('duration', kind=int, default=10, description='As activities '
+            'do not finish, this workload will terminate UIBench after the '
+            'given duration.'),
+    ]
+
+    def setup(self, context):
+        if self.activity is not None:
+            if self.activity not in self.apk.apk_info.activities:
```
pietos01:
```suggestion
            if self.activity not in self.apk.activities:
```
And, in `PackageHandler`:
```python
@property
def activities(self):
    if self.apk_info is None:
        return None
    return self.apk_info.activities
```

wa/workloads/uibench/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+                msg = 'Unknown activity "{}". Known activities:\n\t{}'
+                activities = '\n\t'.join(self.apk.apk_info.activities)
```
pietos01:
```suggestion
                activities = '\n\t'.join(self.apk.activities)
```

wa/workloads/uibench/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+                raise ValueError(msg.format(self.activity, activities))
+            self.apk.apk_info.activity = '.{}'.format(self.activity)
```
pietos01:
```suggestion
            self.apk.activity = '.{}'.format(self.activity)
```
And, in `PackageHandler`:
```python
activity = property(lambda self: return None if self.apk_info is None else self.apk_info.activity,
                    lambda self, activity: self.apk_info.activity = activity)
```

wa/workloads/uibench/__init__.py
pietos01: This whole value check could also be added into `PackageHandler.activity.setter` which would `raise` in this exact situation, making the calling code simpler and the called code more robust.

wa/workloads/uibench/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+        super(Uibench, self).setup(context)
+        time.sleep(self.duration)
```
jimbosjetset: uibench might not be the most familiar workload to developers.  Could we be more helpful by adding a short description of the workload and provide a url (https://android.googlesource.com/platform/frameworks/base.git/+/master/tests/UiBench/) for reference?

wa/workloads/uibench/__init__.py
pietos01: Ignored.

pietos01:
Instead of having an approach where we manually write to an instance variable (`.activity`) before calling a method, I made `activity` a parameter of the method(s) concerned. IMO, it's cleaner, more robust and seems to fit pretty coherently with the purpose of the method(s).

pietos01:
Will need to rebase for the `fixup!` commits and to re-order them (framework changes first, introduction of `Uibench` next).

setrofim:
The methods are methods for a generic  "workload" (or more generally job); Their parameters should not tie them to anything specific such as and Android activity. Please remove this chage.

pietos01:
I have changed:

- `ApkWorkload.setup()`
- `ApkUIWorkload.setup()`
- `ApkUIautoWorkload.setup()`
- `PackageHandler.setup()`
- `PackageHandler.start_activity()`
- `PackageHandler.restart_activity()`

All of these methods deal **exclusively** with Android targets and their APKs. All APKs have [_activities_](https://developer.android.com/guide/components/activities):

> Activities are one of the fundamental building blocks of apps on the Android platform. They serve as the entry point for a user's interaction with an app, and are also central to how a user navigates within an app or between apps.

Under these conditions, adding a feature dealing with an inherent part of any Android application to a set of classes dealing with Android applications seems reasonable; how is this change affecting the genericity of these classes or of their methods?

Please note that this PR does not modify `Workload` and does not compromise the polymorphism of the aforementioned classes when having to be used as generic `Workload`s, thanks to the use of default values for the `activity` parameters.

setrofim:
The `*Workload.setup()` is a workload method and and should implement that signature. You're modifying it part way through the inheritance hierarchy to add another parameter; more over you remove it further down the hierarchy, so you now have
```
Workload.setup(self, context)
ApkWorkload.setup(self, context, activity=None)
UiBench.setup(self. context)
```
To me, this neither cleaner, nor more robust. And since the parameter that is being passed is an attribute of the object, and so is already available inside the method it is being passed to, I really don't see what problem this change is solving that justifies the inheritance hierarchy weirdness.

For `PackageHandler` the activity is part of its state, and all its methods operate on that one same activity. It's conceptually inconsistent to have some of them operate on one activity, and others on another, passed as a parameter. So again, I don't see how this is cleaner. In principle, I have no problem with moving the activity out of the `PackageHandler` state and having it consistently passed as a parameter (with no assumption being made inside `PackageHandler` that different methods are operating on the same activity), if that would result in cleaner overall design; however this not the change you're making.

pietos01:
You are right about the `ApkWorkload.setup`-`UiBench.setup` inheritance; that's not clean and having `UiBench.setup(self, context, activity=None)` wouldn't be cleaner as the `activity` is passed through the agenda.

The "cleaner and more robust" comment referred to "_calling a method with an argument_" vs "_calling a method without argument after having overwritten an instance variable controlling the behaviour of that method_" which is what I was doing previously in this PR. I keep this point of view, however, it's now obvious that the latest change concerns more than just that.

In particular, the reason why I had to modify the `setup` methods is because the application is actually `am start`ed in that method (because `ApkWorkload.setup` calls `PackageHandler.setup` which calls `start_activity`). If, instead,  `ApkWorkload.run` (currently a `pass`) was starting the application (`self.apk.start_activity()`), I could simply overload that method so that it passes the `activity` to the modified version of `PackageHandler` you describe (no `self.activity` is part of its internal state). What do you think of this approach? Why is the `PackageHandler` starting the application during `setup` (insn't there a mismatch between the **meaning** of a workload `setup` and `PackageHandler.setup`)?


> For `PackageHandler` the activity is part of its state, and all its methods operate on that one same activity. It's conceptually inconsistent to have some of them operate on one activity, and others on another, passed as a parameter.

Except from (`re`)`start_activity`, which methods of `PackageHandler` are you referring to?

setrofim:
>The "cleaner and more robust" comment referred to "calling a method with an argument" vs "calling a method without argument after having overwritten an instance variable controlling the behaviour of that method" which is what I was doing previously in this PR. 

You were modifying the state of an object form inside a method of that object; said state was then being used by another method of that object. Conceptually, there is nothing wrong with that -- that's what objects with methods are for (vs just having functions where the entire state is passed as arguments). In practice I agree that the old solution was not the most elegant. However, I still prefer it over this one -- at least, in that case, the cruftiness was contained within one workload.

The issue here is that our model for how we run apk workloads does not fit particularly well how this workload needs to be run. If this is a one-off, I am happy to live with  a bit of cruft, as long as it is contained with this workload. If this is a wider issue with the model, then perhaps it needs to be reexamined and wider, more systematic, changes made. However I'd rather avoid making quick changes to the framework that weaken its conceptual integrity for the sake of making implementation of one workload a little cleaner.

>In particular, the reason why I had to modify the setup methods is because the application is actually am started in that method (because ApkWorkload.setup calls PackageHandler.setup which calls start_activity). If, instead, ApkWorkload.run (currently a pass) was starting the application (self.apk.start_activity()), I could simply overload that method so that it passes the activity to the modified version of PackageHandler you describe (no self.activity is part of its internal state). What do you think of this approach? Why is the PackageHandler starting the application during setup (insn't there a mismatch between the meaning of a workload setup and PackageHandler.setup)?

I'm not sure I'm following the reasoning here. There is no mismatch in the meaning of `setup` for `Workload` and `PackageHandler`. `run` is the method that executes the part of the workload that we actually want to measure. The job of the `setup` method of `Workload` is to get the workload to that point. `start_activity` launches the application. For the vast majority of Android applications, we do not want to start measuring yet as the actual "workload" is not running yet  -- there may be some splash screen to click through, some menus to navigate, some GUI settings to twiddle, etc. This is all done as part of the setup up to that last "click" that actually runs the "interesting" portion of the application. Hence the app is `am start`'d  and the initial menus / gui settings done as part of `setup`, and `run` contains the actual execution of the app. For the few workloads that do run immediately on launch, we typically have not cared if we missed the first few seconds of execution.

pietos01:
> The issue here is that our model for how we run apk workloads does not fit particularly well how this workload needs to be run. If this is a one-off, I am happy to live with a bit of cruft, as long as it is contained with this workload.

As stated in the quote from the official Android documentation I post above, _activities_ are an integral and important part of Android applications. For this reason, `ApkWorkload` (and, by extension, any class concerned with starting an Android application such as `PackageHandler`) should support them, IMO.

The ways I was trying to patch this support into the code were arguably hacky, I agree with that. However, this is not a one-off to satisfy the exotic requirements of a workload! To illustrate this, this PR is also not the first one facing this limitation of the framework _e.g._ #720 had the same requirements.

> For the vast majority of Android applications, we do not want to start measuring yet as the actual "workload" is not running yet -- there may be some splash screen to click through, some menus to navigate, some GUI settings to twiddle, etc.

This is **exactly** why this workload for UIBench uses activities: if you try launching the app (which effectively means starting in the default activity), you will be presented with a menu of all the "tests" the suite provides. However, the application provides an activity (entrypoint) for each one of these tests! Isn't it cleaner to ask the app to start a given test directly, by passing the name of the corresponding activity to `am start` than to start the app (opening the menu), launch an automation (after having had to implement it) that finds the right entry in the menu and simulate a screen input? By using activities as they were intended to be used, we completely get rid of the need to automate the menu of UIBench!

> Basically, the activity these `PackageHandler` methods operate on should either always be passed as a (mandatory) argument, or should always be obtained from the object's state.

I had considered passing the `activity` to the constructor but then, we're unable to validate the passed `activity` by comparing it with the list of available activities extracted from the APK, `.apk_info.activities` as `apk_info` is not available until `initialize`. But now that I think of it, it's probably cleaner to do this and delay the check until then or even until right before the call to `am start`(?)

> I'm not sure I'm following the reasoning here.

I had misinterpreted "`run`" as "_run_ this [command/application]" (as in [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run)) as opposed to "this is the _run_ we capture". Indeed, in most cases, it makes sense to have the applauch take place during the `setup` phase. The point I was trying to make is what I've explained above: I believe that the framework should support activities and I've struggled to neatly integrate them for the reasons mentioned above.

setrofim:
>To illustrate this, this PR is also not the first one facing this limitation of the framework e.g. #720 had the same requirements.

The issue highlighted that PR has already been addressed (pretty much in the way that the PR was solving it) by https://github.com/ARM-software/workload-automation/commit/753786a45c6984d41eb8cd10f526071baf148f18.

>By using activities as they were intended to be used, we completely get rid of the need to automate the menu of UIBench!

Yes, I agree it would be better to use the activity directly, and that's what we've done in the past for  the workloads that supported this. But so far, we could get way with launching them from setup. The issue, as I understand it, is not specifying which activity to launch (which is already supported); but having it launched as part of `run` rather than `setup`.  Or am I misunderstanding the problem?

>I had misinterpreted "run" as "run this [command/application]" (as in subprocess.run) as opposed to "this is the run we capture".

FWIW, we do document the intended functionality of each method here: https://workload-automation.readthedocs.io/en/latest/api/workload.html

pietos01:
Rebased as required to use the `ApkWorkload.activity` field which was merged after opening this PR.

wa/workloads/uibench/__init__.py
pietos01: What about the current version?

pietos01:
> Would it be worth adding in validation of the activity using the functionality that was added in [ARM-software/devlib#357](https://github.com/ARM-software/devlib/pull/357)?

Yes and initially (before I rebased it), it was done in this PR.

The reason why I haven't added it back is that I believe this test should be added in the `PackageHandler` right before the call to `am start`. However, because of how ARM-software/devlib#357 was implemented (partially due to a lack of knowledge about APKs and Android apps on my side), I _think_ it might only be capturing activities that are in the namespace of the main application (notice the [`.` in the activity regex](https://github.com/ARM-uxperf/devlib/blob/7bb4009c3896753710f0cf41b40e35169666f933/devlib/utils/android.py#L135)) which doesn't necessarily include all activities.

Therefore, a test such as `if activity in self.apk_info.activities` would break for workloads that provide activities in other namespaces (or even absolute names) and I believe I have seen such workloads in WA.

My idea was to investigate that and open another PR to fix the activity regex so that we can add the check when running a given activity. This check isn't required for the functionality that this PR attempts to introduce, though. In particular as it's not present in the current `master` so that some workloads are running "untested" activities.

IMO, the main advantage of that check (and the reason why I had implemented it in the first place) is that the error message could then state which activities are accepted (as opposed to the current error message coming back from the target, simply stating that the activity is invalid).

marcbonnici:
Ok that seems fair enough, in that case happy to merge this as is until the best way to perform the validation has been decided.

PR #945, opened 2019-01-10
wa: Bump to Development versions

marcbonnici:


wa/framework/version.py
```diff
@@ -21,7 +21,7 @@
 
 VersionTuple = namedtuple('Version', ['major', 'minor', 'revision'])
 
-version = VersionTuple(3, 1, 1)
+version = VersionTuple(3, 1, 'dev2')
```
setrofim: Shouldn't this be `VersionTuple(3, 1, 'dev2')`?

marcbonnici: Good point, updated to add a development field to the version.

PR #951, opened 2019-01-30
fw/target/info: Fix for KernelConfig refactor

marcbonnici:
The Devlib KernelConfig object was refactored in commit
https://github.com/ARM-software/devlib/commit/f65130b7c7ecccfe2006e40d2735eeb86639772b
therefore update the way KernelConfig objects are deserialized to reflect the new
implementation and provide a conversion for PODs.

### DIFF
--- a/wa/framework/target/info.py
+++ b/wa/framework/target/info.py
@@ -53,9 +53,9 @@ def kernel_version_from_pod(pod):
 
 def kernel_config_from_pod(pod):
     config = KernelConfig('')
-    config._config = pod['kernel_config']
+    config.typed_config._config = pod['kernel_config']
     lines = []
-    for key, value in config._config.items():
+    for key, value in config.items():
         if value == 'n':
             lines.append('# {} is not set'.format(key))
         else:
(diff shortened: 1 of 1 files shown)


wa/framework/target/info.py
```diff
@@ -53,9 +53,9 @@ def kernel_version_from_pod(pod):
 
 def kernel_config_from_pod(pod):
     config = KernelConfig('')
-    config._config = pod['kernel_config']
+    config.typed_config._config = pod['kernel_config']
     lines = []
-    for key, value in config._config.items():
+    for key, value in config.iteritems():
```
setrofim: Should be `items()` to be compatible with Python 3

marcbonnici: This is actual a method of a `KernelConfg` object rather than a python built in.
**https://github.com/ARM-software/devlib/blob/master/devlib/target.py#L1998**

Would it make sense to add an alias to `KernelConfig` to avoid confusion?

setrofim: Ah, my mistake. Yes, and alias might be a good idea.

marcbonnici: Updated to use aliased method.

PR #956, opened 2019-02-20
docs/installation: Add warning about using pip to install from github

marcbonnici:


doc/source/user_information/installation.rst
```diff
@@ -199,6 +199,11 @@ Alternatively, you can also install the latest development version from GitHub
        cd workload-automation
        sudo -H python setup.py install
 
+.. note:: Please note that if using pip to install from github this will most
+          likely result in an older and incompatible version of devlib being
+          installed alongside WA. If you wish to use pip please also manually
+          install the latest version of
+          `devlib <https://github.com/ARM-software/devlib>`_.
```
Sticklyman1936: Does one need to do this before installing WA? After? Does it not matter?

marcbonnici: I don't think it should matter. If it's done before pip should detect the requirement is already satisfied and move on and if it is done after it should just install over the top.

Sticklyman1936: OK, thought as much but wanted to make sure we didn't need to clarify. Thanks!

PR #957, opened 2019-02-20
[RFC] Devlib Version Improvements + Version Bump

marcbonnici:
Dependant on https://github.com/ARM-software/devlib/pull/372


marcbonnici:
Added commits to do version bumps to coincide with new versioning scheme.

marcbonnici:
Updated to limit PyYAML to use latest release version.

PR #962, opened 2019-03-15
workloads: add aitutu

scojac01:
Add a workload to execute the Aitutu benchmark.


wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
```diff
@@ -0,0 +1,147 @@
...
+        if (perm3.exists()) {
+            allow.click();
+        }
+        if (perm4.exists()) {
+            allow.click();
+        }
+
+        UiObject next = mDevice.findObject(selector.textContains("NEXT")
+                                             .className("android.widget.Button"));
+        waitObject(next);
```
marcbonnici: On my nexus5 I do not get any of these permission popups,  so the workload will wait here for 10 minutes before crashing. 
The `waitObject` calls should have a small timeout value and only attempt to select the object if they are present. 

Although are the permissions for the app not being automatically be granted during installation?

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
```diff
@@ -0,0 +1,147 @@
...
+        UiObject cancel = mDevice.findObject(selector.textContains("CANCEL")
+                                             .className("android.widget.Button"));
+        waitObject(cancel);
+        cancel.click();
+    }
+
+    public void runBenchmark() throws Exception {
+        UiSelector selector = new UiSelector();
+        //Start the tests
+        UiObject start = mDevice.findObject(selector.textContains("Start Testing")
```
marcbonnici: By default the target cache is cleared resulting in the additional download to have to be repeated for each run. 
Is this necessary? If not `self.clear_data_on_reset` could be set to `False` for the python workload to prevent this, although in that case it should be noted in the workload description that the first run will involve downloading additional assets and therefore will significantly differ to additional runs.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
```diff
@@ -0,0 +1,147 @@
...
+                                                     .className("android.widget.TextView"));
+        waitObject(start);
+        start.click();
+
+        //Wait for the tests to complete
+        UiObject complete =
+            mDevice.findObject(selector.text("TEST AGAIN")
+                .className("android.widget.Button"));
+        complete.waitForExists(1200000);
```
marcbonnici: The default timeout for the python workload is 10 mins and times out before this is reached. Also set `self.timeout` to match this value.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
```diff
@@ -0,0 +1,147 @@
...
+
+    }
+
+    public void getScores() throws Exception {
```
marcbonnici: Would it make sense to also extract which AIModel was used for the tests for better comparisons?

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
scojac01: It is possible that in future releases the dependencies that the app has to download will change. I'd like to keep this as part of the workload for future proofing.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
scojac01: Do you mean extracting the text element? Image Classification, Object Detection etc? 

If so I don't see the benefit, the scores are pulled to an element named either Image or Object.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
scojac01: I've removed this section as the permissions we're not visible when I ran through either. I added them when first creating the workload and running through manually.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
scojac01: I've changed the wait here to 600000 to match. This still gives enough time.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
marcbonnici: Would the application itself not detect this and re download the relevant assets? 
I was just thinking that if you're measuring anything aside from the workload reported scores, e.g. device power, time etc these would be very much dependent on the speed of the Internet connection at the particular time the workload was ran and would be comparable across iterations?

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
marcbonnici: The shorter time was not enough on my device (an old nexus5), therefore would it make sense to allow the timeout to be a user configurable parameter to the workload?

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
marcbonnici: No sorry, I meant more "TFLite" vs "SNPE" etc.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
scojac01: Can WA manage at what point in the workload it records instrumentation? If so that would solve this problem nicely as the code for the actual testing in the run stage would be all we'd want to record against? 

I'm not sure of the behaviour when assets are missing as it hasn't happened yet. I just wanted to future proof and keep things as generic as possible.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
scojac01: I'm not sure where these are specified in the application?

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
marcbonnici: The idea WA uses is that the `run` section of the workload is only meant to be the work that should be measured and all the downloading / configuration is meant to be performed within the `setup` section, therefore WA will kick off and stop the instrumentation as close as it can before/after that method is executed so unfortunately it does not support configurable starting positions within execution sections.  

Is there any way to split the two sections or wait for the download to complete before ending the `setup` phase and then starting the `run` stage?

scojac01: I could add a wait for the "downloading" text to disappear and then start the run stage, the app would still be performing actions though unless there is a way to force pause it?

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
marcbonnici: On the results screen next to the score for each section it also lists the model that was used during the benchmark, for example here it has used the "TFLite" model. 
aitutu

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
marcbonnici: Hmm... I'm not aware of anyway to forcibly pause an application so waiting for the download to complete might be the better option.. Having a look at the application it does seem to have a intermediate stage checking the resources and atleast the delay / offset should be consistent between different runs?

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
scojac01: Reverted to 1200000 and updated the self.timeout value to match.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
scojac01: Completely missed that. Have now included.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
```diff
@@ -0,0 +1,144 @@
...
+        //Start the tests
+        UiObject start = mDevice.findObject(selector.textContains("Start Testing")
+                                                     .className("android.widget.TextView"));
+        waitObject(start);
+        start.click();
+
+        UiObject check = mDevice.findObject(selector.textContains("Checking AI resources")
+                                                     .className("android.widget.TextView"));
+        waitObject(check);
+        check.waitUntilGone(300000);
```
marcbonnici: On some devices this stage seems to run too quick to be detected so this never fires. 
Could I suggest changing the waiting until "checking AI resources" has gone to wait for the text "classification" which shown on the testing screen instead to make it more reliable?

scojac01:
Hi Marc, think I've handled most of these changes. Can I get another review please?

marcbonnici:
Hi Scott, I don't seem to see able to see any changes to the PR here. Could you try re-pushing your changes?

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
```diff
@@ -0,0 +1,140 @@
...
+        getScores();
+    }
+
+    public void clearPopups() throws Exception {
+        UiSelector selector = new UiSelector();
+
+        UiObject cancel = mDevice.findObject(selector.textContains("CANCEL")
+                                             .className("android.widget.Button"));
+        waitObject(cancel);
+        cancel.click();
```
marcbonnici: This should have a timeout and conditionally click on cancel if the popup is present.

wa/workloads/aitutu/uiauto/app/src/main/java/com/arm/wa/uiauto/aitutu/UiAutomation.java
```diff
@@ -0,0 +1,140 @@
...
+        //Start the tests
+        UiObject start = mDevice.findObject(selector.textContains("Start Testing")
+                                                     .className("android.widget.TextView"));
+        waitObject(start);
+        start.click();
+
+        UiObject check = mDevice.findObject(selector.textContains("classification")
+                                                     .className("android.widget.TextView"));
+        waitObject(check);
+        //check.waitUntilGone(300000);
```
marcbonnici: Can remove commented code

scojac01:
Have responded to changes. Please review.

PR #971, opened 2019-04-25
Add support for "perf record" and "perf report" to WA

markbellamy:


wa/instruments/perf.py
```diff
@@ -77,6 +77,17 @@ class PerfInstrument(Instrument):
                   description="""Provides labels for pref output. If specified, the number of
                   labels must match the number of ``optionstring``\ s.
                   """),
+        Parameter('perf_mode', kind=str, default='stat', allowed_values=['stat', 'record'],
```
setrofim: Parameter can be renamed to just `"mode"`, given that it is for `perf` instrument, calling it `perf_mode` is redundant.

wa/instruments/perf.py
```diff
@@ -77,6 +77,17 @@ class PerfInstrument(Instrument):
...
+                  global_alias='perf_mode',
+                  description="""
+                  Choose between 'perf stat' and 'perf record'. If 'perf record' is selected
+                  'perf report' will also be run. 'report_optionstring' can be used to generate
+                  custom report from the trace.
+                  """),
+        Parameter('report_optionstring', kind=list_or_string, default='',
+                  global_alias='report_perf_options',
+                  description="""Specifies options to be used for the 'perf report' command used
```
setrofim: Please format consistently with the other parameters.

wa/instruments/perf.py
```diff
@@ -134,5 +146,38 @@ def update_output(self, context):
...
+                            if words[4] == 'event':
+                                event_type = words[5]
+                                event_type = event_type.strip("'")
+                        if words[0] != '#' and '%' in words[0] and len(words) == 5:
+                            metric = 'perf/{}/{}/{}'.format(event_type, words[1], words[2].strip("[]"))
+                            temp = words[0]
+                            count = temp.strip('%')
+                            context.add_metric(metric, count, '%')
+
+    def update_output(self, context):
```
setrofim: Should go above internal methods.

wa/instruments/perf.py
```diff
@@ -134,5 +146,38 @@ def update_output(self, context):
...
+        self.logger.info('Extracting reports from target...')
```
setrofim: Given the logging inside the `_stat` and `_output` methods, this is redundant. Please keep info-level logging to a minimum. Should be converted to `debug`.

wa/instruments/perf.py
```diff
@@ -134,5 +146,38 @@ def update_output(self, context):
                                 metric = '{}_{}'.format(label, match.group(3))
                                 context.add_metric(metric, count, classifiers=classifiers)
 
+    def _update_output_record(self, context):
+        outdir = os.path.join(context.output_directory, 'perf')
+        self.collector.get_trace(outdir)
```
setrofim: Given that the above two lines happen in both `_stat` and `record` cases, they should be moved out into `update_output`.

wa/instruments/perf.py
```diff
@@ -134,5 +146,38 @@ def update_output(self, context):
...
+
+        self.logger.info('Processing perf report reports.')
+        for host_file in os.listdir(outdir):
+            label = host_file.split('.rpt')[0]
```
setrofim: use `os.path.splitext` instead.

wa/instruments/perf.py
```diff
@@ -134,5 +146,38 @@ def update_output(self, context):
...
+            with open(host_file_path) as fh:
+                for line in fh:
+                    words = line.split()
+                    if len(words) >= 1:
+                        if words[0] == '#' and len(words) == 6:
+                            if words[4] == 'event':
+                                event_type = words[5]
+                                event_type = event_type.strip("'")
+                        if words[0] != '#' and '%' in words[0] and len(words) == 5:
+                            metric = 'perf/{}/{}/{}'.format(event_type, words[1], words[2].strip("[]"))
```
setrofim: Consider using classifiers rather than packing everything into the metric name.

wa/instruments/perf.py
```diff
@@ -134,5 +146,38 @@ def update_output(self, context):
...
+                            temp = words[0]
+                            count = temp.strip('%')
```
setrofim: `temp` not necessary here. Also, a metric value should be a number, not string.

```
count = numeric(words[0].strip('%'))
```
(note: `numeric` must be imported from `wa.utils.types`.

wa/instruments/perf.py
markbellamy: Thanks, changed in latest version.

wa/instruments/perf.py
markbellamy: Thanks, changed in latest version.

wa/instruments/perf.py
markbellamy: Thanks, changed in latest version.

wa/instruments/perf.py
markbellamy: Thanks, changed in latest version.

wa/instruments/perf.py
markbellamy: Thanks, changed in latest version.

wa/instruments/perf.py
markbellamy: Thanks, changed in latest version.

wa/instruments/perf.py
markbellamy: Thanks, changed in latest version. (I initially thought about doing this but expected it to be written out as a string again so didn't seem necessary.)

wa/instruments/perf.py
```diff
@@ -134,5 +146,38 @@ def update_output(self, context):
                                 metric = '{}_{}'.format(label, match.group(3))
                                 context.add_metric(metric, count, classifiers=classifiers)
 
+    def _update_output_record(self, context):
+        outdir = os.path.join(context.output_directory, 'perf')
+        self.collector.get_trace(outdir)
+
+        self.logger.info('Processing perf report reports.')
+        for host_file in os.listdir(outdir):
```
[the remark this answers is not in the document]
markbellamy: Thanks, changed in latest version.

wa/instruments/perf.py
markbellamy: I've added the classifiers but also left it packed in the metric name. It seems that some of the output processors, e.g. csv, default to classifier output off and that will result to a lot of meaningless/duplicate entries in the results file by default options. Which doesn't seem likely friendly behavior. 

I am very open to peoples thoughts on what is the right behavior here thou.

markbellamy: I had a thought today that change the metric to <event>/<app>  which should be unique key and most critical info. Then use classifier for the other entries.

markbellamy:
>  As an end user, I would personally prefer the perf instrument to be as generic and flexible as possible (in terms of running perf commands, not only stat or record nor having to manually patch/inject redundant code for each subcommand).

I agree, I was surprised that perf support was limited to just `perf stat` and I think a much more generic implementation is the right direction. This implementation is just a "quick" one to add 'perf record' support to the current implementation. as I wanted to use it. It seem the right thing to see about committing it back as might be useful to others as stop gap until wider perf support is added.

wa/instruments/perf.py
```diff
@@ -134,5 +146,38 @@ def update_output(self, context):
...
+            label = host_file.split('.rpt')[0]
+            host_file_path = os.path.join(outdir, host_file)
+            context.add_artifact(label, host_file_path, 'raw')
+            with open(host_file_path) as fh:
```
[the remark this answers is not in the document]
markbellamy: This is just a first implementation to have it parse the basic output. As you say handling all the output options is difficult. If I update this to handle new report formats I'll likely see about committing them but I am not planning to write a more generic/complete one. If you have one that could be integrate here that would be very good.

PR #974, opened 2019-05-13
utils/serializer: fix error reporting for YAML 

setrofim:
When attempting to access the message of a exception check not only that
e.args is populated, but also that e.args[0] actually contains
something, before defaulting to str(e).


marcbonnici:
This change appears to have already been merged. Please re open if this is not the case.

PR #976, opened 2019-05-14
utils/serializer: fix support for YAML anchors.

setrofim:
Change the way maps get processed by YAML constructor to support YAML
features, such as anchors, while still maintaining dict ordering.


setrofim:
Somehow I and/or git got confused and botched this up. Opening https://github.com/ARM-software/workload-automation/pull/976 instead. _sigh_ I need coffee...

PR #979, opened 2019-05-15
Add generic perf Instrument

pietos01:
### Introduction

As discussed in #971, this PR tries to modify the `perf` instrument so that it accepts any `perf` command. It is related with ARM-software/devlib#388 which implements its back-end. It also requires ARM-software/devlib#387 to properly run some `perf` subcommands (ARM-software/devlib#386).

The idea is to accept a YAML input (part of the agenda) that is in the format of what `perf` expects:
```
$ perf help

 usage: perf [--version] [--help] [OPTIONS] COMMAND [ARGS]
```
Which then only requires formatting this command based on the input and doesn't require having logic in the WA instrument to handle every `COMMAND`.

### Single Command

An excerpt from an agenda describing a command (here, `perf record`) would look like:
```YAML
                command: record
                flags:
                    - all-cpus
                    - no-inherit
                kwflags:
                    freq: 1000
                    event:
                        - r8
                        - cs
                    cpu:
                        - 0
                        - 4
                args:
                    command: sleep
                    args: 1000
                stdout: '/dev/null'
                stderr: '&1'
```
the accepted keys (`command`, `flags`, `kwflags`, ...) being the names of the parameters to `devlib.utils.cli.Command.__init__` (see the back-end implementation). This allows the highest level of flexibility, will support any version of `perf` and any `perf` subcommand, is relatively simple to implement and maintain and relatively robust (see the following discussion about fail cases). Having such a level of control over the parts of the `perf` command allows us to take full advantage of YAML anchors. The readability (as seen from the agenda) of this approach partially comes from the expressiveness of the `perf` flags. As an example, notice that `event`s are passed as clearly as in the previous implementation, yet no specific logic had to be implemented for it in the instrument.

### Full Run

On top of this, the proposed `perf` instrument takes 3 dictionaries of commands (`pre_commands`, `commands`, `post_commands`) based on when these commands have to be run: instead of hardcoding this information in the instrument for each subcommand (cfr. #971), we let the user input it, which is a small amount of low-risk work for them. The keys of the command dictionaries are their labels (as introduced by the previous implementation, see further discussion in the back-end PR).

For example, a full agenda:
```YAML
config:
    augmentations:
    - perf
    iterations: 1
    perf:
        force_install: true
        commands:
            first:
                command: record
                flags:
                    - all-cpus
                    - no-inherit
                kwflags:
                    freq: 1000
                    event:
                        - r8
                        - cs
                    cpu:
                        - 0
                        - 4
                args:
                    command: sleep
                    args: 1000
                stdout: '/dev/null'
                stderr: '&1'
        post_commands:
            first:
                command: report
                kwflags:
                    sort: dso
                    field-separator: ;
                stdout: report.stdout

workloads:
-   name: dhrystone
    params:
        cleanup_assets: true
        cpus: 0
        delay: 0
        duration: 0
        mloops: 0
        threads: 4
```
which will run the following on the device (notice the properly escaped `--field-separator`):
```
# perf record --all-cpus --no-inherit --freq=1000 --event=r8,cs --cpu=0,4 -- sleep 1000 1>/dev/null 2>&1
# perf report --sort=dso '--field-separator=;' -- 1>report.stdout
```
according to the following rules:

- `flags` are prepended with `-` or `--` based on their length;
- `kwflags` are CSV lists linked with `=` (valid for all `perf` commands; seems to be a (sometimes implicit) standard for flags taking values. This is probably because they all use the same front-end);
- `args` is recursively parsed as a command, allowing to use `perf` in its original usage (_i.e._ by launching a command through it and only capturing that command);
- `stdout` and `stderr` are used as expected and handle the UNIX `&`-based pipe redirection (see discussion about file names, labels, and which files which commands can see); extra flags could be considered such as `stdin` or pipes but I'm wondering how generic we should go, here ...

**NB**: The user (writing the agenda) needs to have access to the file names so that they can use the `-o`/`-i` flags to properly pipe their `record`/`report`. 

The `force_install` flag is kept from the previous implementation. However, what is the "standard WA way" of providing the `perf` binary? We could write the path in the agenda (_e.g._ for comparing versions of `perf`), we could have a collection of `perf` binaries in a standard location and pick the ideal one (_e.g._ based on kernel version of the target) from within the WA `perf` instrument, we could (but this requires automating the automation, WA, and feels messy) have WA get the binary from a standard location on the host and overwrite that one between runs... Ideally, it would be useful to have both control of which binaries are being used and automation for finding the optimal version (as I believe the `perf`-kernel interface evolves with the kernel).

Parsing of the output into WA metrics hasn't been started yet. I propose to have, in this case, a per-command logic (I don't think there is another way) for the `perf` subcommands that are supported (probably `stat` and `report`). Because of how flexible and variable the output of `perf report` can be (typically a table or a graph), what would be the expected output in terms of WA metrics? I currently have a parser that generates a `pandas.DataFrame` from the tabular output of `perf report` but this seems to be "too different" from the JSON-based outputs WA instrument seem to be using. Feedback required!

#### Full Run (with the power of YAML)

Another agenda, similar to the example mentioned in the docstring of the previous implementation, runs the same `perf stat` on the big and LITTLE clusters (using YAML anchors; #976):
```YAML
    perf:
        commands:
            little: &little
                command: stat
                flags:
                    - all-cpus
                    - no-inherit
                kwflags: &little_kwflags
                    freq: 1000
                    event:
                        - r8
                        - cs
                    cpu: 0-3
                args:
                    command: sleep
                    args: 1000
                stdout: 'little.stat'
                stderr: '&1'
            big:
                <<: *little
                kwflags:
                    <<: *little_kwflags
                    cpu: 4-7
                stdout: 'big.data'
```

### Drawbacks

Obviously, the flexibility of this instrument comes at a cost as this approach won't necessarily fail early:
- Errors at the first "YAML dictionary level" (`post_commands`, `commands`, `pre_commands`, `force_install`) fail early as usual (through using the `Parameter` class);
- Errors at the second "YAML dictionary level" (`command`, `flags`, `kwflags`, ...) fail relatively early, at instantiation-time of `Command` (almost same behaviour as `Parameter`);
- Errors at the third "YAML dictionary level" (values of `flags`, keys of `kwflags`, format of the final command string) fail late, only once the corresponding command is run on the device.

Secondly, because `perf` is a tool intended to run another command, this instrument is implemented while making the assumption that the user of the agenda file is allowed by the owner of the target to run arbitrary commands on it. In the case they are not, additional security should be implemented. For example, it is possible to write an agenda running the following destructive performance-investigating command:
```
perf record -- rm -rf /
```

### TODO:

- [x] ~~Write a "porting guide" from old-`PerfInstrument` to `PerfStatInstrument`~~
- [x] ~~Deprecate `PerfStatInstrument`~~
- [x] Document the instrument
- [x] Decide how labels are used on the device and the host, how output files are named (and how much control the user has over it) and if there are per-label subdirectories;
- [x] Implement parsers for `stat`
- [x] ~~Implement parsers for `report`~~
- [x] Figure out why `Parameter(..., kind=YamlCommandDescriptor)` behaves weirdly and required somewhat hacking the constructor
- [ ] Unlock `teardown`: should we remove everything from the target?
- [ ] Change (or not) `force_install`

### DIFF
--- a/wa/instruments/perf.py
+++ b/wa/instruments/perf.py
@@ -75,6 +75,53 @@ class PerfInstrument(Instrument):
 
     Please refer to the ``sysfile_values`` runtime parameter to do so from an
     agenda.
+
+    When running ``perf stat``, this instrument reports the captured
+    counters as unitless :class:`Metrics` with the following classifiers:
+
+    - ``'name'``: The name of the event as reported by ``perf``. This name
+      may not be unique when aggregation is disabled as the same counter is
+      then captured for multiple hardware threads;
+    - ``'label'``: Label given to the run of ``perf stat``;
+    - ``'target'``: The target ``perf`` reports for the captured events.
+      This is shared across all events of a run and is further specialized
+      by ``'hw_thread'``, ``'core'`` and ``'cluster'`` if applicable;
+    - ``'duration'``, ``'duration_units'``: duration of the ``perf`` run;
+    - ``'count_error'``: A string containing the error corresponding that
+      prevented the counter from being captured. Only available if an error
+      occured. In this case the value of the metric is always ``0``;
+    - ``'hw_thread_count'``: Number of **hardware** threads that were
+      contributing to the counter. Only available when the automatic
+      aggregation done by ``perf stat`` is disabled. See ``'hw_thread'``,
+      ``'core'`` and ``'cluster'``;
+    - ``'hw_thread'``: When the ``--no-aggr`` option is used, holds the
+      index of the hardware thread that incremented the counter. In this
+      case, ``'hw_thread_count'`` is always ``1``. For backward
+      compatibility, the ``'cpu'`` classifier is provided as a synonym of
+      ``'hw_thread'`` (unlike what its name might suggest, on systems
+      supporting hardware multithreading, ``'cpu'`` is not a synonym of
+      ``'core'``!);
+    - ``'cluster'``: When the ``--per-socket`` option is used, holds the
+      index of the cluster (_i.e._ "socket" in ``perf`` terminology) that
+      incremented the counter and ``'hw_thread_count'`` holds the number of
+      hardware threads in the cluster. When the ``--per-core`` option is
+      used, this classifier gives the index of the cluster of the core.
+    - ``'core'``: When the ``--per-core`` option is used, holds the index
+      (within its cluster) of the core that incremented the counter and
+      ``'hw_thread_count'`` holds the number of hardware threads in the
+      core.
+    - ``'enabled'``: When ``perf`` needs to capture more hardware events
+      than there are hardware counters, it shares the hardware counters
+      among the events through time-slicing. This classifier holds the
+      fraction (between ``0.0`` and ``100.0``) of the run that a hardware
+      counter was allocated to the the event. Available only for hardware
+      events and only when time-slicing was required.
+    - ``'comment_value'``, ``'comment_units'``: Some counters may come with
+      an extra "comment" (following a ``#``) added by ``perf``. The
+      ``'comment_value'`` holds the numeric (``int`` or ``float``) value of
+      the comment while ``'comment_units'`` holds the rest of the comment
+      (typically the units). Only available for the events for which
+      ``perf`` added a comment.
     """
 
     parameters = [
@@ -259,54 +306,6 @@ class PerfInstrument(Instrument):
 
     @classmethod
     def _extract_stat_metrics(cls, label, stdout):
-        """
-        When running ``perf stat``, this instrument reports the captured
-        counters as unitless :class:`Metrics` with the following classifiers:
-
-        - ``'name'``: The name of the event as reported by ``perf``. This name
-          may not be unique when aggregation is disabled as the same counter is
-          then captured for multiple hardware threads;
-        - ``'label'``: Label given to the run of ``perf stat``;
-        - ``'target'``: The target ``perf`` reports for the captured events.
-          This is shared across all events of a run and is further specialized
-          by ``'hw_thread'``, ``'core'`` and ``'cluster'`` if applicable;
-        - ``'duration'``, ``'duration_units'``: duration of the ``perf`` run;
-        - ``'count_error'``: A string containing the error corresponding that
-          prevented the counter from being captured. Only available if an error
-          occured. In this case the value of the metric is always ``0``;
-        - ``'hw_thread_count'``: Number of **hardware** threads that were
-          contributing to the counter. Only available when the automatic
-          aggregation done by ``perf stat`` is disabled. See ``'hw_thread'``,
-          ``'core'`` and ``'cluster'``;
-        - ``'hw_thread'``: When the ``--no-aggr`` option is used, holds the
-          index of the hardware thread that incremented the counter. In this
-          case, ``'hw_thread_count'`` is always ``1``. For backward
-          compatibility, the ``'cpu'`` classifier is provided as a synonym of
-          ``'hw_thread'`` (unlike what its name might suggest, on systems
-          supporting hardware multithreading, ``'cpu'`` is not a synonym of
-          ``'core'``!);
-        - ``'cluster'``: When the ``--per-socket`` option is used, holds the
-          index of the cluster (_i.e._ "socket" in ``perf`` terminology) that
-          incremented the counter and ``'hw_thread_count'`` holds the number of
-          hardware threads in the cluster. When the ``--per-core`` option is
-          used, this classifier gives the index of the cluster of the core.
-        - ``'core'``: When the ``--per-core`` option is used, holds the index
-          (within its cluster) of the core that incremented the counter and
-          ``'hw_thread_count'`` holds the number of hardware threads in the
-          core.
-        - ``'enabled'``: When ``perf`` needs to capture more hardware events
-          than there are hardware counters, it shares the hardware counters
-          among the events through time-slicing. This classifier holds the
-          fraction (between ``0.0`` and ``100.0``) of the run that a hardware
-          counter was allocated to the the event. Available only for hardware
-          events and only when time-slicing was required.
-        - ``'comment_value'``, ``'comment_units'``: Some counters may come with
-          an extra "comment" (following a ``#``) added by ``perf``. The
-          ``'comment_value'`` holds the numeric (``int`` or ``float``) value of
-          the comment while ``'comment_units'`` holds the rest of the comment
-          (typically the units). Only available for the events for which
-          ``perf`` added a comment.
-        """
         match = cls._stat_regex.search(stdout)
         if match is None:
             return
--- a/wa/instruments/perf.py
+++ b/wa/instruments/perf.py
@@ -291,14 +291,35 @@ class PerfInstrument(Instrument):
                                        self.commands.items(),
                                        self.post_commands.items())
         for label, cmd in all_commands:
-            if 'stat' in cmd.command:
+            classifiers = {
+                'label': label,
+                'command': cmd.command,
+            }
+
+            outputs = []
+            if 'o' in cmd.kwflags:
+                outputs.append((cmd.kwflags['o'], 'data'))
+            if 'output' in cmd.kwflags:
+                outputs.append((cmd.kwflags['output'], 'data'))
+            if cmd.stdout and not cmd.stdout.startswith('&'):
+                outputs.append((cmd.stdout, 'data'))
+            if cmd.stderr and not cmd.stderr.startswith('&'):
+                outputs.append((cmd.stderr, 'log'))
+
+            metrics_done = False
+            for output, kind in outputs:
                 # perf stat supports redirecting its stdout to --output/-o:
-                stat_file = (cmd.kwflags.get('o', None) or
-                             cmd.kwflags.get('output', None) or
-                             cmd.stdout)
-                with open(os.path.join(outdir, label, stat_file)) as f:
-                    for metric in self._extract_stat_metrics(label, f.read()):
-                        context.add_metric(**metric)
+                output_path = os.path.join(outdir, label, output)
+                if 'stat' in cmd.command and not metrics_done:
+                    metrics_done = True
+                    with open(output_path) as f:
+                        for metric in self._extract_stat_metrics(label,
+                                                                 f.read()):
+                            context.add_metric(**metric)
+                    kind = 'raw'
+                context.add_artifact(os.path.join(label, output),
+                                     output_path, kind,
+                                     classifiers=classifiers)
 
     def teardown(self, context):
         # pylint: disable=unused-argument
--- a/wa/instruments/perf.py
+++ b/wa/instruments/perf.py
@@ -13,34 +13,18 @@
 # limitations under the License.
 #
 
-import collections
+import itertools
 import os
+import re
+import shlex
 
-from devlib.utils.cli import Command
-from devlib.trace.perf import PerfCollector
+from devlib.trace.perf import PerfCollector, PerfCommandDict
 from wa import Instrument, Parameter
 from wa.utils.types import list_or_string, list_of_strs
 
 __all__ = [
     'PerfInstrument',
 ]
-
-
-class YamlCommandDescriptor(collections.OrderedDict):
-
-    def __init__(self, yaml_dict):
-        super(YamlCommandDescriptor, self).__init__()
-        if isinstance(yaml_dict, YamlCommandDescriptor):
-            for k, v in yaml_dict.items():
-                self[k] = v
-            return
-        yaml_dict_copy = yaml_dict.copy()
-        for label, parameters in yaml_dict_copy.items():
-            self[label] = str(Command(kwflags_join=',',
-                                      kwflags_sep='=',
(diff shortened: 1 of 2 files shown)


setrofim:
This is too opaque. Configuring this for someone unfamiliar would be a nightmare. Rather than having a single "command" map, parameters should be exposed individually, validated, and populated with sensible defaults whenever possible. 

As a rule of thumb -- I should be able to run `wa show perf`, and the resulting page should be enough to allow me to at least start using the instrument without also having to look up the man page for perf and/or devlib docs. When I run `wa create create agenda perf dhrystone`, the resulting agenda should generate *some* kind of reasonable output for perf.

pietos01:
You're right, this was written while making the assumption that the user would know how to use `perf` (and this is the reason for which this PR is trying to give to them as much control as possible).

The difficulty with replicating `perf` parameters individually at the input of WA comes from how many subcommands there are, and how each has many flags. It becomes even more complicated to get right once we try to validate them as the simple availability of a flag (let alone the validity of the passed value) depends not only on the kernel version but also on the build flags of `perf` itself.

About the defaults, `perf` already provides a default behaviour for each subcommand (and default values for flags that require values _e.g._ `event`); isn't it confusing to add our own WA defaults, different from the `perf` defaults? Obviously, there's no default subcommand (running `perf` is equivalent to `perf --help`) but having one would barely make sense, in my opinion, because of how different they are.

I *really* like `wa show` as it's an easy way to figure out what to put in the agenda without having to go through documentation (or source code!). For this tool, it would give the meaning of `force_install`, `pre_commands`, `commands`, `post_commands` and define the formats they are expected to be in and how these relate to what ends up being run on the device. Of course, because we can't clearly define what goes inside of the values of the pairs in the `commands` dicts (see above), there's little we can tell about them in the output of `wa show`. In fact, as you say, I intend the help of this instrument to end with a RTFM-ish recommendation to refer to `perf --help`. Documentation of devlib isn't necessary for knowing how to use this instrument, though.

In the end, if the issue is with `wa create perf`, couldn't we simply have a default `commands` (probably calling `stat -a -- sleep 1000` as in the previous implementation)? I haven't looked at how `create` was implemented so this could be hard to do (_e.g._ if it simply takes all `default` values of the `Parameter`s).

setrofim:
>The difficulty with replicating perf parameters individually at the input of WA comes from how many subcommands there are, and how each has many flags. It becomes even more complicated to get right once we try to validate them as the simple availability of a flag (let alone the validity of the passed value) depends not only on the kernel version but also on the build flags of perf itself.

Yes, I would not suggests trying to handle everything in `perf` that way  -- that would ultimately be futile. Instead I would consider some higher-level use cases for why someone might want to use `pref` (e.g.  "I want to get PMU counter values for the execution of this workload) and create a high level interface for that, which would work, say, 90% of the time. Then allow it to be augmented with the pass-though configuration you have at the moment, which will be the fallback for  when the simple setup fails (or when the user that knows what they are doing wants to do something different). 

But the pass-though configuration really should be treated as a fall back, not as the intended primary was of using the instrument in the "common case".

>In the end, if the issue is with wa create perf, couldn't we simply have a default commands (probably calling stat -a -- sleep 1000 as in the previous implementation)? I haven't looked at how create was implemented so this could be hard to do (e.g. if it simply takes all default values of the Parameters).

Yes, that's pretty much how `create` command works, hence exposing parameters with sensible defaults is important to usability.

pietos01:
> But the pass-though configuration really should be treated as a fall back, not as the intended primary was of using the instrument in the "common case".

That's feasible ... and would make everyone happy.

So this means effectively merging the current implementation of the instrument (from `master`) with this new one? We would have the "clear and easy-to-use" parameters for `stat` (with nice default values, allowing to easily `create` agendas) and the (`pre_`|` `|`post_`)`commands` (defaulting to `None`) for advanced usage (which includes any subcommands other than `stat`) with the "old" parameters being ignored once `commands is not None`?

This approach could then also allow backward compatibility (instead of what I'm doing right now) and is very easy to integrate as the constructor could simply build a `command` dictionary from these parameters and the rest of the execution flow would remain unchanged!

setrofim:
Yes, that's the sort of thing I had in mind!

wa/instruments/perf.py
```diff
@@ -13,126 +13,220 @@
...
+                    'args': {
+                        'command': 'sleep',
+                        'args': 1000,
+                    },
+                    'stderr': '&1',
+                    'stdout': 'stat.out',
+                }
+                for label, options in zip(self.labels, self.optionstring)
+            })
+        # TODO: else, detect if events or optionstring were passed and error?
```
pietos01: Is there a standard way for checking that a `Parameter` was passed through the agenda or not? I would like to `raise` and error if the user attempts to use any of `labels`, `optionstring` or `events` when they've passed `command`. I was thinking of testing these variables against their defaults but that doesn't detect the case where an agenda uses incorrectly these parameters while passing their default values ...

wa/instruments/perf.py
```diff
@@ -13,126 +13,220 @@
...
-                            if match:
-                                classifiers = {}
-                                cpu = match.group(1)
-                                if cpu is not None:
-                                    classifiers['cpu'] = int(cpu.replace('CPU', ''))
-                                count = int(match.group(2))
-                                metric = '{}_{}'.format(label, match.group(3))
-                                context.add_metric(metric, count, classifiers=classifiers)
+        self.collector.get_traces(outdir)
+        # HUGE TODO: add parsers for supported post_commands
```
pietos01: Now that we support any `perf` command, what are the expectations for the output parser? I understand that WA captures those metrics into a JSON output file. Although these are adapted for cases having limited numbers of metrics (such as `perf stat`, as was done previously), it might not be ideal for large tables similar to what `perf report` may output. Are we expecting this instrument to parse this table into those JSON results? it's feasible and seems to meet what was done previously but is probably not the most practical approach.

wa/instruments/perf.py
setrofim: We don't have a mechanism for this at the moment. We typically deal with this by not specifying a `default` inside the prameter, but setting it inside `__init__` if the value is `None`.

Now that I think about it, adding such a mechanism should be straight forward.

setrofim: Erm, actually, turns out it's not that straight forward. Never mind.

wa/instruments/perf.py
setrofim: WA supports two kinds of outputs from  plugins: metrics and artifiacts. Metrics are individual values intended to be viewed and processed individually, artefacts are external entities (usually files) that are intended to be be handed as a monolithic unit. 

When you have something that generates a large number of values you have two options: add those values as individual metrics, or write them to a file (e.g. as a csv table) and add that file as an artefact. 

Which of these approaches you chose should be a function of how those values are intended to be used, rather than their number. If you expect to be able to query and compare individual values (e.g. across runs, or across configurations within a single run), they should be added as metrics. If the values only make sense when viewed as a group (e.g. it's a trace, a time series, etc), then they should written out into an artefact.

wa/instruments/perf.py
pietos01: Done (a bit hacky, though)

wa/instruments/perf.py
```diff
@@ -13,126 +13,385 @@
...
-                                       self.force_install)
+                                       self.force_install,
+                                       self.pre_commands,
+                                       self.commands,
+                                       self.post_commands)
 
     def setup(self, context):
         self.collector.reset()
+        version = self.collector.execute('--version').strip()
+        context.update_metadata('versions', self.name, version)
```
pietos01: I hope this is being logged in the right "metadata".

I've noticed the following (optional) idea that hasn't been addressed (from the original post):

> The force_install flag is kept from the previous implementation. However, what is the "standard WA way" of providing the perf binary? We could write the path in the agenda (e.g. for comparing versions of perf), we could have a collection of perf binaries in a standard location and pick the ideal one (e.g. based on kernel version of the target) from within the WA perf instrument, we could (but this requires automating the automation, WA, and feels messy) have WA get the binary from a standard location on the host and overwrite that one between runs... Ideally, it would be useful to have both control of which binaries are being used and automation for finding the optimal version (as I believe the perf-kernel interface evolves with the kernel).

Based on ARM-software/devlib#395 and ARM-software/devlib#396, this may be useful but might be complicated to implement(?)

setrofim: Yes, this is the correct metadata.

WA has  a standard mechanism for "resource resolution" that allows providing alternative versions of resources via various means.

https://workload-automation.readthedocs.io/en/latest/developer_information.html#dynamic-resource-resolution

See how `dhrystone` workload obtains its executable for example:

https://github.com/ARM-software/workload-automation/blob/master/wa/workloads/dhrystone/__init__.py#L78

wa/instruments/perf.py
```diff
@@ -245,11 +238,160 @@ def stop(self, context):
...
+
+        Parameters:
+            match        A :class:`re.Match` from :attr:`_stat_counter_regex`
+            classifiers  A dictionary to be completed for the matched counter
+
+        Returns:
+            A (name, value) tuple for the matched counter (value is 0 if an
+            error occurred).
+        """
+        name = f'{classifiers["label"]}_{match["name"]}'.replace(' ', '_')
```
setrofim: As per earlier discussion, we'd like to continue supporting Python 3.5 for the foreseeable future in order to avoid forcing potentially  painful migrations on existing WA users. Because of that, 3.6 features, such as format strings, should not be used in WA code.

wa/instruments/perf.py
```diff
@@ -13,126 +13,385 @@
...
+          fraction (between ``0.0`` and ``100.0``) of the run that a hardware
+          counter was allocated to the the event. Available only for hardware
+          events and only when time-slicing was required.
+        - ``'comment_value'``, ``'comment_units'``: Some counters may come with
+          an extra "comment" (following a ``#``) added by ``perf``. The
+          ``'comment_value'`` holds the numeric (``int`` or ``float``) value of
+          the comment while ``'comment_units'`` holds the rest of the comment
+          (typically the units). Only available for the events for which
+          ``perf`` added a comment.
+        """
```
pietos01: How can we get this documentation to render to HTML? Appending it to the class docstring doesn't really feel right.

setrofim: put this in the `description` class attribute of the plugin.

pietos01: Done

wa/instruments/perf.py
pietos01: I wasn't sure what the conclusion of ARM-Software/devlib#389 was; I have changed the f-strings to `format` calls.

wa/instruments/perf.py
pietos01: But `dhrystone` is picking its binaries based on the ABI which is pretty easy to check with an ELF reader; what I was considering was to get the version of `perf` which (AFAIK) can only be obtained by running the binaries themselves ... which requires something able to run them. Another approach would require storing this information in the file system (_e.g._ in the name).

Let's drop this idea due to its complexity.

setrofim: >But dhrystone is picking its binaries based on the ABI which is pretty easy to check with an ELF reader;

Actually, it's picking the binary based on an arbitrary string, which happens to be the ABI -- we're not checking the ELF header, just resolving based on directory structure. So the same can be done for perf (it's basically what you're saying about encoding in the name, except we use directories rather than modifying the file name).

wa/instruments/perf.py
setrofim: Yeah, that discussion kinda petered out without a concrete conclusion. However, I assume that earlier issue regarding support for Ubuntu 16.04 stands, and since there is nothing _essential_ in 3.6, we're going to stick with 3.5 for the next release, at least.

wa/instruments/perf.py
pietos01: I see, my bad for not looking further through the call stack of the code you shared!

I believe this is a good idea. However, this feature isn't required for using `perf` as a single version is enough in almost all cases. As users of WA-`perf` seem to have been satisfied with the single version approach until now, let's keep this feature for a potential future PR, once we are certain it will be necessary and used!

wa/instruments/perf.py
```diff
@@ -13,126 +13,385 @@
...
+from devlib.trace.perf import PerfCollector, PerfCommandDict
 from wa import Instrument, Parameter
 from wa.utils.types import list_or_string, list_of_strs
 
-PERF_COUNT_REGEX = re.compile(r'^(CPU\d+)?\s*(\d+)\s*(.*?)\s*(\[\s*\d+\.\d+%\s*\])?\s*$')
+__all__ = [
+    'PerfInstrument',
+]
+
+DEFAULT_EVENTS = ['migration', 'cs']
```
marcbonnici: The `migration` default event should be `migrations`.

pietos01: Done

pietos01:
I've added logging of the artefacts extracted from the device.

pietos01:
Due to the complexity of this implementation and the resulting instability of the solution I have experienced, combined with the difficulties of integrating it with the WA framework, this PR will be closed in favour of an upcoming one, built on top of the Python scripts provided for `simpleperf`. Concerning `perf`, #971 may be enough to cover the needs of the users.

PR #986, opened 2019-05-28
workloads/uibench: Pylint Fix

marcbonnici:


pietos01:
Sorry, I hadn't noticed you had a `.pylintrc` (as it's not at the root of the repo). I've integrated that in my development setup so that we can avoid the need for PRs like this one in the future :)

PR #993, opened 2019-06-24
Add a workload for UIBench JankTests

pietos01:
This workload uses `am instrument` to launch the official AOSP automation of UIBench (see the existing WA workload for that one).

### DIFF
--- a/wa/framework/workload.py
+++ b/wa/framework/workload.py
@@ -1,4 +1,4 @@
-#    Copyright 2014-2018 ARM Limited
+#    Copyright 2014-2019 ARM Limited
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,7 +14,13 @@
 #
 import logging
 import os
+import threading
 import time
+
+try:
+    from shlex import quote
+except ImportError:
+    from pipes import quote
 
 from devlib.utils.android import ApkInfo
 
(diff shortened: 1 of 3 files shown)


wa/framework/workload.py
```diff
@@ -865,3 +865,42 @@ def teardown(self):
...
+        w_flag = '-w' if self.wait else ''
+        n_flag = '--no-hidden-api-checks' if self.no_checks else ''
+        flags = ' '.join(filter(None, [w_flag, r_flag, n_flag, options]))
+
+        target = (self.apk_info.package if not self.apk_info.activity else
+                  '{}/{}'.format(self.apk_info.package, self.apk_info.activity))
+        self.cmd = 'am instrument {} {}'.format(flags, target)
+
+    def start_activity(self):
+        """NB: Unlike regular applications, this is a blocking call."""
```
setrofim: This breaks the substitution principle. Ether:

- Make this non-blocking by kicking it of  in a thread internally ant returning right away (and then prividing `wait()` or similar), or
-  Create a common base for this and `PackageHandler`, rather than subclassing `PackageHandler`, and implement a `run_activity()` rather than `start_activity()`.

pietos01: This difference in behaviour comes from the difference between `am start` (non-blocking) and `am instrument` (blocking) and my attempt at adding support for `am instrument`-based APKs in a somewhat generic yet lightweight way (in terms of required change).

I agree that the best way would be to rethink the class inheritance tree to properly integrate that distinction while keeping polymorphism but that's more than I'm ready to take up right now. I've implemented the non-blocking approach and the workload's `run` method starts the activity then waits for the instrument to return, while the `TestPackageHandler.start_activity` remains non-blocking (as it creates a thread on the host).

wa/workloads/uibenchjanktests/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+
+            if self.test:
+                klass += '#{}'.format(self.test)
+            self.apk.args['class'] = klass
+
+    _OUTPUT_SECTION_REGEX = re.compile(
+            r'(\s*INSTRUMENTATION_STATUS: gfx-[\w-]+=[-+\d.]+\n)+'
+            r'\s*INSTRUMENTATION_STATUS_CODE: (?P<code>-+\d)\n?', re.M)
+    _OUTPUT_GFXINFO_REGEX = re.compile(
+            r'INSTRUMENTATION_STATUS: (?P<name>[\w-]+)=(?P<value>[-+\d.]+)')
```
marcbonnici: Could these be placed at the beginning of the class with the others?

wa/framework/workload.py
```diff
@@ -14,6 +14,8 @@
 #
 import logging
 import os
+import shlex
```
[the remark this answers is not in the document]
pietos01: I've changed that to a reliable import of `quote` (which is slightly different and arguably more robust than what I had already done in another (the `perf`?) PR).

wa/workloads/uibenchjanktests/__init__.py
pietos01: Done

pietos01:
Let me know if any more changes are required or if the PR is accepted, in which case I'll autosquash the commit stack.

marcbonnici:
Sorry, a few additional minor stylistic corrections from https://www.travis-ci.org/ARM-software/workload-automation/builds/551205505
```
wa/workloads/uibenchjanktests/__init__.py:34:13: E126 continuation line over-indented for hanging indent
wa/workloads/uibenchjanktests/__init__.py:37:13: E126 continuation line over-indented for hanging indent
wa/workloads/uibenchjanktests/__init__.py:108:43: W503 line break before binary operator
wa/framework/workload.py:875:1: E302 expected 2 blank lines, found 1
```

pietos01:
Is Travis using a different `.pylintrc` than the one provided with WA?

It has been updated (and rebased).

marcbonnici:
No it shouldn't be, the command we use within travis to run pylint should be the same running it manually.
(https://github.com/ARM-software/workload-automation/blob/master/.travis.yml#L32)

PR #994, opened 2019-06-26
Add a workload for Motionmark tests

scojac01:


wa/workloads/motionmark/__init__.py
```diff
@@ -0,0 +1,77 @@
+#    Copyright 2014-2018 ARM Limited
```
setrofim: Please make sure the copyright is correct.

wa/workloads/motionmark/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+from wa import UiautoWorkload, Parameter
+from wa.framework.exception import ValidationError, WorkloadError
+from wa.utils.types import list_of_strs
+from wa.utils.misc import unique
+
+
+class Motionmark(UiautoWorkload):
+
+    name = 'motionmark'
+    regex_matches = [re.compile(r'Multiply Score (.+)'),
```
setrofim: Theses should go after the description, parameters, and other "externally-facing" attributes.

wa/workloads/motionmark/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    '''
+
+    def __init__(self, target, **kwargs):
+        super(Motionmark, self).__init__(target, **kwargs)
+        self.gui.timeout = 1500
+
+    def setup(self, context):
+        super(Motionmark, self).setup(context)
+        url = 'am start -a android.intent.action.VIEW -d https://browserbench.org/MotionMark/'
+        self.target.execute(url)
```
setrofim: Why not use `self.gui.start_activity()` here?

wa/workloads/motionmark/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+                for regex in self.regex_matches:
+                    match = regex.search(line)
+                    # Check if we have matched the score string in logcat
+                    if match:
+                        score_match = self.score_regex.search(match.group(1))
+                        # Check if there is valid number found for the score.
+                        if score_match:
+                            result = float(score_match.group(1))
+                        else:
+                            result = 'NaN'
```
setrofim: should be `result = float('NaN')`

wa/workloads/motionmark/__init__.py
setrofim: Also, rename this variable as it contains regex's rather than regex matches.

wa/workloads/motionmark/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+                        entry = regex.pattern.rsplit(None, 1)[0]
+                        context.add_metric(entry, result, 'Score', lower_is_better=False)
+                        expected_results -= 1
+        if expected_results > 0:
```
setrofim: Variable name does not match its function. Rename to something like `num_unprocessed_results`. Or even better, why not just throw an exception if `match` above is `None` and remove the need for this variable entirely?

wa/workloads/motionmark/__init__.py
scojac01: Hi Sergei, 

I'm wondering what is this method? How does it work and what is the benefit over what I have?

wa/workloads/motionmark/__init__.py
scojac01: I assume you mean the "score_match"?

wa/workloads/motionmark/__init__.py
setrofim: The `gui` object encapsulates all GUI interactions, so the workload itself does not issue specific GUI commands, and instead deals with the abstracted "gui".

The benefit is that all GUI interactions are preformed in a "standard" manner. If something needs to be updated (e.g. to handle a future android version) it would need to be updated in only one place. This also introduces consistency across different workloads' code.

wa/workloads/motionmark/__init__.py
setrofim: Erm, ok, never mind he second part -- I assumed you were matching the entire text; as you matching individual lines, that would not work. Just rename `expected_results` to something more representative of it's actual function.

wa/workloads/motionmark/__init__.py
marcbonnici: The `start_activity` method is actually part of the `PackageHandler` rather than `UiAutomatorGUI` (`gui`). In this case it is relying on a pre-installed browser to execute the action rather than installing one and therefore wouldn't have access to that method. We do however have an `open_url` method available as part `AndroidTarget` which could be used instead.

wa/workloads/motionmark/uiauto/app/src/main/java/com/arm/wa/uiauto/motionmark/UiAutomation.java
```diff
@@ -0,0 +1,182 @@
...
+        setScreenOrientation(ScreenOrientation.RIGHT);
+        UiScrollable list = new UiScrollable(new UiSelector().scrollable(true));
+
+        UiObject start =
+            mDevice.findObject(new UiSelector().text("Run Benchmark")
+                .className("android.widget.Button"));
+        list.scrollIntoView(start);
+        start.click();
+
+        setScreenOrientation(ScreenOrientation.NATURAL);
```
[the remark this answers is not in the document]
scojac01: This is rotating back once the test has started. 

The workflow should be 

Portrait by default when launched
Switch to landscape to run the tests
Return to portrait to see all the results on one screen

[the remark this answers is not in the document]
marcbonnici: Ok, in which case should that call not be after `results.waitForExists(2100000);` as this will signify the tests have completed and the results are ready to be extracted?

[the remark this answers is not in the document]
scojac01: Yep, which I changed to on the latest commit after revisiting :)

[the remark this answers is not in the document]
scojac01: Well I thought I had but it's not showing here so I will redo that.

wa/workloads/motionmark/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    description = '''
+    A workload to execute the motionmark web based benchmark
+
+    Test description:
+    1. Open browser application
+    2. Navigate to the motionmark website - http://browserbench.org/motionmark/
+    3. Execute the benchmark
+    '''
+
+    name = 'motionmark'
```
setrofim: `name` above `description`.

wa/workloads/motionmark/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+class Motionmark(UiautoWorkload):
+
+    description = '''
+    A workload to execute the motionmark web based benchmark
+
+    Test description:
+    1. Open browser application
+    2. Navigate to the motionmark website - http://browserbench.org/motionmark/
+    3. Execute the benchmark
```
setrofim: Please add a  sentence or two to describe what the bechmark actually does.

wa/workloads/motionmark/__init__.py
setrofim: Should set `requires_network` to `True`.

wa/workloads/motionmark/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+                        if score_match:
+                            result = float(score_match.group(1))
+                        else:
+                            result = float('NaN')
+                        entry = regex.pattern.rsplit(None, 1)[0]
+                        context.add_metric(entry, result, 'Score', lower_is_better=False)
+                        num_unprocessed_results -= 1
+        if num_unprocessed_results > 0:
+            msg = "The Motionmark workload has failed. Expected {} scores, Detected {} scores."
+            raise WorkloadError(msg.format(len(self.regex), num_unprocessed_results))
```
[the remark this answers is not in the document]
scojac01: I don't think so. 

We set num_unprocessed _results = len(self.regex) on line 62 so at this point it would be 9. 

After a entry is processed correctly and added we minus the num_unprocessed_results by 1. 

If all are processed correctly num_unprocessed_results will be 0. 

The exception is then raised if num_unprocessed_results is > 0.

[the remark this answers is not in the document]
marcbonnici: Sorry I agree the logic of raising the error if correct, I meant in the message itself it lists the number of expected scores and the number of "detected scores" which in this case would actually be reporting the inverse, i.e. how many score were not detected. I would suggest either rewording the error message or adjusting the value reported.

wa/workloads/motionmark/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+    name = 'MotionMark'
+
+    description = '''
+    A workload to execute the motionmark web based benchmark
+
+    MotionMark is a graphics benchmark that measures a browser capability to animate complex scenes at a target frame rate
+
+    Test description:
+    1. Open browser application
+    2. Navigate to the motionmark website - http://browserbench.org/motionmark/
```
marcbonnici: Please change this url to the correct capitalisation of the destination site `http://browserbench.org/MotionMark/`

rabbivalentino:
Kamlerhans@gmail.com

PR #1002, opened 2019-07-18
Storing directories as Artifacts + Postgres Fixes

marcbonnici:
This set of commits allows for storing directories as an Artifact and implements the functionality of being transparently uploaded and retrieved from a database.

Items to note:
- Upon uploading artifacts to the database the directory will be compressed and stored as a `.tar.gz` file.
- Upon invocation of `get_artifact_path` the `.tar.gz` file will be streamed from the database and  extracted to a temporary location with the prefix `'wa_'` before returning the path of the newly extracted directory. It is currently left to the user to delete the folder once they are finished with it. 
- The naming scheme for the `SysfsExtractor` instrument artifacts is currently in the form `<path> [stage]` for example `/sys/devices/system/cpu [before]`.

### DIFF
--- a/doc/source/api/output.rst
+++ b/doc/source/api/output.rst
@@ -315,9 +315,12 @@ methods
 
 .. method:: RunDatabaseOutput.get_artifact_path(name)
 
-    Returns a `StringIO` object containing the contents of the artifact
-    specified by ``name``. This will only look at the run artifacts; this will
-    not search the artifacts of the individual jobs.
+    If the artifcat is a file this method returns a `StringIO` object containing
+    the contents of the artifact specified by ``name``. If the aritifcat is a
+    directory, the method returns a path to a locally extracted version of the
+    directory which is left to the user to remove after use. This will only look
+    at the run artifacts; this will not search the artifacts of the individual
+    jobs.
 
     :param name:  The name of the artifact who's path to retrieve.
     :return: A `StringIO` object with the contents of the artifact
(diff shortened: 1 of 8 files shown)


setrofim:
Do we need the `is_dir` parameter for `add_artifact`? Can we not simply test `os.path.isdir()` when creating the actual `Artifact` object?

setrofim:
> Upon retrieving the filepath of a directory artifact the .tar.gz file will be extracted to a temporary location with the prefix 'wa_' before returning the new file path. It is currently left to the user to delete the folder once they are finished with it.

I'm not sure I understand the logic here. Should it not return the path to the extracted directory structure?

marcbonnici:
> Do we need the `is_dir` parameter for `add_artifact`? Can we not simply test `os.path.isdir()` when creating the actual `Artifact` object?

Good point that would be a nicer way of doing things, the only issue I can see is that the `Artifact` only knows its relative path,  so will need to think if there is an easy way of doing this.  

> I'm not sure I understand the logic here. Should it not return the path to the extracted directory structure?

Yes sorry that should say the new file path of the extracted directory. The actual `.tar.gz` file is read directly from the database when invoked via `get_artifact_path()` and the file itself is not stored on the host machine.

setrofim:
> the only issue I can see is that the Artifact only knows its relative path, so will need to think if there is an easy way of doing this.

Yeah, good point. That will need to be done here: https://github.com/ARM-software/workload-automation/blob/master/wa/framework/output.py#L148

So `Output.add_artifact()` doesn't need `is_dir`, but `Result.add_artifiact()` does.

PR #1014, opened 2019-09-06
Update gmail workload to work against most recent version

robF8:
* Updated gmail to work against 2019.05.26.252424914.release.
* Removed google photos dependency to stop incorrect image being used.


wa/workloads/gmail/uiauto/app/src/main/java/com/arm/wa/uiauto/gmail/UiAutomation.java
```diff
@@ -189,58 +229,28 @@ public void attachImage() throws Exception {
         if (rootMenu.exists()){
             rootMenu.click();
         }
-        // Check for Photos
```
marcbonnici: Not all devices have the "Images" entry in the sidebar (testing on nexus5) and therefore still rely on GooglePhotos. I would suggest leaving the use of google photos as a fall back option.

robF8:
Re-added google photos as a fall back in case images entry is unavailable

PR #1017, opened 2019-09-19
Antutu: Updating the workload to handle the shortcut popup

scojac01:


wa/workloads/antutu/uiauto/app/src/main/java/com/arm/wa/uiauto/antutu/UiAutomation.java
```diff
@@ -44,7 +44,8 @@
 
     @Test
     public void setup() throws Exception {
-        dismissAndroidVersionPopup();
+       //dismissAndroidVersionPopup();
```
setrofim: Remove commented code.

wa/workloads/antutu/uiauto/app/src/main/java/com/arm/wa/uiauto/antutu/UiAutomation.java
```diff
@@ -65,6 +65,15 @@ public void hitTest() throws Exception {
         sleep(1);
     }
 
+    public void clearPopups() throws Exception {
+        UiObject cancel =
+            mDevice.findObject(new UiSelector().textContains("CANCEL"));
+        cancel.waitForExists(60000);
```
marcbonnici: A minute seems a bit long for this timeout? Do you know of any reasons it would take more than a few seconds for the pop to appear if it is going to?

scojac01: No reasoning behind it really. What would the suggested timeout value be?

marcbonnici: I think our default timeout that we have used is something more like 4 seconds depending on the type of popup it is.

PR #1026, opened 2019-10-03
Cpustates Improvments

marcbonnici:
Improvements to the cpustate output processor when dealing with missing information.


setrofim:
The second commit message is a little misleading as it deals with cpuidle as well as cpufreq.

marcbonnici:
Split the second commit into separate cpufreq and cpuidle commits.

PR #1038, opened 2019-11-22
Geekbench: Updating supported versions

scojac01:


setrofim:
Just to confirm -- these versions have been confirmed to be working with the existing implementation. Or should there be additional changes to make the workload work with these new versions?

Either way, could please add a commit message body stating which it is.

scojac01:
Correct, the updated versions are working and tested with the existing implementation.

setrofim:
Great! Please add that to the commit message.

PR #1041, opened 2019-11-28
PCMark: Updating to handle new Android 10 permission warnings

scojac01:
Android 10 has introduced a new permission warning and a
seperate warning about the APK being built for an older version
of the Android OS. Have added two checks to accept these
permissions and continue with the workload and the given APK
rather than attempting to update.


wa/workloads/pcmark/uiauto/app/src/main/java/com/arm/wa/uiauto/pcmark/UiAutomation.java
```diff
@@ -47,6 +47,7 @@ public void initialize(){
     @Test
     public void setup() throws Exception{
         dismissAndroidVersionPopup();
+        clearPopups();
```
marcbonnici: If the version check is performed by the OS I'm guessing this will start affecting other applications as well? 
If so would it be possible to updated the existing `dismissAndroidVersionPopup()` method instead of an additional check here so that other applications can benefit as well?

scojac01: We've only seen it on the PCMark application so far. Otherwise this would be something I'd have added there as you suggest.

marcbonnici: Ok, if this does turn out to be the case we can move it across then.

PR #1052, opened 2020-01-10
fw/getters: update Executable resolution

setrofim:
Use Executable.match() rather than just checking the path inside
get_from_location(); this allows for alternative matching semantics
(e.g. globbing) inside derived implementations.

### DIFF
--- a/wa/framework/getters.py
+++ b/wa/framework/getters.py
@@ -84,9 +84,13 @@ def get_from_location(basepath, resource):
         if os.path.exists(path):
             return path
     elif resource.kind == 'executable':
-        path = os.path.join(basepath, 'bin', resource.abi, resource.filename)
-        if os.path.exists(path):
-            return path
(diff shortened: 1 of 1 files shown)


wa/framework/getters.py
```diff
@@ -84,7 +84,13 @@ def get_from_location(basepath, resource):
...
-        path = os.path.join(basepath, 'bin', resource.abi, resource.filename)
+        bin_dir = os.path.join(basepath, 'bin', resource.abi)
+        if not os.path.exists(bin_dir):
+            return None
+        for entry in os.listdir(bin_dir):
+            path = os.path.join(bin_dir, entry)
+            if resource.match(path):
+                return path
         if os.path.exists(path):
             return path
```
marcbonnici: This should be removed otherwise this will always return a path?

setrofim: Good point.

PR #1054, opened 2020-01-14
Antutu: Updating to work with major version 8.

scojac01:
The update to Antutu major version 8 has changed a lot of element names.
There have also been changes to the tests run in three of the four categories.

This commit handles those updates while also retaining backwards compatibility
with major version 7.


wa/workloads/antutu/__init__.py
```diff
@@ -33,32 +33,81 @@ class Antutu(ApkUiautoWorkload):
...
-                            result = float(match.group(1))
-                        except ValueError:
-                            result = float('NaN')
-                        entry = regex.pattern.rsplit(None, 1)[0]
-                        context.add_metric(entry, result, lower_is_better=False)
-                        expected_results -= 1
-        if expected_results > 0:
-            msg = "The Antutu workload has failed. Expected {} scores, Detected {} scores."
-            raise WorkloadError(msg.format(len(self.regex_matches), expected_results))
+        if '8' in self.version:
```
setrofim: Maybe use `if self.version.startswith('8')` as it's a bit more robust.

wa/workloads/antutu/__init__.py
```diff
@@ -33,32 +33,81 @@ class Antutu(ApkUiautoWorkload):
...
 
-    Known working APK version: 7.0.4
+    Known working APK version: 8.0.4
     '''
+   
+    supported_versions = ['7.0.4', '7.2.0', '8.0.4', '8.1.9']
+
+    parameters = [
+        Parameter('version', kind=str, allowed_values=supported_versions, override=True,
+                 description=('Specify the version of Antutu to be run. '
```
setrofim: Use multi-line strings for descriptions, as the result is cleaner:
```
                 description='''
                 Specify the version of Antutu to be run.  If not specified, the latest available 
                 version will be used.
                 ''')
```

wa/workloads/antutu/__init__.py
```diff
@@ -33,32 +33,81 @@ class Antutu(ApkUiautoWorkload):
...
+                            except ValueError:
+                                result = float('NaN')
+                            entry = regex.pattern.rsplit(None, 1)[0]
+                            context.add_metric(entry, result, lower_is_better=False)
+                            expected_results -= 1
+            if expected_results > 0:
+                msg = "The Antutu workload has failed. Expected {} scores, Detected {} scores."
+                raise WorkloadError(msg.format(len(self.regex_matches_v8), expected_results))
+        if '7' in self.version:
+            expected_results = len(self.regex_matches_v7)
```
setrofim: Rather than duplicating the loop for version 7, split common code out into a function, and pass the regex matches to use as a parameter.

wa/workloads/antutu/uiauto/app/src/main/java/com/arm/wa/uiauto/antutu/UiAutomation.java
```diff
@@ -56,7 +65,11 @@ public void runWorkload() throws Exception{
 
     @Test
     public void extractResults() throws Exception{
-        getScores();
+        if (version.contains("8")){
```
setrofim: Again, `startsWith("8")` is more robust.

wa/workloads/antutu/__init__.py
```diff
@@ -14,51 +14,87 @@
...
+                    If not specified, the latest available version will be used.
+                    ''')
+                )
+    ]
+
+    def setup(self, context):
+        self.gui.uiauto_params['version'] = self.version
+        super(Antutu, self).setup(context)
+
+    def checkScores(self, context, regex_version):
```
setrofim: Don't use camel case in Python code. Also, this extracts metrics rather than merely "checking" that the scores are present, so the name is misleading. So something like `extract_scores` would be better.

wa/workloads/antutu/__init__.py
```diff
@@ -61,4 +90,11 @@ def update_output(self, context):
                         expected_results -= 1
         if expected_results > 0:
             msg = "The Antutu workload has failed. Expected {} scores, Detected {} scores."
-            raise WorkloadError(msg.format(len(self.regex_matches), expected_results))
+            raise WorkloadError(msg.format(len(self.regex_version), expected_results))
```
marcbonnici: `self.regex_version` should not have `self.`

wa/workloads/antutu/__init__.py
```diff
@@ -14,42 +14,71 @@
...
+                      If not specified, the latest available version will be used.
+                      ''')
+                  )
+    ]
+
+    def setup(self, context):
+        self.gui.uiauto_params['version'] = self.version
+        super(Antutu, self).setup(context)
+
+    def extract_scores(self, context, regex_version):
```
marcbonnici: Could you add a `# pylint: disable=no-self-use` to `extract_scores` to keep pylint happy?

PR #1057, opened 2020-01-17
fw/exec: Add add_classifier() method

setrofim:
Add add_classifier() method to context. Allow plugins to add classifiers
to the current job, or the run as a whole. This will ensure that the new
classifiers are propagated to all relevant current and future artifacts
and metrics.

### DIFF
--- a/wa/framework/output.py
+++ b/wa/framework/output.py
@@ -165,6 +165,9 @@ class Output(object):
         artifact = self.get_artifact(name)
         return self.get_path(artifact.path)
 
+    def add_classifier(self, name, value, overwrite=False):
+        self.result.add_classifier(name, value, overwrite)
+
     def add_metadata(self, key, *args, **kwargs):
         self.result.add_metadata(key, *args, **kwargs)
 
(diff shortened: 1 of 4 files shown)


wa/framework/output.py
```diff
@@ -410,6 +413,17 @@ def get_artifact(self, name):
...
+    def add_classifier(self, name, value, overwrite=False):
+        if name in self.classifiers and not overwrite:
+            raise ValueError('Cannot overwrite "{}" classifier.'.format(name))
+        self.classifiers[name] = value
+
+        for metric in self.metrics:
+            metric.classifiers[name] = value
+
+        for artifact in self.artifacts:
+            artifact.classifiers[name] = value
```
marcbonnici: Is it worth checking to see if the classifier already exists for any metrics or artifacts?

setrofim: I considered that. However, the way those classifiers would typically get there is form the job, so the `overwrite` check above would already handle them.

The other way, is that if the plugin itself adds them when adding the metric, but then the plugin author is responsible for both, and I figured it's better to let them sort it out at plugin implementation level, as here, we can't tell the difference between this case and the first.

setrofim: Actually, scratch that. It does't really matter. We _should_ check for clashes here.

PR #1063, opened 2020-02-04
fw/plugin: move cleanup_assets to TargetedPlugin

setrofim:
Move cleanup_assets from Workload up into TargetedPlugin. This way,
Instruments may also utilize it if they deploy assets.

More generally, it makes sense for it to be inside TargetedPlugin, as
any plugin that interacts with the target may conceivably need to clean
up.

### DIFF
--- a/wa/framework/plugin.py
+++ b/wa/framework/plugin.py
@@ -384,6 +384,16 @@ class TargetedPlugin(Plugin):
     """
 
     suppoted_targets = []
+    parameters = [
+        Parameter('cleanup_assets', kind=bool,
+                  global_alias='cleanup_assets',
+                  aliases=['clean_up'],
+                  default=True,
+                  description="""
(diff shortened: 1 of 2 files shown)


wa/framework/plugin.py
```diff
@@ -384,6 +384,16 @@ class TargetedPlugin(Plugin):
...
 
     suppoted_targets = []
+    parameters = [
+        Parameter('cleanup_assets', kind=bool,
+                  global_alias='cleanup_assets',
+                  aliases=['clean_up'],
+                  default=True,
+                  description="""
+                  If ``True``, assets that are deployed or created as part of the
+                  workload will be removed again from the device.
```
marcbonnici: Should "workload" now be changed to something more generic?

setrofim: Good point. Have updated.

PR #1067, opened 2020-02-07
instruments: add proc_stat

setrofim:
Add an instrument that monitors CPU load using data from /proc/stat

### DIFF
--- a/wa/instruments/proc_stat/__init__.py
+++ b/wa/instruments/proc_stat/__init__.py
@@ -1,0 +1,94 @@
+#    Copyright 2020 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


wa/instruments/proc_stat/gather-load.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
marcbonnici: This shebang causes the script to fail on devices that do not have sh present at that path. Can this be updated to invoke via the busybox implementation?

setrofim: Yup, done.

wa/instruments/proc_stat/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    description = '''
+    Collect CPU load information from /proc/stat.
+    '''
+
+    parameters = [
+        Parameter('period', int, default=5,
+                    constraint=lambda x: x > 0,
+                    description='''
+                    Time (in seconds) between collections.
+                    '''),
```
marcbonnici: Could you align these lines to keep pep8 happy? 
(https://travis-ci.org/ARM-software/workload-automation/jobs/647303406)

setrofim: Ah my bad. I had an old version of `flake8` on my machine that didn't pick this up until I upgraded it. Fixed now.

PR #1080, opened 2020-03-26
Force speedometer to use chrome browser

robF8:
* Workload was failing when chrome was not set as default browser so
  altered to use chrome every time.
* On some devices (mate20) chrome gets into a state where it keeps crashing
  when opened so added step to clear data on setup.
* Chrome clogs logcat with errors whilst running speedometer so delayed writing
  scores to logcat until after tabs have been closed.
* Tested on pixel3, mate20, s10-exynos, pixel4, s9-exynos, s9-snapdragon, xperia.


robF8:
It might be worth leaving it as UiautoWorkload and adding browser parameter so it can be extended to use others. What do you think @marcbonnici ?

wa/workloads/speedometer/__init__.py
```diff
@@ -48,14 +49,11 @@ class Speedometer(UiautoWorkload):
     def __init__(self, target, **kwargs):
         super(Speedometer, self).__init__(target, **kwargs)
         self.gui.timeout = 1500
-        self.gui.uiauto_params['version'] = self.version
+        self.gui.uiauto_params['version'] = self.speedometer_version
 
     def setup(self, context):
+        self.target.execute('pm clear com.android.chrome', check_exit_code=False)
```
marcbonnici: Now this workload is associated with the chrome application this should now be done automatically by the framework.

robF8: Fixed in latest update. Thanks Marc

PR #1081, opened 2020-04-01
Fix device sleep issue for jankbench workload

robF8:
* Sleeping device when waiting for workload to end was causing workload
  to occasionally freeze. Changed device sleep to host sleep.
* Panda dataframe.ix() is deprecated so changed to .loc()

### DIFF
--- a/wa/workloads/jankbench/__init__.py
+++ b/wa/workloads/jankbench/__init__.py
@@ -24,6 +24,7 @@ import threading
 import sqlite3
 import subprocess
 import sys
+import time
 from copy import copy
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


setrofim:
The sleep on target is quite deliberate -- this allows the workload to be used with simulations, where on-target time might proceed at very different rate from wall-clock time. Could you provided a bit more detail regarding the freezes you're encountering when waiting on target? E.g. is it target specific? does the workload actually freeze, or is it running but the sleep not returning? or does the entire target crash for some reason? etc.

robF8:
It happens on quite a few of the devices, reproducible 100% on a mate20 and pixel 4. I thought it was the sleep not returning properly but now not so sure. The workload just seems to hang after stop_event.set() and self.join() is called in JankbenchRunMonitor::stop() and the logcat stream process isn't being terminated.

marcbonnici:
A few things to check: 
- Is there anything abnormal in the logcat for the device that may indicate the problem? 
- Is this a locally connected device? 
- Are you able to set a breakpoint to determine where the code is getting stuck?

PR #1085, opened 2020-04-09
Create app launch workload

robF8:
* Created applaunch workload that is not dependent on app launch interface.

### DIFF
--- a/wa/workloads/adb_applaunch/__init__.py
+++ b/wa/workloads/adb_applaunch/__init__.py
@@ -1,0 +1,197 @@
+import os
+import re
+import time
+
+from wa import Parameter, Workload, WorkloadError, File
+from wa.utils.types import list_or_string
+
+LAUNCH_TIME_TYPE_DISPLAYED = 'displayed'
+LAUNCH_TIME_TYPE_FULLY_DRAWN = 'fully_drawn'
+
+COMMON_APPS = {'camera' : 'com.google.android.GoogleCamera',
+               'reddit' : 'com.reddit.frontpage',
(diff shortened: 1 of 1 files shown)


marcbonnici:
Could you please explain what the benefit and differences are of this workload over the existing applaunch workload?

robF8:
This workload allows launching any app as opposed to the existing workload which only allows you to launch apps which already have a workload that implements ApplaunchInterface. It also allows for launching specific activities of any app and measuring start up time.

marcbonnici:
The aim behind the `ApplaunchInterface` was to provide a clean way of adding support for measuring the time taken to launch any workload that implemented the interface and check it has loaded correctly.
What sort of measurements etc. do you get from this workload compared to the other?

robF8:
They measure the same thing. This workload checks whether the app has launched correctly by making sure Displayed or Report fully drawn have been logged by the Activity Manager. It's intended to be a more lightweight version of the existing applaunch workload, for example if you wanted to gather app launch data for 50 different apps, you could do so without having to produce and maintain 50 different workloads. I realise that the two workloads are fairly similar however I didn't want to replace the existing applaunch workload since it allows for more control in setting up activities etc

setrofim:
>  realise that the two workloads are fairly similar however I didn't want to replace the existing applaunch workload since it allows for more control in setting up activities etc

Why not add the features of this workload to existing one, possibly configurable as an option? I don't want to bloat the workload list with nearly identical entries so that an unfamiliar user won't know how to chose between them. Plus this unnecessarily duplicates functionality and adds to our maintenance burden.

PR #1088, opened 2020-04-24
Update gfxbench to log correct scores.

robF8:
* Updated regex to reflect correct test name.

* Enabling/disabling tests on setup was missing tesselation on some devices
  so changed order of toggle.

* Sometimes whilst collecting scores the workload grabs the wrong score.
  Updated to check the name of the test before grabbing the score.

* Tested on mate20, xperia, s9-exynos, s10-exynos, pixel-4


wa/workloads/gfxbench/__init__.py
```diff
@@ -51,7 +51,7 @@ def update_output(self, context):
         super(Gfxbench, self).update_output(context)
         expected_results = len(self.regex_matches)
         logcat_file = context.get_artifact_path('logcat')
-        with open(logcat_file, errors='replace') as fh:
+        with open(logcat_file, encoding="utf8", errors='ignore') as fh:
```
marcbonnici, line 54: Was this still causing issues with the xpereia device? Should this be updated across other workloads?

robF8: Reverted this line, I had authored this commit before the logcat parsing change and made mistake in the rebase, sorry about that.

PR #1100, opened 2020-06-03
Implement caching of local ApkInfo Objects 

marcbonnici:
To prevent the requirement of locking and parsing an APK file multiple times cache the information using its path and last modified time. This should improve execution times when using large APK files and reduce contention when running with multiple instances of WA. 

Dependent on https://github.com/ARM-software/devlib/pull/488


setrofim:
Looks good. My only concern with this is that, as cache grows, the time it takes to parse and serialize my match or even exceed the comparable time for parsing info from an Apk. I would add a persistent Cache object that would maintain the parsed data and only re-read the cache file as necessary (i.e. when it has been updated); all cache queries would then go via that.

PR #1109, opened 2020-06-23
Target Manager Parameter Updates

marcbonnici:
- Fix the implementation of Target Manager parameters
- Enable the configuration of the target_info cache
- Fix typo


wa/framework/target/manager.py
```diff
@@ -135,6 +132,11 @@ def verify_target_responsive(self, context):
                 raise TargetNotRespondingError('Target unresponsive and hard reset not supported; bailing.')
 
     def _init_target(self):
+        # Import here to prevent circular import
+        # pylint: disable=wrong-import-position,cyclic-import
+        from wa.framework.target.descriptor import (get_target_description,
```
setrofim, line 146: What's the cycle here?

marcbonnici: Looking at this again, I think I moved the wrong import and should have moved the new `TargetManger` import in the descriptor to be more obvious.

But essentially the cycle here is that the `TargetManager` needs access to the target descriptor methods as before, however this PR also adds the requirement for the target descriptor to access the `TargetManager`s parameters so it can add them to the target descriptor for filtering.

setrofim: >however this PR also adds the requirement for the target descriptor to access the TargetManagers parameters 

But those are accessed via an argument passed into `instantiate_target` when called. I don't see the load-time  `descriptor.py` --> `manager` dependency. Am I missing something?

setrofim: oh, wait, never mind; just saw the update.

setrofim: OK, so issue here is that `TargetManager` parameters are being populated inside `TargetDescription.get_descriptions()`. This breaks conceptual integrity -- `TargetDescription` should not know about `TargetManager`; it only describes a target.

Rather than doing this, I would pass the manager parameters as arguments to `list_target_descriptions` and add them to each description there.

PR #1115, opened 2020-06-30
workloads/gfxbench: Swtich to the home screen before run

marcbonnici:
Alter the element to check popups have closed to be present
on any main screen and ensure we switch to the homescreen of the
app before performing setup steps.


wa/workloads/gfxbench/uiauto/app/src/main/java/com/arm/wa/uiauto/gfxbench/UiAutomation.java
```diff
@@ -122,7 +128,7 @@ public void clearFirstRun() throws Exception {
         }
 
         UiObject home =
-            mDevice.findObject(new UiSelector().resourceId(packageID + "main_homeBack")
+            mDevice.findObject(new UiSelector().resourceId(packageID + "main_view_back")
```
setrofim, line 131: Is this version dependent? And if so, should we be checking for both?

marcbonnici: I don't think so no. This appears to be a background element that is present only the homescreen which has been updated to now be a generic background element of the application which is shared by any of the main screens which is used to detect when any popups have finished closing.

PR #1116, opened 2020-07-02
WA/Jankbench: Update Pandas function to remove deprecated .ix access

credp:
Pandas removed .ix as a way to iterate the index, .loc is the replacement
in most cases. Jankbench as a workload fails on a clean install due to
this call.

Replacing this works for me on a native install of Lisa with Ubuntu 20.04

Signed-off-by: Chris Redpath <chris.redpath@arm.com>

### DIFF
--- a/wa/workloads/jankbench/__init__.py
+++ b/wa/workloads/jankbench/__init__.py
@@ -145,7 +145,7 @@ class Jankbench(ApkWorkload):
             results.columns = ['mean', 'std_dev', 'count_jank', 'jank_p']
 
             for test_name, rep in results.index:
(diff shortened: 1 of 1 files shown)


credp:
Thanls Marc, I think this is done but I'm not 100% sure which Pandas version deprecated the .ix and if .loc is always present.

credp:
It looks like .loc has been present since at least 0.11.0, and .ix was deprecated in 0.20.0 and eventually removed. I've got pandas 1.0.5 on my machine.

marcbonnici:
Thanks for checking the version numbers, WA requires pandas >0.23 so there shouldn't be any issues making the switch.

PR #1117, opened 2020-07-06
framework, tests, commands: Update state tracking

jrpdotjpg:
- Test benches provided to track the serialisation of execution state to
the state file, and for the state of a retried job
 
  - These were necessary to show that
         a) Retried jobs did not have the correct status
         b) Job status was not reflected promptly in the state file

- Serialisation was coupled with setting job status to prevent this,
and JobState was consolidated inside Job to aid status consistency.

- Added runstatus command for the monitoring of ongoing and
completed WA runs. 

- Fixed signal disconnection and added testcase

### DIFF
--- a/tests/test_execution.py
+++ b/tests/test_execution.py
@@ -1,0 +1,225 @@
+#    Copyright 2020 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import tempfile
+from unittest import TestCase
+
+from mock.mock import Mock
+from nose.tools import assert_equal
+from datetime import datetime
+
+from wa.framework.configuration import RunConfiguration
+from wa.framework.configuration.core import JobSpec, Status
+from wa.framework.execution import ExecutionContext, Runner
+from wa.framework.job import Job
+from wa.framework.output import RunOutput, init_run_output
+from wa.framework.output_processor import ProcessorManager
+import wa.framework.signal as signal
+from wa.framework.run import JobState
+
+
+class MockConfigManager(Mock):
+
+    @property
+    def jobs(self):
+        return self._joblist
+
+    @property
+    def loaded_config_sources(self):
+        return []
+
+    @property
+    def run_config(self):
+        return RunConfiguration()
+
+    @property
+    def plugin_cache(self):
+        return MockPluginCache()
+
+    def __init__(self, *args, **kwargs):
+        super(MockConfigManager, self).__init__(*args, **kwargs)
+        self._joblist = None
+        self._run_config = RunConfiguration()
+
+    def to_pod(self):
+        return {}
+
+
+class MockPluginCache(Mock):
+
+    def list_plugins(self, kind=None):
+        return []
+
+
+class MockProcessorManager(Mock):
+
+    def __init__(self, *args, **kwargs):
+        super(MockProcessorManager, self).__init__(*args, **kwargs)
+
+    def get_enabled(self):
+        return []
+
+
+class JobState_force_retry(JobState):
+
+    @property
+    def status(self):
+        return self._status
+
+    @status.setter
+    def status(self, value):
+        if(self.retries != self.times_to_retry) and (value == Status.RUNNING):
+            self._status = Status.FAILED
+            if self.output:
+                self.output.status = Status.FAILED
+        else:
+            self._status = value
+            if self.output:
+                self.output.status = value
+
+    def __init__(self, to_retry, *args, **kwargs):
+        self.retries = 0
+        self._status = Status.NEW
+        self.times_to_retry = to_retry
+        self.output = None
+        super(JobState_force_retry, self).__init__(*args, **kwargs)
+
+
+class Job_force_retry(Job):
+    '''This class imitates a job that retries as many times as specified by
+    ``retries`` in its constructor'''
+
+    def __init__(self, to_retry, *args, **kwargs):
+        super(Job_force_retry, self).__init__(*args, **kwargs)
+        self.state = JobState_force_retry(to_retry, self.id, self.label, self.iteration, Status.NEW)
+
+
+class TestRunState(TestCase):
+
+    def setUp(self):
+        self.path = tempfile.mkstemp()[1]
+        os.remove(self.path)
+        self.initialise_signals()
+
+        config = MockConfigManager()
+        output = init_run_output(self.path, config)
+
+        self.context = ExecutionContext(config, Mock(), output)
+
+        self.job_spec = JobSpec()
+        self.job_spec.augmentations = {}
+        self.job_spec.finalize()
+
+    def tearDown(self):
+        signal.disconnect(self._verify_serialized_state, signal.RUN_INITIALIZED)
+        signal.disconnect(self._verify_serialized_state, signal.JOB_STARTED)
+        signal.disconnect(self._verify_serialized_state, signal.JOB_RESTARTED)
+        signal.disconnect(self._verify_serialized_state, signal.JOB_COMPLETED)
+        signal.disconnect(self._verify_serialized_state, signal.JOB_FAILED)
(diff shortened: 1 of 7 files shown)


wa/commands/runstatus.py
```diff
@@ -0,0 +1,298 @@
+import os
+import logging
+from datetime import datetime, timedelta
+
+import colorama
+
+from wa.framework.configuration.core import Status
```
setrofim, line 7: These should be in alphabetical order. In `nvim`:
1. `<shift>-v` to enter line-select mode.
2. Move down to select all import lines below, then
3. `:sort<cr>`

tests/test_execution.py
```diff
@@ -0,0 +1,222 @@
...
+    def test_job_state_transitions_retry(self):
+        #Tests state equality when job fails initially
+        job = Job_force_retry(1, self.job_spec,1, self.context)
+        job.workload = Mock()
+
+        self.context.cm._joblist=[job]
+        self.context.run_state.add_job(job)
+
+        runner = Runner(self.context, MockProcessorManager())
+        runner.run()
```
marcbonnici, line 162: This test is failing for me and travis.

wa/commands/runstatus.py
```diff
@@ -0,0 +1,298 @@
...
+from wa.framework.exception import NotFoundError
+from wa.utils.terminalsize import get_terminal_size
+from wa.utils.doc import format_column
+from wa.utils.log import COLOR_MAP, RESET_COLOR
+
+class StatusCommand(Command):
+
+    name = 'runstatus'
+    description = '''
+    Monitor ongoing runs and provide information on their progress.
```
setrofim, line 19: Add a more detailed description below, e.g. explaining exactly what it does, and the format in which the output appears. Think of this as the `man` page for this command.

wa/commands/runstatus.py
```diff
@@ -0,0 +1,298 @@
...
+from wa import Command, settings
+from wa.framework.output import RunOutput, discover_wa_outputs
+from wa.framework.exception import NotFoundError
+from wa.utils.terminalsize import get_terminal_size
+from wa.utils.doc import format_column
+from wa.utils.log import COLOR_MAP, RESET_COLOR
+
+class StatusCommand(Command):
+
+    name = 'runstatus'
```
setrofim, line 17: Rename to just `status` (the same goes for file name).

wa/commands/runstatus.py
```diff
@@ -0,0 +1,298 @@
...
+                        continue
+                    else:
+                        break
+
+                run_output = possible_outputs[select]
+
+            else:
+                run_output = possible_outputs[0]
+
+        rs = RunMonitor(run_output)
```
setrofim, line 69: Why `rs`? Wouldn't `rm` (or, even better, `monitor`) make more sense?

wa/commands/runstatus.py
```diff
@@ -0,0 +1,298 @@
...
+        rc = self.ro._combined_config.run_config
+        self.segmented = segment_jobs_by_state(self.jobs,
+                                                rc.max_retries,
+                                                rc.retry_on_status
+                                                )
+
+    def generate_run_header(self):
+        info = self.ro.info
+        
+        header = ('\n=========================\n'
```
setrofim, line 118: Better to use multi-line string for this (using triple quotes` -- That will remove the need for `\n`'s

wa/commands/runstatus.py
marcbonnici, line 19: This should probably be phrased in the singular form as well since we currently only support one run at a time.

wa/commands/runstatus.py
```diff
@@ -0,0 +1,298 @@
...
+        summary = ('\n=========================\n'
+                   'Job Summary'
+                   '\n=========================\n\n'
+                   )
+        summary += f'Total: {total}, Completed: {num_fin} ({(num_fin/total)*100}%)\n'
+
+        job_list = []
+        for run_state, jobs in ((k, v) for k, v in self.segmented.items() if v):
+            if run_state == 'finished':
+                end_states = {
```
setrofim, line 161: Look into https://docs.python.org/3/library/collections.html#collections.Counter

wa/commands/runstatus.py
```diff
@@ -0,0 +1,298 @@
...
+        return summary
+
+    def generate_job_detail(self):
+        detail = ('\n=========================\n'
+                  'Job Detail'
+                  '\n=========================\n\n'
+                  )
+
+        for segment, jobstates in self.segmented.items():
+            if jobstates:
```
setrofim, line 195: I would switch this up to 
```
if not jobstates:
   continue
for job in jobstates:
   # ...
```
This will allow you to remove an indent level. Generally, the fewer indent levels there are in the code, the better.

wa/commands/runstatus.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def generate_job_summary(self):
+        total = len(self.jobs)
+        num_fin = len(self.segmented['finished'])
+
+        summary = ('\n=========================\n'
+                   'Job Summary'
+                   '\n=========================\n\n'
+                   )
+        summary += f'Total: {total}, Completed: {num_fin} ({(num_fin/total)*100}%)\n'
```
marcbonnici, line 156: This should check if the total number of runs is valid. In cases where the jobs were not generated correctly this results in a divide by 0 error.

wa/commands/runstatus.py
```diff
@@ -0,0 +1,298 @@
...
+        else:
+            segmented['other'].append(jobstate)
+
+    return segmented
+
+
+class _simple_formatter:
+    color_map = {
+        'running': COLOR_MAP[logging.INFO],
+        'ok': colorama.Fore.WHITE,
```
setrofim, line 273: Ideally, this should be the default foreground color for the user's terminal, rather than explicitly white. IIRC colorama should have an alias for that? If not, I think jus using `None` and avoiding formatting to  begin with, should be way to go. Do not assume that people use white as their foreground color (e.g. this will work terribly on terminals with white bacground and dark foreground).

wa/commands/runstatus.py
marcbonnici: The problem with using `status` is that we already have an output processor with that name which causes a clash.

wa/commands/runstatus.py
marcbonnici: Or alternatively may it could make use of the `underline` method?

wa/commands/runstatus.py
jrpdotjpg: Ah, old name for the class was RunState before I realise how bad that was.

tests/test_execution.py
jrpdotjpg: Should be all passing now

setrofim:
> The problem with using status is that we already have an output processor with that name which causes a clash.

Ah, right. OK, how about `instpect`  instead? I don't want to limit this command to "run" as we might want to overload it later for other things (e.g. like we do with create). Besides, thinking about it, it should really be a verb in the first place.

wa/commands/runstatus.py
setrofim: Ah, right. OK, how about instpect instead? I don't want to limit this command to "run" as we might want to overload it later for other things (e.g. like we do with create). Besides, thinking about it, it should really be a verb in the first place.

marcbonnici: That makes sense, and seems like a good option. The only other one I could think of was 'report' but happy to go with 'inspect'.

setrofim: Yours is better. Let's go with `report`.

tests/test_execution.py
```diff
@@ -0,0 +1,213 @@
...
+        job = Job(self.job_spec, 1, self.context)
+        job.workload = Mock()
+
+        self.context.cm._joblist = [job]
+        self.context.run_state.add_job(job)
+
+        runner = Runner(self.context, MockProcessorManager())
+        runner.run()
+
+    def test_job_state_transitions_fail(self):
```
marcbonnici, line 134: Now this test seems to be failing instead.
(https://travis-ci.org/github/ARM-software/workload-automation/jobs/705788405)

wa/commands/report.py
```diff
@@ -0,0 +1,288 @@
...
+            header += "Project stage: {}\n".format(info.project_stage)
+
+        if info.start_time:
+            duration = _seconds_as_smh(self.elapsed_time.total_seconds())
+            header += ("Start time: {}\n"
+                       "Duration: {:02}:{:02}:{:02}\n"
+                       ).format(info.start_time,
+                                duration[2], duration[1], duration[0],
+                                )
+            if len(self.segmented['finished']) > 0 and not info.end_time:
```
marcbonnici, line 157: C:157,15: Do not use `len(SEQUENCE)` to determine if a sequence is empty (len-as-condition)
https://travis-ci.org/github/ARM-software/workload-automation/jobs/706972953

PR #1120, opened 2020-07-10
WIP: Code Optimisations 

marcbonnici:
- To prevent the requirement of parsing an APK file multiple times cache the information using its path and last modified time and keep a copy in memory. This should improve execution times when using large APK files and reduce contention when running with multiple instances of WA.
- To prevent long timeouts occurring during to file locking on both reads and writes implement and replace locking with atomic writes. While this may results in cache entries being overwritten, the amount of time used in duplicated retrievals will likely be saved with the prevention of stalls due to waiting to acquire the file lock.
- Generate the md5 hash of assets in chunks to prevent loading large files into memory at once.

Supersedes #1100

### DIFF
--- a/wa/utils/android.py
+++ b/wa/utils/android.py
@@ -14,9 +14,16 @@
 #
 
 import logging
+import os
 from datetime import datetime
 
+from devlib.utils.android import ApkInfo as _ApkInfo
+
+from wa import settings
+from wa.framework.exception import ConfigError
+from wa.utils.serializer import read_pod, write_pod, Podable
 from wa.utils.types import enum
+from wa.utils.misc import lock_file
 
 
 LogcatLogLevel = enum(['verbose', 'debug', 'info', 'warn', 'error', 'assert'], start=2)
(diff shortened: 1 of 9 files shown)


wa/utils/android.py
```diff
@@ -132,3 +135,68 @@ def _pod_upgrade_v1(pod):
...
+        self._update_cache()
+        pod = self.cache.get(key)
+
+        info = ApkInfo.from_pod(pod) if pod else None
+        return info
+
+    def _update_cache(self):
+        if not self._initialised:
+            return
+        if self.last_modified != os.stat(self.path):
```
setrofim, line 174: Can the `os.stat` not result in a `FileNotFoundError` being raised when an `ApkInfoCache` is created in a new environment for the first time, and `store` has not previously been called, so that the cache file does not exist?

marcbonnici: That was meant to have been prevented by the use of `_initialised` however it was not set correctly so have removed and added a check to see if the file exists explicitly.

PR #1122, opened 2020-07-15
workloads/speedometer: offline version of speedometer

stephenkyle-ARM:
This version replaces the previous uiauto version of Speedometer with a new version.

* Supports both chrome and chromium again, this is selected with the
  chrome_package parameter.
* No longer needs internet access.
* Version 1.0 of Speedometer is no longer supported.


wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+
+        logging.info(
+            "Using port {} for temporary webserver".format(self.webserver_port)
+        )
+
+        self.speedometer_url = "http://localhost:{}/Speedometer2.0/index.html".format(self.webserver_port)
+
+        # Host a copy of Speedometer locally
+        subprocess.check_output(
+            "tar -xf {} -C {}".format(self.archive_asset, context.output_directory),
```
setrofim, line 86: use Python's `tarfile` do not assume that `tar` is present on the host system.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+            shell=True,
+        )
+
+        webserver_cmd = "cd {} && python3 -mhttp.server {}".format(
```
setrofim, line 90: Do not not do this. Import `http.server` and start it in a thread.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+            self.document_root, self.webserver_port
+        )
+
+        with open(os.devnull, "w") as dev_null:
```
setrofim, line 94: Don't need to do this, just use `stdout=None`

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+            self.webserver = subprocess.Popen(
+                webserver_cmd, shell=True, stdout=dev_null, stderr=dev_null
+            )
+
+        subprocess.check_output(
+            "adb reverse tcp:{0} tcp:{0}".format(self.webserver_port), shell=True
```
setrofim, line 100: Do not use `adb` directly -- all interactions must go through devlib. Add an appropriate utility there. Or at least used devlib's `adb_command`.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
-    2. Navigate to the speedometer website - http://browserbench.org/Speedometer/
-    3. Execute the benchmark
+    1. Host a local copy of the Speedometer website, and make it visible to the device via ADB.
+    2. Open chrome via an intent to access the local copy.
+    3. Execute the benchmark - the copy has been modified to trigger the start of the benchmark.
 
-    known working chrome version 80.0.3987.149
-    '''
+    known working chrome version 83.0.4103.106
+    """
```
setrofim, line 51: and `supported_platforms=['android']`, as this is android-specific.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+        self.target.execute("pm clear {}".format(self.chrome_package, as_root=True))
+
+        # Launch the browser for the first time...
+        browser_launch_cmd = "am start -a android.intent.action.VIEW -d {} {}".format(
+            self.speedometer_url, self.chrome_package
+        )
+
+        self.target.execute(browser_launch_cmd)
+        time.sleep(1)
+        self.target.execute("am force-stop {}".format(self.chrome_package))
```
setrofim, line 113: Maybe leave a comment to explain why this is needed.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+            ) as write_fh:
+                for line in lines:
+                    write_fh.write(line)
+
+        self.target.push(
+            os.path.join(context.output_directory, self.preferences_xml + ".new"),
+            "/data/data/{}/shared_prefs/{}".format(
+                self.chrome_package, self.preferences_xml
+            ),
+            as_root=True,
```
setrofim, line 155: Since this is now relies on the target being rooted, you must check for that during `initialize` and raise a `WorkloadError`, so that the run aborts early if the requirement is not met.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+            ),
+            context.output_directory,
+            as_root=True,
+        )
+
+        with open(
+            os.path.join(context.output_directory, self.preferences_xml)
+        ) as read_fh:
+            lines = read_fh.readlines()
+            lines.insert(
```
setrofim, line 132: erm, `lines.append`?

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+    def run(self, context):
+        super(Speedometer, self).run(context)
+
+        browser_launch_cmd = "am start -a android.intent.action.VIEW -d {} {}".format(
+            self.speedometer_url, self.chrome_package
+        )
+        self.target.execute(browser_launch_cmd)
+
+        # Wait 60 seconds at least, and then wait until we don't see the 'sandboxed_process' process for 10 seconds.
+        time.sleep(60)
```
setrofim, line 167: Why sixty? You should avoid long sleeps if _at all_ possible. Can you not poll running packages or logcat instead?

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+
+        countdown = 5
+        while countdown > 0:
+            busiest_line = self.target.execute("top -n1 -m1 -q -b").split("\n")[0]
```
setrofim, line 171: Use the `busybox` top instead; do not rely on the top on target being present and accepting those parameters.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+            while "sandboxed_process" in busiest_line:
+                countdown = 5
```
setrofim, line 173: I don't understand the logic of setting the countdown inside the inner loop here. And this inner loop doesn't seem to have a break condition in case "standboxed_process" never goes away. 

This whole nested loop method seems unnecessarily convoluted. Can this be simplified/done via a different method? At the _very_ least,  leave a comment describing wtf is going on here.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
         else:
-            raise WorkloadError("The Speedometer workload has failed. No score was obtainable.")
+            raise WorkloadError(
+                "The Speedometer workload has failed. No score was obtainable."
+            )
+
+    def teardown(self, context):
+        super(Speedometer, self).teardown(context)
+        subprocess.check_output(
+            "adb reverse --remove tcp:{}".format(self.webserver_port), shell=True
```
setrofim, line 208: Again, `adb` should not be invoked directly -- do not assume it is in path. Should be done via `adb_command`  from devlib utils.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+        )
+
+        self.webserver.terminate()
+        # We have to kill the child python3 process independently.
+        webserver_cmdline = ["python3", "-mhttp.server", str(self.webserver_port)]
```
setrofim, line 213: Again, do not invoke another python process via shell -- import the module and run in a thread.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,221 @@
...
+            self.speedometer_url, self.chrome_package
+        )
+
+        self.target.execute(browser_launch_cmd)
+        time.sleep(1)
+        self.target.execute("am force-stop {}".format(self.chrome_package))
+        time.sleep(1)
+
+        # Pull the preferences file from the device, modify it, and push it back.
+        # This is done to bypass the 'first launch' screen of the browser.
```
marcbonnici, line 117: Just a thought, would it have been more robust / remove the root requirement to still use uiautomator for performing this part the setup stage? If I remember correctly as long as it had exited before the run stage started it didn't impact the score?

wa/workloads/speedometer/__init__.py
stephenkyle-ARM: The idea is to reduce the amount of polling we do to the system while it's supposed to be running a performance-sensitive benchmark. Since we inevitably are going to need to poll in some way to work out which it has finished, a compromise was to wait at least 60 seconds - I know Speedometer usually takes around 90 seconds on current devices. Is there something intrinsically wrong with sleeping for this long? What does it affect?

wa/workloads/speedometer/__init__.py
stephenkyle-ARM: I've liberally commented this now. :) If 'sandboxed_process' never stops being the hottest process in the system (something my comments/changes will hopefully make clearer is what I'm looking for), then we should indeed not break out of the inner loop, because the benchmark is still running.

wa/workloads/speedometer/__init__.py
stephenkyle-ARM: As I recall, the 'uiautomator dump' would only actually see the score when executed as root anyway, so I think we need to require root always.

stephenkyle-ARM:
Hopefully that's all changes addressed. 

On the root requirement side, perhaps it is possible to use UIAutomator again to replace the (pm-clear, launch the browser, kill it, modify the shared_prefs file) section for non-rooted devices. Is it easy enough to have UIAutomator be launched in this way just for a portion of the workload? i.e., I imagine it would be more complex than switching back to subclassing UIAutomatorApkWorkload again? On a second look, I noticed that I was now able to use `uiautomator dump` without `as_root` 
- not sure why it hadn't worked in the past. Perhaps this is something that could be investigated later and added back in as a separate PR?

PTAL.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,259 @@
...
+        top_cmd = "{} top -b -n1".format(self.target.busybox)
+
+        def _top_output_for_busiest_process():
+            # busybox's top always outputs the Mem/CPU/Load/Header lines first,
+            # so line 5 is the hottest process.
+            return self.target.execute(top_cmd).split("\n")[4]
+
+        countdown = 5
+        while countdown > 0:
+            while "sandboxed_process" in _top_output_for_busiest_process():
```
marcbonnici, line 216: Testing this in chrome on my nexus 5 the process line is:
` 8845  1438 99021    S <  1651m 90.9   1 17.1 {dboxed_process0} com.android.chrome:sandboxed`
Do you have access to more devices to compare what the process appears as on other devices?

stephenkyle-ARM:
Updated with the following changes:

* no longer relies on old devlib behaviour of preserving ownership of a file that is overwritten with adb push
* addressing Marc's feedback, the workload complains if it is unable to find _any_ process matching the search term it uses to find the active render process. This term is normally '<chrome package>:sandbox'. A user can configure this search term by setting 'render_process_search_term' in the workload parameters.
* we go back to executing uiautomator dump as root - I was observing it randomly not containing the score in it again (despite it clearly being on the screen).
* make sure we only delete temporary files that exist during teardown, stopping crashes during recovery from a bad run.
* force-stop the browser when the workload is over, to prevent it persisting during the execution of other workloads (the speedometer render process has been seen at times to consume at least 3% CPU when another browser workload is running, for example)

PTAL!

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,295 @@
...
+        render_processes = [x for x in _get_all_processes() if search_term in x]
+        if len(render_processes) == 0:
+            raise WorkloadError(
+                "Unable to find any suitable render process. Try setting "
+                "'render_process_search_term'."
+            )
+
+        countdown = 5
+        while countdown > 0:
+            while "sandboxed_process" in _top_output_for_busiest_process():
```
marcbonnici, line 247: This doesn't appear to use the `search_term` you've checked for?

stephenkyle-ARM: Good catch! :) fixed.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,295 @@
...
+
+        self.speedometer_url = "http://localhost:{}/Speedometer2.0/index.html".format(
+            self.webserver_port
+        )
+
+        # The rest of this function is in service of making sure we start with
+        # a 'fresh' browser - no other tabs, nothing in the page cache, etc.
+
+        # Clear the application's cache.
+        self.target.execute("pm clear {}".format(self.chrome_package, as_root=True))
```
marcbonnici, line 147: `as_root` should be outside of the `format` call.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,295 @@
...
+            This directory is captured in |document_root|. It also suppresses
+            logging."""
+
+            def translate_path(self, path):
+                path = SimpleHTTPRequestHandler.translate_path(self, path)
+                requested_uri = os.path.relpath(path, os.getcwd())
+                return os.path.join(document_root, requested_uri)
+
+            # Disable the logging.
+            def log_message(self, format, *args):
```
marcbonnici, line 47: pylint nit: Can you add a `# pylint: disable=redefined-builtin` to ignore the dummy `format` argument?

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,295 @@
...
+            return self.target.execute(top_cmd).split("\n")[4]
+
+        def _get_all_processes():
+            return self.target.execute(top_cmd).split("\n")[4:]
+
+        search_term = self.chrome_package + ":sandbox"
+        if self.render_process_search_term is not None:
+            search_term = self.render_process_search_term
+        render_processes = [x for x in _get_all_processes() if search_term in x]
+        if len(render_processes) == 0:
```
marcbonnici, line 239: pylint nit: "Do not use `len(SEQUENCE)` to determine if a sequence is empty (len-as-condition)"

stephenkyle-ARM: I think this is actually a bug in the version of pylint you're using. It is designed to catch use of `len(X)` as the sole condition in an if statement, when `len(X) == 0` is perfectly OK, and in my opinion, more readable as its intent is more explicit. This particular version of pylint (assuming you're using 1.9.2) rejects this line, but later versions do not - I didn't see this initially when I installed the latest version of pylint through pip (2.5.3). See https://github.com/PyCQA/pylint/issues/2684 for the issue which led to pylint being fixed in light of this.

marcbonnici: Hmm I see. Yes you're correct, we're still using v1.9.2 as at the time we hit a bug in the later version but we could probably do with an upgrade by now.

wa/workloads/speedometer/__init__.py
setrofim: Long sleeps introduce unnecessary delays, this can end up significantly increasing the over all time. Yo're also making assumptions about how long things take which may not be valid when run on a system that is vastly different from the system on which the workload is first implemented. Fixed waits are brittle. 

This is why, as a rule, we generally prefer some form of polling or signalling guarded by generous timeouts.   A slow low-overhead poll (e.g. with cadence of 1-2s) will typically not impact most workloads significantly enough to affect results -- given the level of backgrounds activity on a typical rich-OS system (especially Android), it's usually in the noise.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,298 @@
...
-    regex = re.compile(r'Speedometer Score ([\d.]+)')
-    versions = ['1.0', '2.0']
-    description = '''
-    A workload to execute the speedometer web based benchmark
+class ArchiveServer:
+    def __init__(self):
+        self._port = None
+
+    def start(self, document_root, target):
+        class _DifferentDirectoryHTTPRequestHandler(SimpleHTTPRequestHandler):
```
setrofim, line 35: Is there an issue with defining these classes at the top level? Function-local definitions should be avoided in Python -- they case problems.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,298 @@
...
 
-class Speedometer(ApkUiautoWorkload):
 
-    name = 'speedometer'
-    package_names = ['com.android.chrome']
-    regex = re.compile(r'Speedometer Score ([\d.]+)')
-    versions = ['1.0', '2.0']
-    description = '''
-    A workload to execute the speedometer web based benchmark
+class ArchiveServer:
```
setrofim, line 30: Move this below the Workload definition.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,298 @@
...
+            "chrome_package",
+            allowed_values=package_names,
+            kind=str,
+            default="com.android.chrome",
+            description="""
+                  The app package for the browser that will be launched.
+                  """,
+        ),
+        Parameter(
+            "render_process_search_term",
```
setrofim, line 109: This seems to be a cop-out, and the whole process seems very brittle...

Since you are modifying the Speedometer code anyway, could you not also get it to signal its own completion, e.g. by creating a file, or opening an alert dialog that you can then look for with UIAutomatior?

wa/workloads/speedometer/__init__.py
setrofim: oh my....

stephenkyle-ARM:
Rewritten to check for a value written to the browser's LocalStorage, which we can find on the file system as root. This value indicates the benchmark is complete.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,363 @@
...
     ]
 
-    requires_network = True
-
     def __init__(self, target, **kwargs):
         super(Speedometer, self).__init__(target, **kwargs)
-        self.gui.timeout = 1500
-        self.gui.uiauto_params['version'] = self.speedometer_version
+        self.archive_server = ArchiveServer()
+        if not target.is_rooted:
```
setrofim, line 101: This should be done inside `initialize()` not `__init__` -- do not rely on target connection existing when the workload object is instantiated.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,363 @@
...
+            raise WorkloadError(
+                "Device must be rooted for the speedometer workload currently"
+            )
+
+    def setup(self, context):
+        super(Speedometer, self).setup(context)
+        self.document_root = os.path.join(context.output_directory, "document_root")
+
+        # Host a copy of Speedometer locally
+        tarball = os.path.join(os.path.dirname(__file__), "speedometer_archive.tgz")
```
setrofim, line 111: Use WA's resource discovery framework rather than hard-coding the location. This should also be done inside `intialize()` rather than `setup`() as  it only needs to be done once per WA run, not for each iteration of the workload.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,363 @@
...
+            os.path.join(context.output_directory, self.preferences_xml + ".new"),
+        ]
+        for file_to_delete in files_to_delete:
+            if os.path.exists(file_to_delete):
+                subprocess.check_output("rm -r {}".format(file_to_delete), shell=True)
+
+        # The browser's processes can stick around and have minor impact on
+        # other performance sensitive workloads, so make sure we clean up.
+        self.target.execute("am force-stop {}".format(self.chrome_package))
+
```
setrofim, line 319: Add a `finalize()` to clean up the workload files deployed to the target if `self.cleanup_assets` is `True`

stephenkyle-ARM:
Okay, changes addressed, PTAL.

stephenkyle-ARM:
Friendly request for another review? :) @marcbonnici @setrofim ?

setrofim (approved):
Sorry for the delay in looking at this.

marcbonnici (changes requested):

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,365 @@
...
+                score_read = True
+            else:
+                if iterations >= 10:
+                    raise WorkloadError(
+                        "The Speedometer workload has failed. No score was obtainable."
+                    )
+                else:
+                    # Sleep and retry...
+                    time.sleep(2)
+                    iterations += 1
```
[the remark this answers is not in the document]
stephenkyle-ARM: Yep, from testing a number of devices in our farm, the uiautomator dump command seems to fail to capture the score in the first one or two runs on a number of occasions - it doesn't seem to be fixed by running as root, and it varied from device to device. Adding this iterative check helped. Perhaps we only need 5 iterations, but taking 20s didn't seem too bad to make an attempt to read the score correctly before we give up.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,365 @@
...
+        self._httpd.document_root = document_root
+        _, self._port = self._httpd.server_address
+
+        self._thread = ArchiveServerThread(self._httpd)
+        self._thread.start()
+
+        adb_command(target.adb_name, "reverse tcp:{0} tcp:{0}".format(self._port))
+
+    def stop(self, target):
+        adb_command(target.adb_name, "reverse --remove tcp:{}".format(self._port))
```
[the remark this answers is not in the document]
stephenkyle-ARM: As discussed, it seems this is an issue with devices before Android 9 when connected via TCP. Added documentation to note which Android versions are supported (9+ via TCP, 5+ via USB).

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,365 @@
...
+                    # Each local storage file is in a binary format. Depending on the grep you use, it
+                    # might print out the line '[KEY][VALUE]' or it might just say 'Binary file X
+                    # matches' so handle both just in case.
+                    grep_cmd = '{} grep {} "{}"'.format(
+                        self.target.busybox, report_end_id, ls_file
+                    )
+                    output = self.target.execute(
+                        grep_cmd, as_root=True, check_exit_code=False
+                    )
+                    if "Binary file {} matches" in output or report_end_id in output:
```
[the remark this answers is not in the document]
stephenkyle-ARM: Removed this anyway.

wa/workloads/speedometer/__init__.py
```diff
@@ -12,57 +12,365 @@
...
+    5. Run 'tar -cpzf speedometer_archive.tgz document_root'
+
+    6. Copy the tarball into the workloads/speedometer directory
+
+    7. If appropriate, update the commit info in the LICENSE file.
+    """
+    supported_platforms = ["android"]
+
+    package_names = ["org.chromium.chrome", "com.android.chrome"]
+    regex = re.compile('text="(\d+.\d+)" resource-id="result-number"')
```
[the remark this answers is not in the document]
stephenkyle-ARM: I have created a new regex that will capture text or content-desc, and hopefully if the attribute orderings were to change randomly it should be protected against that too!

wa/workloads/speedometer/__init__.py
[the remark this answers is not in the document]
stephenkyle-ARM: What is `adb_server` set to? Sorry I'm not familiar with the usage here - is that something that can be found in `self`, i.e. `adb_server=self.adb_server` if you've set an ADB server in your config?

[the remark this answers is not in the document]
marcbonnici: Ah sorry, it would set similar to they way you retrieve the `adb_name` ie. `adb_server=target.adb_server` as it would be a part of the supplied device config.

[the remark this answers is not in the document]
stephenkyle-ARM: As discussed elsewhere, this isn't going to work without some kind of extra port forwarding being setup between the machine running WA and the machine running the remote ADB server - so we detect if adb_server is non-None in target, and refuse to run the workload in this case, saying it isn't supported.

marcbonnici (commented):
Can you clean up the indentation to keep pep8 happy?
https://travis-ci.org/github/ARM-software/workload-automation/jobs/726261564#L3003

marcbonnici (approved):

Issue #1106, opened 2020-06-17
Provide runtime configuration for keeping a device awake.


marcbonnici:
When running long workloads the maximum screen timeout is sometimes not long enough and the device screen can switch off.

It would be worth investigating https://developer.android.com/reference/android/provider/Settings.Global#STAY_ON_WHILE_PLUGGED_IN to prevent the display from switching off.

marcbonnici:
Implemented with https://github.com/ARM-software/workload-automation/pull/1125

PR #1125, opened 2020-08-11
target/runtime_config: Add support for stay-on

jrpdotjpg:
Adds runtime config support for the android setting
``stay_on_while_plugged_in``.

Uses feature introduced [here](https://github.com/ARM-software/devlib/pull/505)
[Github Issue](https://github.com/ARM-software/workload-automation/issues/1106)

The pull request title is misleading, as since the initial
request the stay-on parameter is **no longer** a runtime
configuration, and is now a device configuration.

### DIFF
--- a/wa/framework/target/assistant.py
+++ b/wa/framework/target/assistant.py
@@ -35,6 +35,9 @@ class LinuxAssistant(object):
     def __init__(self, target):
         self.target = target
 
+    def initialize(self):
+        pass
+
     def start(self):
         pass
 
@@ -42,6 +45,9 @@ class LinuxAssistant(object):
         pass
 
     def stop(self):
+        pass
+
+    def finalize(self):
         pass
 
 
@@ -68,12 +74,29 @@ class AndroidAssistant(object):
                   temporary locaiton on the host. Setting the value of the poll
                   period enables this behavior.
                   """),
+        Parameter('stay_on_mode', kind=int,
+                  constraint=lambda x: 0 <= x <= 7,
+                  description="""
+                  Specify whether the screen should stay on while the device is
+                  charging:
+
+                    0: never stay on
+                    1: with AC charger
+                    2: with USB charger
+                    4: with wireless charger
+
+                  Values can be OR-ed together to produce combinations, for
+                  instance ``7`` will cause the screen to stay on when charging
+                  under any method.
+                  """),
     ]
 
-    def __init__(self, target, logcat_poll_period=None, disable_selinux=True):
+    def __init__(self, target, logcat_poll_period=None, disable_selinux=True, stay_on_mode=None):
         self.target = target
         self.logcat_poll_period = logcat_poll_period
         self.disable_selinux = disable_selinux
+        self.stay_on_mode = stay_on_mode
+        self.orig_stay_on_mode = self.target.get_stay_on_mode() if stay_on_mode is not None else None
         self.logcat_poller = None
         self.logger = logging.getLogger('logcat')
         self._logcat_marker_msg = None
@@ -82,8 +105,11 @@ class AndroidAssistant(object):
         if self.logcat_poll_period:
             signal.connect(self._after_workload, signal.AFTER_WORKLOAD_EXECUTION)
 
+    def initialize(self):
         if self.target.is_rooted and self.disable_selinux:
             self.do_disable_selinux()
+        if self.stay_on_mode is not None:
+            self.target.set_stay_on_mode(self.stay_on_mode)
 
     def start(self):
         if self.logcat_poll_period:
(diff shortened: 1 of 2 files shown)


wa/framework/target/runtime_config.py
```diff
@@ -930,6 +935,19 @@ def initialize(self):
...
+            param_name = 'stay_on'
+            self._runtime_params[param_name] = \
+                RuntimeParameter(
+                    param_name, kind=int,
+                    constraint=lambda x: 0 <= x <= 7,
+                    default=0,
+                    setter=self.set_stay_on_mode,
+                    description="""
+                    Specify whether the screen should stay on while
+                    the device is charging
```
setrofim, line 947: Specify how? The meaning of values `0-7` should be documented here.

setrofim:
Thinking about it, why is this a runtime parameter? This seems to configure the overall device behaviour, and should probably be a device parameter instead.

setrofim:
>I had figured that it was in a similar vein as screen_on, and that it might be preferrable to only have the screen forced on for certain parts of a certain workload, instead of over an entire run. On a more practical side, this might benefit OLED phones for reducing burn-in.

But that's what `screen_on` is for, no? That directly manipulates the state of the device. This seems to be setting behaviour policy.  I think I'm missing the wider context for this change -- where does requirement for this come from?

jrpdotjpg:
You're right, the context of the [original issue](https://github.com/ARM-software/workload-automation/issues/1106) was for the total workload not just parts of it. But `screen_on` only turns the screen on, and doesn't control it thereafter, so provided a method that ensures the screen is on and stays on for the joint purpose. If it was just a device parameter, then the screen might be on for workloads where it needn't be?

setrofim:
Yes, but that's where you would use `screen_on`  runtime parameter.  This is an addition, no an alternative to it. I.e. your agenda uses `screen_on` to turn the screen on/off for workloads as needed. and then this would be an additional device config for devices that have the issue with short screen timeouts.

Also, please edit the original pull request description to add a link to the issue.

jrpdotjpg:
Ah, I understand - augment the existing `screen_on` runtime parameter with `stay_on` as a device config

wa/framework/target/manager.py
```diff
@@ -155,3 +156,4 @@ def _init_target(self):
         self.target.setup()
 
         self.assistant = instantiate_assistant(tdesc, self.parameters, self.target)
+        self.assistant.initialize()
```
setrofim, line 159: I would do this inside `initialize()` rather than `_init_target()`.

setrofim:
Current implementation sets the mode unconditionally, meaning you adding three additional call to target for a feature that is fairly niche. Do not specify a default value, leaving it at `None`, and only get the current mode, set the new one, and reset the original if the user explicitly sets this to an integer value.

This will also avoid unnecessarily causing problems on devices where this  setting isn't implemented, for whatever reason.

wa/framework/target/assistant.py
```diff
@@ -68,12 +74,29 @@ class AndroidAssistant(object):
...
+                  """),
     ]
 
-    def __init__(self, target, logcat_poll_period=None, disable_selinux=True):
+    def __init__(self, target, logcat_poll_period=None, disable_selinux=True, stay_on_mode=None):
         self.target = target
         self.logcat_poll_period = logcat_poll_period
         self.disable_selinux = disable_selinux
+        self.stay_on_mode = stay_on_mode
+        self.orig_stay_on_mode = self.target.get_stay_on_mode() if stay_on_mode else None
```
marcbonnici, line 99: This should probably explicitly check for `None` otherwise it will fail to work for when `stay_on_mode` is set to `0`.

wa/framework/target/assistant.py
```diff
@@ -82,8 +105,11 @@ def __init__(self, target, logcat_poll_period=None, disable_selinux=True):
         if self.logcat_poll_period:
             signal.connect(self._after_workload, signal.AFTER_WORKLOAD_EXECUTION)
 
+    def initialize(self):
         if self.target.is_rooted and self.disable_selinux:
             self.do_disable_selinux()
+        if self.stay_on_mode:
```
marcbonnici, line 111: Ditto

marcbonnici (changes requested):

wa/framework/target/assistant.py
```diff
@@ -98,6 +124,10 @@ def stop(self):
         if self.logcat_poller:
             self.logcat_poller.stop()
 
+    def finalize(self):
+        if self.stay_on_mode:
```
marcbonnici, line 128: `wa/framework/target/assistant.py:128:30: W291 trailing whitespace`

wa/framework/target/assistant.py
jrpdotjpg: Oops, thanks!

marcbonnici (approved):

setrofim (approved):

PR #1128, opened 2020-08-27
PCMark: do not clear on reset

vdonnefort:
When set to true, the data associated with the package will be cleared
(pm clear), before each workload run. The default value is "true" and therefore,
WA behavior remains the same. However, it gives a way of bypassing this
operation which is interesting for a workload such as PCMark, which wouldn't
have to download the benchmark data-set before each run. This download is
time-consuming and pollutes the benchmark instrumentation.

### DIFF
--- a/wa/workloads/pcmark/__init__.py
+++ b/wa/workloads/pcmark/__init__.py
@@ -38,11 +38,14 @@ class PcMark(ApkUiautoWorkload):
     1. Open PCMark application
     2. Swipe right to the Benchmarks screen
     3. Select the Work 2.0 benchmark
-    4. Install the Work 2.0 benchmark
+    4. If needed, install the Work 2.0 benchmark (requires an internet connection)
     5. Execute the Work 2.0 benchmark
 
     Known working APK version: 2.0.3716
     '''
+
+    # Do not delete Work 2.0 data-set before each run
+    clear_data_on_reset = False
 
     def __init__(self, target, **kwargs):
         super(PcMark, self).__init__(target, **kwargs)


setrofim:
Why not just set `clear_data_on_reset` to `False` inside PCMark workload? Why does this need to be configurable by the user?

vdonnefort:
I originally didn't want to change the current behavior. But if you think it is alright to modify it, I can enforce clear_data_on_reset to false only for PCMark, without any option.

setrofim:
Well, you'd only be changing the behaviour for PCMark, and whole point is that it needs it. Since this is tied to the workload, and no the user's setup (i.e. it always makes sense for PCMark), this should be part of the workload's implementation, rather than forcing the user to have to specify the parameter each time.

wa/workloads/pcmark/__init__.py
```diff
@@ -47,6 +47,9 @@ class PcMark(ApkUiautoWorkload):
     def __init__(self, target, **kwargs):
         super(PcMark, self).__init__(target, **kwargs)
         self.gui.timeout = 1500
+        # Do not delete Work 2.0 data-set before each run
+        self.apk.clear_data_on_reset = False
```
marcbonnici, line 51: `self.apk.clear_data_on_reset` shouldn't need to be set directly. `self.clear_data_on_reset` should propagate through to the `PackageHandler`.

vdonnefort: I added a patch to handle `apk.clear_data_on_reset` within `ApkWorkload`, as it wasn't working out of the box. `PackageHandler` being constructed within `ApkWorkload`'s `__init()__.

marcbonnici: Apologies I should have said if you set the `clear_data_on_reset` property of the workload itself (for example in the `gfxbench` workload https://github.com/ARM-software/workload-automation/blob/master/wa/workloads/gfxbench/__init__.py#L25) then when initialising the `PackageHandler` the updated value should be propagated without needing to be manually set.

vdonnefort: Woops, embarrassing mistake...

marcbonnici (approved):

PR #1131, opened 2020-09-16
pep8: Ignore line break before binary operator

marcbonnici:
PEP8 has switched its guidance [1] for where a line break should occur
in relation to a binary operator, so don't raise this warning for
new code.

[1] https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

### DIFF
--- a/dev_scripts/pep8
+++ b/dev_scripts/pep8
@@ -6,7 +6,7 @@ DEFAULT_DIRS=(
 
 EXCLUDE=wa/tests,wa/framework/target/descriptor.py
 EXCLUDE_COMMA=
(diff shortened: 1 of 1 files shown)


setrofim:
> In Python code, it is permissible to break before or after a binary operator, **as long as the convention is consistent locally.** For new code Knuth's style is suggested.

Our existing code base uses the old convention. We should continue to enforce it. Or alternatively, we should update all of the existing code to be in line with the updated guidance, and then enforce the new convention. However, we should not allow both.

marcbonnici:
Good point. Added a todo to look at updating our linters / code base style https://github.com/ARM-software/workload-automation/issues/1132

PR #1139, opened 2020-11-06
fw/descriptor: Update sudo_cmd default

marcbonnici:
The WA default `sudo_cmd` is out of date compared to devlib.
Update the command template to include the latest version.
This enables sudo to read a password from stdin and fixes the password
prompt being included in a command output [1].

[1] https://github.com/ARM-software/devlib/pull/509


setrofim:
Hm, I wonder if could be broken out into a variable inside `devlib` and then reused there to avoid this in the future?

marcbonnici:
Updated to use the variable directly from devlib.

Dependent on https://github.com/ARM-software/devlib/pull/512

PR #1171, opened 2021-04-26
perf: support report-sample

JaviMerino:
devlib learnt to use report-sample in
ARM-software/devlib@fe2fe3ae0470 ("collector/perf: run simpleperf
report-sample in the target if requested").  Adapt the perf instrument
to use the new parameters of the PerfCollector.

### DIFF
--- a/wa/instruments/perf.py
+++ b/wa/instruments/perf.py
@@ -21,7 +21,7 @@ import re
 
 from devlib.collector.perf import PerfCollector
 
-from wa import Instrument, Parameter
+from wa import Instrument, Parameter, ConfigError
 from wa.utils.types import list_or_string, list_of_strs, numeric
 
 PERF_COUNT_REGEX = re.compile(r'^(CPU\d+)?\s*(\d+)\s*(.*?)\s*(\[\s*\d+\.\d+%\s*\])?\s*$')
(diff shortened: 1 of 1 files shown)


wa/instruments/perf.py
```diff
@@ -95,6 +95,11 @@ class PerfInstrument(Instrument):
                   description="""Specifies options to be used to gather report when record command
                   is used. It's highly recommended to use perf_type simpleperf when running on
                   android devices as reporting options are unstable with perf"""),
+        Parameter('run_report_sample', kind=bool, default=False, description="""If true, run
+                  'perf/simpleperf report-sample'. It only works with the record command."""),
```
marcbonnici, line 99: What happens if this is used attempted to be used with `stat`? Is it worth adding a check to warn the user if this is that combination is provided?

JaviMerino: It's ignored if it's used with stat. It's similar to what currently happens if `report_option_string` is used with stat I think. Do you want me to put a warning here (or devlib)?

marcbonnici: Ah I see, I guess devlib might be the better option then. If you don't mind adding it that would be great thanks.

JaviMerino: Sure: https://github.com/ARM-software/devlib/pull/545

setrofim: Configuration shouldn't be ignored. This should result in a `ConfigError` (probably something that needs to be added for `report_option_string` as well...)

JaviMerino: Makes sense. I have added a `validate()` that raises a `ConfigError` if `report_option_string`, `run_report_sample` or `report_sample_options` are specified when the command is not `record`.

marcbonnici (approved):

JaviMerino:
@setrofim Is there anything else that you want me to improve?

setrofim (approved):
LGTM!

JaviMerino:
Great, thanks!

PR #1173, opened 2021-05-26
Update in energy measurement instrument to recover data for failed iteration

deepaliyemul:
Update in teardown of energy measurement instrument to recover data if an iteration fails

### DIFF
--- a/wa/instruments/energy_measurement.py
+++ b/wa/instruments/energy_measurement.py
@@ -30,7 +30,7 @@ from devlib.instrument.acmecape import AcmeCapeInstrument
 from devlib.instrument.monsoon import MonsoonInstrument
 from devlib.platform.arm import JunoEnergyInstrument
 from devlib.utils.misc import which
-
+import logging
 from wa import Instrument, Parameter
 from wa.framework import pluginloader
 from wa.framework.plugin import Plugin
@@ -38,6 +38,7 @@ from wa.framework.exception import ConfigError, InstrumentError
 from wa.utils.types import (list_of_strings, list_of_ints, list_or_string,
                             obj_dict, identifier, list_of_numbers)
 
+logger = logging.getLogger('energy_measurement')
 
 class EnergyInstrumentBackend(Plugin):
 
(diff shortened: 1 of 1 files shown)


marcbonnici:
Hi,

Thank you for creating a PR however I would like to understand what the reasoning is behind this proposed change. 
Would you be able to provide an overview of the problem that you're trying to solve?

Thanks.

wa/instruments/energy_measurement.py
```diff
@@ -536,5 +538,9 @@ def extract_metrics(self, context):
                 context.add_metric(name, value, units)
 
     def teardown(self, context):
+        logger.debug('Job status {}'.format(context.current_job.status))
```
setrofim, line 541: This should not be logged by the instrument. This should be logged by the Runner, or possibly by the job itself, at the relevant point.

wa/instruments/energy_measurement.py
```diff
@@ -536,5 +538,9 @@ def extract_metrics(self, context):
...
+        if context.current_job.status == 'FAILED':
+            self.update_output(context)
```
setrofim, line 543: Output should not be updated during teardown stage. If this really needs to be done, then you can register a callback for JOB_OUTPUT_PROCESSED signal. However, as @marcbonnici  commented already -- why is this being done? Generally, if a job has failed than any data generated for it is suspect, and should not be used. If this is a hack to workaround a specific issue then it should not be upstreamed.

setrofim (changes requested):

wa/instruments/energy_measurement.py
deepaliyemul: The usecase is mainly for a methodology where the an iteration is run back to back for a period of time, without the setup and teardown. However, if a run fails due to any reason after a certain time, then the collected results go to __failed directory for the workload,. But for the energy instrument, the energr_measurement.csv file is reset and all the data till that point is lost. 
This change will put the data in the __failed directory for whatever has been collected. It will be on the user to device if it should be kept of discarded.

setrofim: My point remains-- if the job has failed, the data should be discarded. You're dealing with expected intermittent failures in the underlying application, then it should be handled at the workload level, so the job would not be marked as failed. If the failures are not "expected" and are a result of unstable platform, etc, then registering a handler for JOB_FAILED signal (contingent on user configuring a parameter, i.e. not as the default) would be a better solution (though still feels a bit hacky).

deepaliyemul: ok thank you.

marcbonnici:
Closing due to inactivity. Please feel to reopen if required.

PR #1175, opened 2021-07-13
workloads/pcmark: Add PCMark 3.0 support

mrkajetanp:
Add support for PCMark for Android version 3.
Use a 'version' Parameter to maintain support for v2.

### DIFF
--- a/wa/workloads/pcmark/__init__.py
+++ b/wa/workloads/pcmark/__init__.py
@@ -17,55 +17,82 @@ import re
 import sys
 import zipfile
 
-from wa import ApkUiautoWorkload
+from wa import ApkUiautoWorkload, Parameter
 from wa.framework.exception import WorkloadError
 
 
 class PcMark(ApkUiautoWorkload):
 
     name = 'pcmark'
+
+    supported_versions = ['3', '2']
     package_names = ['com.futuremark.pcmark.android.benchmark']
-    regex_matches = [re.compile(r'PcmaWebV2Score>([\d.]+)'),
-                     re.compile(r'PcmaVideoEditingScore>([\d.]+)'),
-                     re.compile(r'PcmaDataManipulationScore>([\d.]+)'),
-                     re.compile(r'PcmaPhotoEditingV2Score>([\d.]+)'),
-                     re.compile(r'PcmaWorkv2Score>([\d.]+)'),
-                     re.compile(r'PcmaWritingV2Score>([\d.]+)')]
+    regex_matches = {
+        '2': [re.compile(r'PcmaWebV2Score>([\d.]+)'),
+              re.compile(r'PcmaVideoEditingScore>([\d.]+)'),
+              re.compile(r'PcmaDataManipulationScore>([\d.]+)'),
+              re.compile(r'PcmaPhotoEditingV2Score>([\d.]+)'),
+              re.compile(r'PcmaWorkv2Score>([\d.]+)'),
+              re.compile(r'PcmaWritingV2Score>([\d.]+)')],
+
+        '3': [re.compile(r'PcmaWebV3Score>([\d.]+)'),
+              re.compile(r'PcmaVideoEditingV3Score>([\d.]+)'),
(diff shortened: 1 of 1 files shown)


wa/workloads/pcmark/__init__.py
```diff
@@ -17,55 +17,76 @@
...
-from wa import ApkUiautoWorkload
+from wa import ApkUiautoWorkload, Parameter
 from wa.framework.exception import WorkloadError
 
 
 class PcMark(ApkUiautoWorkload):
 
     name = 'pcmark'
+
+    supported_versions = ['3.0.4061', '2.0.3716']
```
marcbonnici, line 28: For the supported versions we probably don't need to be this specific as it looks like things don't tend to break in between versions much so we could get away with just `3` or `2`. The file resolver will still be able to choose more specific versions as appropriate.

wa/workloads/pcmark/__init__.py
```diff
@@ -17,55 +17,76 @@
...
+    A workload to execute the Work x.0 benchmarks within PCMark - https://www.futuremark.com/benchmarks/pcmark-android
 
     Test description:
     1. Open PCMark application
     2. Swipe right to the Benchmarks screen
-    3. Select the Work 2.0 benchmark
-    4. If needed, install the Work 2.0 benchmark (requires an internet connection)
-    5. Execute the Work 2.0 benchmark
-
-    Known working APK version: 2.0.3716
```
marcbonnici, line 44: Maybe add the specific versions you've tested back as "Known working" versions.

wa/workloads/pcmark/__init__.py
```diff
@@ -17,55 +17,76 @@
...
+            result_file = self.target.list_directory(results_path)[-1]
+
         self.result_file = result_file.rstrip()
         result = self.target.path.join(results_path, result_file)
         self.target.pull(result, context.output_directory)
         context.add_artifact('pcmark-result', self.result_file, kind='raw')
 
     def update_output(self, context):
-        expected_results = len(self.regex_matches)
+        expected_results = len(self.regex_matches[self.version.strip()[0]])
```
marcbonnici, line 84: If you're reusing `self.version.strip()[0]` multiple times, it might be cleaner to have a `major_version = self.version.strip()[0]` and reuse the variable instead.

marcbonnici (changes requested):

scojac01:
Out of interest, what devices has this been tested on? We ran in to issues with PCMark 2 on S21 devices as the UI layout wasn't in line with the buttons on screen.

marcbonnici (approved):

mrkajetanp:
> Out of interest, what devices has this been tested on? We ran in to issues with PCMark 2 on S21 devices as the UI layout wasn't in line with the buttons on screen.

That's interesting, I've been testing it on a Pixel 4 and both PCMark 2 & 3 seem to work fine there. Now I tried running it on a S9+ with Android 11, both of the versions used UI automation to install the Work benchmark properly but then neither actually ran it. They just kept sitting there on the main screen. Maybe there's some issue with the UI automation part in general?

Either way since it's the case on PCMark 2 as well it doesn't seem to be related to supporting v3 or this PR so maybe you should open a separate issue for that?

marcbonnici:
> Either way since it's the case on PCMark 2 as well it doesn't seem to be related to supporting v3 or this PR so maybe you should open a separate issue for that?
Testing this on my device both versions execute successfully so I'm happy to merge this for now. 

@scojac01 If you are still having some issues with this workload on another device, as suggested could you open another issue and attach UI dumps of where it is getting stuck?

scojac01:
> > Either way since it's the case on PCMark 2 as well it doesn't seem to be related to supporting v3 or this PR so maybe you should open a separate issue for that?
> 
> Testing this on my device both versions execute successfully so I'm happy to merge this for now.
> 
> @scojac01 If you are still having some issues with this workload on another device, as suggested could you open another issue and attach UI dumps of where it is getting stuck?

Hi, I'll open this up later yeah. We currently have a work around that requires having the full UL licence so it's not a blocker for us but it could become an issue for those without this.

PR #1180, opened 2021-09-28
Add support for Antutu 9.1.6.

pcc:
The code for reading the results is almost the same as for Antutu 8,
but it needs to be adjusted to account for a slightly different set
of benchmarks.

At least on my device, Antutu 9 takes over 10 minutes to run, so increase
the timeout to 20 minutes.


pcc:
This will require an update to the APK prebuilt. Do you want this as part of this PR or will you handle that yourselves?

setrofim:
Yes, please rebuild the UI automation APK and add it to the PR.

pcc:
Okay, this will require landing #1179 first as I have no convenient way of rebuilding the APKs otherwise. I'll rebuild the APKs on that PR and update this once it lands.

pcc:
> Okay, this will require landing #1179 first as I have no convenient way of rebuilding the APKs otherwise. I'll rebuild the APKs on that PR and update this once it lands.

Done.

marcbonnici (approved):

PR #1184, opened 2022-01-31
Add gfxbench

qais-yousef:
A fix for PCMark 'hanging' on running the test for a long number of minutes - on latest version at least. I am wary that this fix might introduce regressions on older versions.

And add support for non-corporate version of gfxbench. We just need to ignore Fixed Time Tests. Otherwise it seems they are the same (I don't have corporate version to try - hence the need to support the normal one :-))


qais-yousef:
It seems that the apk need to be signed, that's why I didn't push the regenerated ones. The ones I created required disabling install verification via USB. Let me know if I should push the new apks too.

wa/workloads/gfxbench/__init__.py
```diff
@@ -20,8 +20,9 @@
 
 class Gfxbench(ApkUiautoWorkload):
 
-    name = 'gfxbench-corporate'
-    package_names = ['net.kishonti.gfxbench.gl.v50000.corporate']
+    name = 'gfxbench'
+    package_names = ['com.glbenchmark.glbenchmark27']
```
marcbonnici, line 24: Sorry if this is an obvious question but is this the correct package name for the non-corporate gfxbenchmark application? If so how does it differ to our `glbenchmark` workload?

https://github.com/ARM-software/workload-automation/blob/72298ff9acb4eecca061261ee09567926f4df8db/wa/workloads/glbenchmark/__init__.py#L58

qais-yousef: Hmm, good question. That's odd. I tried this on Pixel 6 and Oneplus Nord and it worked fine. glbenchmark refused to start on Pixel 6 due to version mismatch, haven't tried on phone.

It seems gfxbenchmark supersedes what used to be known as glbenchmark

https://kishonti.net/graphics-benchmarking.jsp

we can put explicit version check to help. I am not sure what would be the best way to handle this tbh

qais-yousef: I must clarify. I didn't install glbench, just tried to run this workload with gfxbench installed to see if it'll start it instead. It didn't and got a complain of wrong version instead.

marcbonnici: Right I see, so you're using the non corporate version of gfxbench mark in your testing but that didn't work correctly with either the `gfxbench` or `glbench` workload? Do you have the error output generated / what version it is reporting? 
The two apps I've seen for these workloads have greatly differing UI's so I'm just trying to figure out which version you require.

qais-yousef: > Right I see, so you're using the non corporate version of gfxbench mark in your testing but that didn't work correctly with either the gfxbench or glbench workload?

marcbonnici: Right ok, this makes sense. Agreed, if we have a similar version check for 4.X/5.X as we do for the glbenchmark workload has for v2.7/2.5 this would prevent any overlap between the two workloads and provide the appropriate error message if the wrong application is available on the device.

qais-yousef: Okay. Looking at other examples, it seems adding `supported_versions = ['4', '5']` is enough. Tested that with only '4' in the supported version and I got a complaint about v 4 not preset. Once I added '5', it worked. I fixed the 2 required 2 lines complaint flagged by the automatic checks and pushed the updated branch.

marcbonnici (approved):
LGTM, can you rebuild and include the 2 uiauto apks inside this PR as well?

qais-yousef:
Sure. I updated the PR with the updated uiauto APKs.

By the way, I just realized that for gfxbench, the workload name `gfxbench` now refers to the free version. The corporate version is called `gfxbench-corporate` which matches what's done for geekbench. Existing agendas will be impacted by this change. To keep backward comparability we could use `gfxbench-free` and keep `gfxbench` reference to the corporate version. Opinions? I updated the commit message anyway to refer to this potential comparability problem.

scojac01:
Hi, just wanted to point out before this is approved that the non-corporate version of gfxbench will stop working after five consecutive runs on a device. This happens whether or not the apk data is cleared. I can't recall if uninstalling/reinstalling the apk resolves this but it should be tested and documented either way.

marcbonnici:
>Sure. I updated the PR with the updated uiauto APKs.

Thanks. 

>  To keep backward comparability we could use gfxbench-free and keep gfxbench reference to the corporate version.

Thanks for raising this as breaking backwards compatibility is something we always aim to avoid, however unless I'm mistaken the existing corporate workload was already called `gfxbench-corporate` so by introducing the `gfxbench` workload this should not impact existing users of this workload?

marcbonnici:
> the non-corporate version of gfxbench will stop working after five consecutive runs on a device.

Thanks @scojac01  for pointing this out, that would definitely be good to document any limitations. Do you know if that is 5 runs in a certain time frame or a per installation / device restriction? 

@qais-yousef have you seen this behavior on your side?

qais-yousef:
> Thanks for raising this as breaking backwards compatibility is something we always aim to avoid, however unless I'm mistaken the existing corporate workload was already called `gfxbench-corporate` so by introducing the `gfxbench` workload this should not impact existing users of this workload?

Oh, yes. Sorry last minute confusion from my side. Let me re-update the commit message and push again.

qais-yousef:
> > the non-corporate version of gfxbench will stop working after five consecutive runs on a device.
> 
> Thanks @scojac01 for pointing this out, that would definitely be good to document any limitations. Do you know if that is 5 runs in a certain time frame or a per installation / device restriction?
> 
> @qais-yousef have you seen this behavior on your side?

I just had a 6 iterations run finished without a problem. I triggered a single run afterwards it that worked.

I have created a free account in the past that I learned to just login when I install it. I logged out to see if this makes any difference. A 6 iteration run is on going now. Let's see if it'll finish too.

qais-yousef:
> > > the non-corporate version of gfxbench will stop working after five consecutive runs on a device.
> > 
> > 
> > Thanks @scojac01 for pointing this out, that would definitely be good to document any limitations. Do you know if that is 5 runs in a certain time frame or a per installation / device restriction?
> > @qais-yousef have you seen this behavior on your side?
> 
> I just had a 6 iterations run finished without a problem. I triggered a single run afterwards it that worked.
> 
> I have created a free account in the past that I learned to just login when I install it. I logged out to see if this makes any difference. A 6 iteration run is on going now. Let's see if it'll finish too.

I can't run anything if I am not signed in actually. I didn't hit any 5 iteration limitation myself. I tried running 6 iterations twice while signed in and they both ran fine.

marcbonnici:
> I can't run anything if I am not signed in actually. I didn't hit any 5 iteration limitation myself. I tried running 6 iterations twice while signed in and they both ran fine.

Interesting, I've been doing some runs on my side and didn't seem to need to login or hit a limit at 5 runs either. I did noticed a popup saying that they have noticed a lot of runs on this device but the workload was able to dismiss and continue running. 
The only issue I'm seeing lately are network errors when attempting to run select testcases but not sure if that unrelated?

qais-yousef:
> Interesting, I've been doing some runs on my side and didn't seem to need to login or hit a limit at 5 runs either. I did noticed a popup saying that they have noticed a lot of runs on this device but the workload was able to dismiss and continue running. The only issue I'm seeing lately are network errors when attempting to run select testcases but not sure if that unrelated?

I checked on another phone and it does allow me to run without logging in.. I couldn't find a note on what are the limitations on their website. I'm not sure if we can document much as things could change suddenly anyway. IMHO best leave it to the user to figure it out if they hit issues. There's no guarantee whatever advise we give will hold true across updates.

marcbonnici:
> I checked on another phone and it does allow me to run without logging in.. I couldn't find a note on what are the limitations on their website. I'm not sure if we can document much as things could change suddenly anyway. IMHO best leave it to the user to figure it out if they hit issues. There's no guarantee whatever advise we give will hold true across updates.

It would have been nice to give a heads up however I have still be unable to hit this restriction as well.

@scojac01 Do you have any further information about this scenario before we merge this?

qais-yousef:
@scojac01, any more info on the scenario you mentioned?

marcbonnici (approved):
I think we can merge this for now and can add a note if more information becomes available?

qais-yousef:
> I think we can merge this for now and can add a note if more information becomes available?

Sounds good to me

PR #1186, opened 2022-02-28
geekbench: Add/fix support for Geekbench5

qais-yousef:
The non corporate version of geekbench5 didn't work although the code
had everything needed, except for a number of tiny required tweaks:

1. Add '5' in the supported versions in __init__.py
2. Fix the name of the android package in__init__.py and
   UiAutomation.java
3. Improve handling of minorVersion to fix potential exception when we
   don't specify the minorVersion number in the yaml file. Launching
   geekbench5 works fine when it's the only one installed. But if you
   have multiple versions, then using the version string in the yaml
   agenda didn't like specifying '5' as the version and threw exception
   out of bound because we assume '5.X' as input. No reason I'm aware of
   to force support for a specific version of geekbench5. So keep it
   relaxed until we know for sure it breaks with a specific version.

Signed-off-by: Qais Yousef <qais.yousef@arm.com>


marcbonnici (approved):
Tested to work on my devices with a small fix in devlib. Raised a PR to fix the issue: https://github.com/ARM-software/devlib/pull/582

qais-yousef:
> Tested to work on my devices with a small fix in devlib. Raised a PR to fix the issue: https://github.com/ARM-software/devlib/pull/582

Thanks for the testing!

PR #1194, opened 2022-10-05
add support for loading runtime-config as a plugin

steve-ts:
This patch allows plugins to define their own runtime configurations. The primary motivation is to allow complex configurations that may be specific to a chip or device and have that configuration modifiable in sections to test sensitivity to those changes.

### DIFF
--- a/wa/framework/target/runtime_parameter_manager.py
+++ b/wa/framework/target/runtime_parameter_manager.py
@@ -22,6 +22,7 @@ from wa.framework.target.runtime_config import (SysfileValuesRuntimeConfig,
                                                 CpuidleRuntimeConfig,
                                                 AndroidRuntimeConfig)
 from wa.utils.types import obj_dict, caseless_string
+from wa.framework import pluginloader
 
 
 class RuntimeParameterManager(object):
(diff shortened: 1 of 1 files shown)


steve-ts:
The runtime parameters are currently defined by workload automation and there is no way to extend them AFAIK. I have other runtime parameters that I would like to create and use to configure devices and/or test runs. I can do this with instruments where I execute the configuration after a reboot, but I can't modify instrument parameters in sections. I am left to either re-implement sections within my instrument plugin or move from instrument to runtime configuration.

To give a few ideas of the use, there might be proprietary IP blocks in an SoC where we can vary voltage/frequency or test environment where equipment manufacturers don't allow exposing their protocols (e.g. setting a thermal chamber temperature over RS-232). With this patch, I would be able to create my own runtime config plugin and, using sections, run a workload at 10 different ambient temperatures or at 5 different frequencies.

steve-ts:
That is correct. I'll squash and update with a more informative message. Thanks.

marcbonnici (commented):
Great thanks, although would you be able to update the format of the commit header/style to match the rest of the repo?
We roughly try to follow the rules outlined at https://cbea.ms/git-commit/#seven-rules

setrofim:
Please could you document the new plugin type.

PR #1207, opened 2023-03-30
framework/target: Add persistent_data_path to Android target descriptor

mrkajetanp:
Include persistent_data_path in the Android target descriptor within framework/target/descriptor.py in line with adding persistent_data_path to AndroidTarget in devlib. This commit relies on the relevant change being merged into devlib first.

### DIFF
--- a/wa/framework/target/descriptor.py
+++ b/wa/framework/target/descriptor.py
@@ -541,6 +541,10 @@ TARGETS = {
                           description='''
                           Directory containing Android data
                           '''),
+                Parameter('persistent_data_path', kind=str, default='/storage/emulated/0/Android/data',
+                          description='''
+                          Directory containing persistent Android data
+                          '''),
                ], None),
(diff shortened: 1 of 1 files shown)


wa/framework/target/descriptor.py
```diff
@@ -541,6 +541,10 @@ def get_descriptions(self):  # pylint: disable=no-self-use
                           description='''
                           Directory containing Android data
                           '''),
+                Parameter('persistent_data_path', kind=str, default='/storage/emulated/0/Android/data',
+                          description='''
+                          Directory containing persistent Android data
+                          '''),
                ], None),
     'chromeos': ((ChromeOsTarget, 'ChromeOsConnection', []), COMMON_TARGET_PARAMS +
```
marcbonnici, line 549: I assume this should be the same for a ChomeOS target? In which case should we include a similar parameter for that target as well?

mrkajetanp:
Closing due to the PR no longer being needed due to changes here:
https://github.com/ARM-software/devlib/pull/622

PR #1208, opened 2023-03-31
drarm: Add DrArm workload

mrkajetanp:
Add a workload for the Dr Arm demo app. Includes functionality for automatically pulling the ADPF FPS report file from the target if one was generated by the app.

### DIFF
--- a/wa/workloads/drarm/__init__.py
+++ b/wa/workloads/drarm/__init__.py
@@ -1,0 +1,111 @@
+#    Copyright 2023 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+import os
+import pandas as pd
+from wa import ApkWorkload, Parameter
+from devlib.exception import TargetStableCalledProcessError
+
+
+class DrArm(ApkWorkload):
+
+    name = 'drarm'
+    package_names = ['com.Arm.DrArm']
+    activity = 'com.unity3d.player.UnityPlayerActivity'
+    install_timeout = 200
+    description = """
+    Dr. Arm’s Amazing Adventures is a “Souls-Like” Mobile Action Role Playing Game developed at Arm.
+    """
+
+    parameters = [
+        Parameter('timeout', kind=int, default=126,
+                  description='The amount of time the game should run for'),
+        Parameter('auto_demo', kind=bool, default=False,
+                  description='Start the demo automatically'),
+        Parameter('show_fps', kind=bool, default=False,
+                  description='Show the FPS count window in-game'),
+        Parameter('adpf', kind=bool, default=True,
+                  description='Enable ADPF'),
+        Parameter('adpf_auto', kind=bool, default=True,
+                  description='Enable automatic ADPF mode'),
+        Parameter('adpf_logging', kind=bool, default=False,
+                  description='Enable ADPF logging'),
+        Parameter('verbose_log', kind=bool, default=False,
+                  description='Emit reported stats as debug logs'),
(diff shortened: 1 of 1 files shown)


wa/workloads/drarm/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        if self.fixed_time_step is not None:
+            args['fixedTimeStep'] = self.fixed_time_step
+
+        if self.fps_report_file is not None:
+            args['fpsReportFileName'] = self.fps_report_file
+
+        return args
+
+    def run(self, context):
+        self.target.sleep(self.timeout)
```
marcbonnici, line 82: Does this workload not need to do anything in the run stage? Usually we want the workload to be kicked off here so the operation that is measured is consistent across runs.

wa/workloads/drarm/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def update_output(self, context):
+        super(DrArm, self).update_output(context)
+        outfile_glob = self.target.path.join(self.target.persistent_data_path, self.apk.package, 'files', '*.csv')
+
+        try:
+            ls_output = self.target.execute('ls {}'.format(outfile_glob), as_root=True)
+        except Exception:
```
marcbonnici, line 90: Can you be more specific in the exception here. Should this be a `TargetStableError` etc.?

wa/workloads/drarm/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        on_target_output_files = [f.strip() for f in ls_output.split('\n') if f]
+
+        self.logger.info('Extracting the ADPF FPS report from target...')
+        for file in on_target_output_files:
+            host_output_file = os.path.join(context.output_directory, os.path.basename(file))
+            self.target.pull(file, host_output_file)
+            context.add_artifact('adpf', host_output_file, kind='data',
+                                 description='ADPF report log in CSV format.')
+
+
```
marcbonnici, line 103: nit from linting: Extra new line.

wa/workloads/drarm/__init__.py
mrkajetanp: Good point, thanks, it should be `TargetStableCalledProcessError`

wa/workloads/drarm/__init__.py
mrkajetanp: The build we have has a command line option (`autoDemo`) that just starts the workload automatically a few seconds after the app is opened so we don't need to do anything in the run stage explicitly. It just always starts in the same way from the get-go.

mrkajetanp:
Updated in line with changes to https://github.com/ARM-software/devlib/pull/622

wa/workloads/drarm/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+class DrArm(ApkWorkload):
+
+    name = 'drarm'
+    package_names = ['com.Arm.DrArm']
+    activity = 'com.unity3d.player.UnityPlayerActivity'
+    install_timeout = 200
+    description = """
+    Dr. Arm’s Amazing Adventures is a “Souls-Like” Mobile Action Role Playing Game developed at Arm.
```
marcbonnici, line 30: Is this app publicly available? If so would it be possible to provide a link to where the application can be obtained from?

wa/workloads/drarm/__init__.py
marcbonnici: Is it worth executing that command to kick the workload off in the run stage itself? That way any results wouldn't be affected by the number/which instruments were also enabled on a given run?

wa/workloads/drarm/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+            args['fpsReportFileName'] = self.fps_report_file
+
+        return args
+
+    def run(self, context):
+        self.target.sleep(self.timeout)
+
+    def update_output(self, context):
+        super(DrArm, self).update_output(context)
+        outfile_glob = self.target.path.join(self.target.external_storage_app_dir, self.apk.package, 'files', '*.csv')
```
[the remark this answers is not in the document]
mrkajetanp: sure, done

wa/workloads/drarm/__init__.py
mrkajetanp: Hm I don't think so, it's an Arm-internal demo. It's been shown publicly like here:

https://community.arm.com/arm-community-blogs/b/graphics-gaming-and-vr-blog/posts/how-to-use-adaptive-performance

But we don't have the actual app published anywhere outside the internal systems as far as I know

PR #1226, opened 2023-08-14
instruments/trace_cmd: Allow setting top_buffer_size

mrkajetanp:
Allow optionally setting the top level ftrace buffer size separately from the devlib buffer size. The parameter will be passed to the devlib FtraceCollector and take effect there.

### DIFF
--- a/wa/instruments/trace_cmd.py
+++ b/wa/instruments/trace_cmd.py
@@ -125,6 +125,13 @@ class TraceCmdInstrument(Instrument):
                   value by going down from the specified size in
                   ``buffer_size_step`` intervals.
                   """),
+        Parameter('top_buffer_size', kind=int, default=None,
+                  global_alias='trace_top_buffer_size',
+                  description="""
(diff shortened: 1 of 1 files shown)


wa/instruments/trace_cmd.py
```diff
@@ -125,6 +125,12 @@ class TraceCmdInstrument(Instrument):
                   value by going down from the specified size in
                   ``buffer_size_step`` intervals.
                   """),
+        Parameter('top_buffer_size', kind=int, default=None,
+                  global_alias='trace_top_buffer_size',
+                  description="""
+                  The same as buffer_size except sets the top-level buffer
+                  size instead of the devlib buffer size.
```
marcbonnici, line 132: Would you be able to add an extra sentence as you mentioned in the devlib Pr to say that if not set, the top buffer size will default to the same as `buffer_size`?

mrkajetanp: No problem, done

marcbonnici (approved):

PR #1232, opened 2023-08-23
instruments/trace_cmd: Support setting trace-cmd mode

mrkajetanp:
The doc comment for the instrument mentions the option to select the trace_cmd_mode ('start' or 'record'). As far as I can tell this was never implemented. This commit adds the support for passing trace_cmd_mode to the FtraceCollector in devlib.

While at it, it also does some slight cleanup aronud the instrument, e.g. in the teardown() implementation which should be folded inside the collector instead of reimplemented here.

### DIFF
--- a/wa/instruments/trace_cmd.py
+++ b/wa/instruments/trace_cmd.py
@@ -158,6 +158,13 @@ class TraceCmdInstrument(Instrument):
                   description="""
                   The mode that trace-cmd will be started in.
                   For more details, consult the trace-cmd documentation.
+                  """),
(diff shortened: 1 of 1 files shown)


metin-arm:
@mrkajetanp , https://github.com/ARM-software/workload-automation/pull/1276 already added necessary support. Should we close this PR? Thanks.

mrkajetanp:
> already added necessary support. Should we close this PR? Thanks

If it's already been done then there's not much point keeping this around indeed. Well done! :)

PR #1246, opened 2024-01-05
fw/rt_config: Add unlock_screen config option in runtime_parameters

metin-arm:
Introduce 'unlock_screen' option in order to help in automating Android tests by unlocking device screen automatically. Surely this works only if no passcode is set.

User can tell how to unlock screen (e.g., diagonal, vertical, ...) via 'unlock_screen' option. If unlocking the screen is undesired, then 'unlock_screen' should be set to empty string (e.g., '').

'unlock_screen' option is effective if 'screen_on' option is set to true. Its default value is 'vertical'. Since 'screen_on' is also true by default, the screen will be unlocked via 'vertical' direction in default settings.

Note that this patch depends on
https://github.com/ARM-software/devlib/pull/659 in devlib repo.

### DIFF
--- a/doc/source/user_information/user_reference/runtime_parameters.rst
+++ b/doc/source/user_information/user_reference/runtime_parameters.rst
@@ -33,6 +33,7 @@ states.
                   iterations: 1
                   runtime_parameters:
                         screen_on: false
+                        unlock_screen: 'vertical'
                 - name: benchmarkpi
                   iterations: 1
         sections:
@@ -207,6 +208,13 @@ Android Specific Runtime Parameters
 
 :screen_on: A ``boolean`` to specify whether the devices screen should be
     turned on. Defaults to ``True``.
+
+:unlock_screen: A ``String`` to specify how the devices screen should be
+    unlocked. Unlocking screen is disabled by default. ``vertical``, ``diagonal``
+    and ``horizontal`` are the supported values (see :meth:`devlib.AndroidTarget.swipe_to_unlock`).
+    Note that unlocking succeeds when no passcode is set. Since unlocking screen
+    requires turning on the screen, this option overrides value of ``screen_on``
+    option.
 
 .. _setting-sysfiles:
 
--- a/wa/framework/target/runtime_config.py
+++ b/wa/framework/target/runtime_config.py
@@ -878,6 +878,11 @@ class AndroidRuntimeConfig(RuntimeConfig):
         if value is not None:
             obj.config['screen_on'] = value
 
+    @staticmethod
+    def set_unlock_screen(obj, value):
+        if value is not None:
+            obj.config['unlock_screen'] = value
+
     def __init__(self, target):
         self.config = defaultdict(dict)
         super(AndroidRuntimeConfig, self).__init__(target)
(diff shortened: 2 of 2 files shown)


wa/framework/target/runtime_config.py
```diff
@@ -930,6 +935,17 @@ def initialize(self):
                     Specify whether the device screen should be on
                     """)
 
+            param_name = 'unlock_screen'
+            self._runtime_params[param_name] = \
+                RuntimeParameter(
+                    param_name, kind=str,
+                    default='vertical',
```
metin-arm, line 942: I'm open to feedback about the default value/behavior.
Setting default value to empty string to disable unlocking screen feature also makes sense in order to keep current behavior of workload-automation.

wa/framework/target/runtime_config.py
```diff
@@ -939,6 +955,12 @@ def check_target(self):
     def validate_parameters(self):
         pass
 
+    def handle_screen_unlock(self):
+        if 'unlock_screen' in self.config and self.config['unlock_screen']:
+            screen_state = self.target.execute('dumpsys window')
+            if 'mDreamingLockscreen=true' in screen_state:
```
marcbonnici, line 961: Would it be worth adding a helper to devlib to try and detect whether the lockscreen is enabled or not rather than having this contained here?

wa/framework/target/runtime_config.py
```diff
@@ -939,6 +955,12 @@ def check_target(self):
     def validate_parameters(self):
         pass
 
+    def handle_screen_unlock(self):
+        if 'unlock_screen' in self.config and self.config['unlock_screen']:
```
douglas-raillard-arm, line 959: Can simplify that statement:
```
        if self.config.get('unlock_screen'):
```

wa/framework/target/runtime_config.py
douglas-raillard-arm: > Setting default value to empty string

Isn't there a way to have `None` by default (`null` in YAML) ? I'm not familiar with that parameter system.

metin-arm: `None` is kind of equal to empty string.
I wonder if we should disable unlocking screen feature to preserve current behavior of WA, or enable it by default.

wa/framework/target/runtime_config.py
metin-arm: Nope. It throws an exception if I don't first check if `unlock_screen` is in `self.config`.

wa/framework/target/runtime_config.py
marcbonnici: Agreed, to be consistent with the other parameters we should use `None` here rather than an empty string.
Also I don't think we should change the existing default behaviour, WA recommends that the screen lock is disabled by default so in an ideal case we should not have to try unlocking the device and therefore have less setup steps.

wa/framework/target/runtime_config.py
marcbonnici: I think this would be fixed by using `None` as the disabled case as this is what the setter method is checking for before populating `self.config`.

wa/framework/target/runtime_config.py
metin-arm: Makes sense. Will do that.

metin-arm:
@marcbonnici , @douglas-raillard-arm : can you please review again? Thanks.

doc/source/user_information/user_reference/runtime_parameters.rst
```diff
@@ -208,6 +209,13 @@ Android Specific Runtime Parameters
 :screen_on: A ``boolean`` to specify whether the devices screen should be
     turned on. Defaults to ``True``.
 
+:unlock_screen: A ``String`` to specify how the devices screen should be
+    unlocked. Unlocking screen is disabled by default. ``vertical``, ``diagonal``
+    and ``horizontal`` are the supported values (see ``swipe_to_unlock()``).
```
douglas-raillard-arm, line 214: ` ``swipe_to_unlock()`` ` should be replaced by the appropriate restructured text reference unless the rest of that doc uses that, e.g. :
```
:meth:`devlib.AndroidTarget.swipe_to_unlock`
```

metin-arm:
I'm unsure if the test suite check failures relate to my changes. I just updated a documentation in the last push. And the error logs look to be internal of the test suite itself.

marcbonnici (approved):

PR #1254, opened 2024-03-19
configuration: Allow including multiple files into one mapping.

ola1olssonarm:
The key for 'include#' can now be either a scalar or a list. A scalar triggers the same behaviour as before.
If the value is a list it must be a list of scalars (filepaths). The paths will be loaded and merged in order, and finally the resulting dict is included into the current scope.


marcbonnici (commented):
The change makes sense to me. Would would you be able to make the linter happy? [1]

[1] https://github.com/ARM-software/workload-automation/pull/1254/checks

ola1olssonarm:
Sorry Marc. I am new to github and I have no idea what I did in git. I hope I submitted a new PR for this.

PR #1258, opened 2024-04-21
workloads/speedometer: Add ApkUiautomation workload

eliftopuz:
When running speedometer workload on chrome package on Android 14, it gives a pop-up window. The pop-window is blocking the screen unless manually clicked. Uiautomation is implemented to dismiss the pop-up window. It's tested with Android 12 and Android 14 devices, but it might not be ready for merge.


wa/workloads/speedometer/__init__.py
```diff
@@ -120,6 +120,11 @@ class Speedometer(Workload):
                   The app package for the browser that will be launched.
                   """,
         ),
+        Parameter(
+            'version',
+            kind=str, default="123",
+            description='Specifies which version of the workload should be run.',
+            override=True),
```
marcbonnici, line 127: A `ApkUiautoWorkload` base class should already define a `version` Parameter, is there a specific reason this needs to be overridden?

wa/workloads/speedometer/__init__.py
```diff
@@ -24,15 +24,15 @@
...
+from wa import ApkUiautoWorkload, Parameter, Workload, File
 from wa.framework.exception import WorkloadError
 from wa.utils.exec_control import once
 from wa.utils.misc import safe_extract
 
 from devlib.utils.android import adb_command
 
 
-class Speedometer(Workload):
+class Speedometer(ApkUiautoWorkload):
```
marcbonnici, line 35: Just to add some background to this workload you may not be aware of.
 
A previous iteration of this workload used to be an `ApkUiautoWorkload` however after an investigation it seems that using UIAutomator introduced a performance overhead when running this workload as compared the current version [1].

It would be worth trying to confirm if this is still an issue.

[1] https://github.com/ARM-software/workload-automation/commit/7489b487e1fa808a793581617b0f8ed2b303f491#diff-83da7fd8a8e5e6b156a68f2d37658c2468594da1ceb7f48d0720a6f5324540ee

wa/workloads/speedometer/__init__.py
eliftopuz: Hi Marc,
you're right. I missed that. I deleted the `version` parameter. 
Instead I have added `prefer_host_package` and put default as `false` to avoid WA tries to install the version on the host.

wa/workloads/speedometer/__init__.py
eliftopuz: Hi,
The UiAutomator is only for chrome itself and Speedometer still runs offline. To dismiss the pop-ups coming from Chrome, I have added UiAutomator to `setup` and `extractResults` stages only. When the workload is in the `runWorkload` stage, there is no UiAutomator. That's why I believe I didn't see any drop on the score. 
Or I can try to find out another solution without UiAutomator.

eliftopuz:
Hi,
I am closing this PR as there is another possible solution without using UIAutomator. I will open a new PR with that one. 
Thanks

PR #1278, opened 2025-02-12
workloads/geekbench: Add support for Geekbench command-line build

eliftopuz:
Geekbench command-line build is added for Android. Geekbench apks allow to user to run the tests altogether. Using the command-line build, a single test or multiple tests can be specified.

### DIFF
--- a/wa/workloads/geekbench/__init__.py
+++ b/wa/workloads/geekbench/__init__.py
@@ -1,4 +1,4 @@
-#    Copyright 2013-2018 ARM Limited
+#    Copyright 2013-2025 ARM Limited
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -20,10 +20,11 @@ import tempfile
 import json
 from collections import defaultdict
 
-from wa import ApkUiautoWorkload, Parameter
+from wa import Workload, ApkUiautoWorkload, Parameter
 from wa.framework.exception import ConfigError, WorkloadError
 from wa.utils.misc import capitalize
-from wa.utils.types import version_tuple
+from wa.utils.types import version_tuple, list_or_integer
+from wa.utils.exec_control import once
 
 
 class Geekbench(ApkUiautoWorkload):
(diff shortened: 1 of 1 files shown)


wa/workloads/geekbench/__init__.py
```diff
@@ -20,10 +20,11 @@
 import json
```
metin-arm, line 20: Please follow Linux kernel coding [guideline](https://www.kernel.org/doc/html/latest/process/submitting-patches.html) regarding the commit message.

> Describe your changes in imperative mood, e.g. “make xyzzy do frotz” instead of “[This patch] makes xyzzy do frotz” or “[I] changed xyzzy to do frotz”, as if you are giving orders to the codebase to change its behaviour.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        self.workloads = self.gb6_workloads
+        self.params = ''
+        self.output = ''
+        self.tar_file = ''
+        self.file_exists = False
+
+    def init_resources(self, context):
+        # Necessary files to run the benchmark will be retrieved from WA as a tar file.
+        # WA will look for 'gb_cli_artifacts_<version>.tar' file to deploy them to the
+        # working directory. If there is no specified version, it will look for version 6.3.0 by default.
```
metin-arm, line 491: This line (and few more lines below) is too long (106 chars). Please ensure your changes are PEP-8 compliant. `autopep8` command might be helpful.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+    @once
+    def initialize(self, context):
+        if self.version[0] == '5':
+            self.workloads = self.gb5_workloads
+        # If the tar file does not exist in the target, deploy the assets
+        if not self.file_exists:
+            super(GeekbenchCmdline, self).initialize(context)
+            # Extract the tar file
+            self.target.execute("{} tar -xf {} -C {}".format(self.target.busybox,
+                                self.tar_file, self.target_working_directory))
```
metin-arm, line 516: This would be easier to maintain since we would not need to match spacing until `(`.
```Python
self.target.execute(
    "{} tar -xf {} -C {}".format(
        self.target.busybox, self.tar_file, self.target_working_directory))
```

wa/workloads/geekbench/__init__.py
```diff
@@ -20,10 +20,11 @@
 import json
 from collections import defaultdict
 
-from wa import ApkUiautoWorkload, Parameter
+from wa import Workload, ApkUiautoWorkload, Parameter
```
metin-arm, line 23: `Workload` is imported, but I'm unsure if it's used.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        321: 'SFFT',
+        # Single-Core, Multi-Core, OpenCL and Vulkan
+        303: 'Gaussian Blur',
+        305: 'Face Detection',
+        306: 'Horizon Detection',
+    }
+
+    parameters = [
+        Parameter('cpumask', kind=str,
+                  default=None, override=False, mandatory=False,
```
metin-arm, line 446: Shouldn't default value be `''` since the type is `str`?

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+                  description='CPU mask used by taskset'),
+        Parameter('section', kind=int, default=1, allowed_values=[1, 4],
+                  override=False, mandatory=False, description="""
+                  Run the specified sections. It should be 1 for cpu benchmarks,
```
metin-arm, line 450: `s/cpu/CPU/`

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        Parameter('workload', kind=list_or_integer, override=False, mandatory=False,
+                  default=301, description='Specify workload to run'),
+        Parameter('iterations', kind=int, default=5,
+                  description='Number of iterations'),
+        Parameter('workload_gap', kind=int, default=2000, mandatory=False,
+                  description='N milliseconds gap between workloads'),
+        Parameter('output_file', kind=str, default='',
+                  override=False, mandatory=False, description="""
+                  Specify the name of the output results file. It can be 'json', 'csv',
+                  'xml', 'html', 'text' files. If it is not specified, the output file
```
metin-arm, line 465: `'json', 'csv',  'xml', 'html', 'text'` part should be replaced by something like `', '.join(allowed_extensions)` to have single version of data.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+                  will be generated as a json file."""),
```
metin-arm, line 466: `s/json/JSON/`

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+                  mandatory=False, description='Upload results to Geekbench Browser'),
+        Parameter('is_single_core', kind=bool, default=True, override=False, mandatory=False,
+                  description='Run workload in single-core or multi-core mode'),
+        Parameter('workload', kind=list_or_integer, override=False, mandatory=False,
+                  default=301, description='Specify workload to run'),
+        Parameter('iterations', kind=int, default=5,
+                  description='Number of iterations'),
+        Parameter('workload_gap', kind=int, default=2000, mandatory=False,
+                  description='N milliseconds gap between workloads'),
+        Parameter('output_file', kind=str, default='',
```
metin-arm, line 462: Shouldn't default value be `json` per the parameter description?

wa/workloads/geekbench/__init__.py
metin-arm, line 491: You could format it like this and have function doc automatically:
```Python
    """
    Retrieves necessary files to run the benchmark in TAR format from workload-automation-assets (?).
    WA will look for `gb_cli_artifacts_<version>.tar` file to deploy them to the
    working directory. If there is no specified version, it will look for version
    6.3.0 by default.
    """
```

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        self.deployable_assets = [''.join(['gb_cli_artifacts', '_', self.version, '.tar'])]
+
+        # Create a working directory
+        self.target.working_directory = self.target.path.join(self.target.working_directory, f'gb_cli-{self.version}')
+        self.target.execute("mkdir -p {}".format(self.target.working_directory))
+        GeekbenchCmdline.target_working_directory = self.target.working_directory
+
+        # Path for the artifacts tar file
+        self.tar_file = self.target.path.join(self.target_working_directory, self.deployable_assets[0])
+        # Check the tar file if it exists
```
metin-arm, line 501: Please also tell what happens if it does not.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+
+    def setup(self, context):
+        super(GeekbenchCmdline, self).setup(context)
+
+        self.params = ''
+        if self.section == 4:
+            # Run OpenCL, Vulkan benchmark
+            self.params += '--section {} '.format(self.section)
+        else:
+            # Run cpu benchmark
```
metin-arm, line 529: `s/cpu/CPU/`

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+            self.params += '--section {} '.format(self.section)
+            if self.is_single_core:
+                self.params += '--single-core '
+            else:
+                self.params += '--multi-core '
```
metin-arm, line 534: `self.params += '--single-core ' if self.is_single_core else '--multi-core '` would be slightly more Pythonic.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+
+        if self.upload:
+            self.params += '--upload '
+        else:
+            self.params += '--no-upload '
```
metin-arm, line 539: `self.params += '--upload ' if self.upload else '--no-upload '` would be slightly more Pythonic.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        if self.iterations:
+            self.params += '--iterations {} '.format(self.iterations)
+
+        if self.workload_gap:
+            self.params += '--workload-gap {} '.format(self.workload_gap)
+
+        extension = os.path.splitext(self.output_file)[1][1:]
+        if self.output_file and extension not in self.allowed_extensions:
+            msg = 'No allowed extension specified. Allowed extensions: {}'
+            raise ValueError(msg.format(', '.join(self.allowed_extensions)))
```
metin-arm, line 557: Would this be more appropriate to build error message in a single place?
```Python
msg = f"No allowed extension specified. Allowed extensions: {', '.join(self.allowed_extensions)}"
raise ValueError(msg)
```

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        super(GeekbenchCmdline, self).update_output(context)
+        if not self.output:
+            return
+        for workload in self.workload:
+            scores = []
+            matches = re.findall(self.workloads[workload] + '(.+\d)', self.output)
+            for match in matches:
+                scores.append(int(re.search(r'\d+', match).group(0)))
+            if self.section == 4:
+                context.add_metric("OpenCL Score " + self.workloads[workload], scores[0])
```
metin-arm, line 591: It can also be `Vulkan` in addition to `OpenCL`, right?

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+            else:
+                context.add_metric("Single-Core Score " + self.workloads[workload], scores[0])
+                if not self.is_single_core:
+                    context.add_metric("Multi-Core Score " + self.workloads[workload], scores[1])
+
+    def extract_results(self, context):
+        super(GeekbenchCmdline, self).extract_results(context)
+        # Extract results on the target
```
metin-arm, line 599: The comment should be swapped with the previous line?

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,237 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        self.target.pull(self.target_result_json, self.host_result_json)
+        context.add_artifact('GeekbenchCmdline_results', self.host_result_json, kind='raw')
+
+    @once
+    def finalize(self, context):
+        # Clean up the working directory
```
metin-arm, line 605: You can drop this comment. It does not add much value.

wa/workloads/geekbench/__init__.py
metin-arm: I think we need to update copyright year (`#    Copyright 2013-2018 ARM Limited` at the beginning of file) as well.

wa/workloads/geekbench/__init__.py
eliftopuz: `Workload` is used from `GeekbenchCmdline` class

wa/workloads/geekbench/__init__.py
eliftopuz: `OpenCL` and `Vulkan` has different sections. I changed this part a bit. You can check the updated version.

wa/workloads/geekbench/__init__.py
eliftopuz: I was using `maxLineLength=120`, I didn't see any rule for max length in the WA documentation. Also in the same file I can see that some lines > 100.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,239 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        306: 'Horizon Detection',
+    }
+
+    binary_name = 'geekbench_aarch64'
+
+    allowed_extensions = ['json', 'csv', 'xml', 'html', 'text']
+
+    parameters = [
+        Parameter('cpumask', kind=str,
+                  default='', override=False, mandatory=False,
```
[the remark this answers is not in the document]
eliftopuz: I will remove those options.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,239 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        # If the tar file does not exist in the target, deploy the assets
+        if not self.file_exists:
+            super(GeekbenchCmdline, self).initialize(context)
+            # Extract the tar file
+            self.target.execute(
+                "{} tar -xf {} -C {}".format(
+                    self.target.busybox, self.tar_file, self.target_working_directory))
+
+    def validate(self):
+        super(GeekbenchCmdline, self).validate()
```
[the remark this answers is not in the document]
eliftopuz: I will remove that one.

wa/workloads/geekbench/__init__.py
```diff
@@ -370,3 +371,239 @@ class GeekbenchCorproate(Geekbench):  # pylint: disable=too-many-ancestors
...
+        """
+        Retrieves necessary files to run the benchmark in TAR format.
+        WA will look for `gb_cli_artifacts_<version>.tar` file to deploy them to the
+        working directory. If there is no specified version, it will look for version
+        6.3.0 by default.
+        """
+        self.deployable_assets = [''.join(['gb_cli_artifacts', '_', self.version, '.tar'])]
+
+        # Create a working directory
+        self.target.working_directory = self.target.path.join(self.target.working_directory, f'gb_cli-{self.version}')
```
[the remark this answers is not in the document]
eliftopuz: I will change this part.
But I am still changing the `working_directory` in the agenda file when I am running the workload. Otherwise it gives me a permission error and cannot execute the binary.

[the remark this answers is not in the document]
marcbonnici: Changing the `working_directory` from the agenda is fine as this would be a conscious decision by the user and would impact all workloads equally. 

> Otherwise it gives me a permission error and cannot execute the binary.

If these are binaries that require execution permission, is there a reason they need to be in the working directory? Or could we use `target.install` for this instead?

wa/workloads/geekbench/__init__.py
metin-arm: https://peps.python.org/pep-0008/#maximum-line-length says 79 & 72, but I think 100 is OK in general.

metin-arm (approved):

scojac01:
Does Geekbench command line not require a license to run? Should we be upstreaming workloads that have such a dependency?

wa/workloads/geekbench/__init__.py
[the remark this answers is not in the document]
eliftopuz: Hi Marc,
I did change this part and now I am extracting the assets in the `executables_directory`. Also there is no need to change `working_directory` in the agenda file.

lukaszluba-arm:
Hi Scott,
1. Yes, it needs a license.
2. This dependency is easy to solve and we can ask our partners to get their own license (internally in different teams - we don't have that issue).

On the other hand... Not having the GB6 cmd line version prevents us from being efficient in our experiments (and aligned across different teams that everyone can reproduce exactly the same scenario).

The APK from public store doesn't allow you to run single sub-test and be more precised how it should be run.
This pull request will be beneficial for a lot of use cases that we (and our partners) investigate.

wa/workloads/geekbench/__init__.py
marcbonnici: Is this comment still pending?

marcbonnici:
> If we agree that the dependency is for users of WA to sort themselves then that's fine. We have a geekbench cli workload that we have used for years and have not upstreamed because we were under the impression that such a thing was frowned upon.

Hi @scojac01,

That's interesting, I had thought we had set precedence for this back with the geekbench-corporate workload, in that if we can assume that it is reasonable that a non-trivial amount of other users would be able to source the appropriate workloads, then it could still be beneficial to upstream the workload. 

Are you aware of of substantial differences between the availability of the corporate vs the cli flavours of the workloads?

wa/workloads/geekbench/__init__.py
eliftopuz: No, I did change the comment with the previous line.

scojac01:
> > If we agree that the dependency is for users of WA to sort themselves then that's fine. We have a geekbench cli workload that we have used for years and have not upstreamed because we were under the impression that such a thing was frowned upon.
> 
> Hi @scojac01,
> 
> That's interesting, I had thought we had set precedence for this back with the geekbench-corporate workload, in that if we can assume that it is reasonable that a non-trivial amount of other users would be able to source the appropriate workloads, then it could still be beneficial to upstream the workload.
> 
> Are you aware of of substantial differences between the availability of the corporate vs the cli flavours of the workloads?

Hi @marcbonnici That's fair enough. We must have missed that but happy with that precedent. The cli workload is the corporate one just triggered in a different method with a few extra params. The license should be the same so no impact there.

marcbonnici (approved):

PR #1280, opened 2025-02-26
speedometer: Add version 3.0 support

metin-arm:
- [speedometer: Introduce trivial cleanups](https://github.com/ARM-software/workload-automation/commit/fa8a93262da612e8025f746c0da233783f3f70bd)
  - Remove unused imports
  - Handle the case that `@candidate_files` may be undefined
  - Customize the log message regarding Speedometer timeout
- [speedometer: Add version 3.0 support](https://github.com/ARM-software/workload-automation/commit/6cdd4c92c2cb08a9e6468d5b631b2aafd48326bc)
  Port v3.0 of Speedometer from [Webkit](https://github.com/WebKit/WebKit/tree/main/Websites/browserbench.org/Speedometer3.0) repo and update tarballs.

  `version` field can be added to the workload agenda file to specify
  which version of Speedomer should be used.


marcbonnici (commented):
The changes look good to me however I'm a little worried about the size of the archives if we include multiple versions.
The original 2.0 archive was ~3MB, the new 3.0 archive is almost 12MB. If we expect to keep adding new versions of speedometer should we look at adding the archive into the asset's repo instead of the main WA repo? 

What do you think?

edit: Also if you have a chance could you rebase the PR on the latest master so the CI has another change to run?

metin-arm:
> The changes look good to me however I'm a little worried about the size of the archives if we include multiple versions. The original 2.0 archive was ~3MB, the new 3.0 archive is almost 12MB. If we expect to keep adding new versions of speedometer should we look at adding the archive into the asset's repo instead of the main WA repo?
> 
> What do you think?
> 
> edit: Also if you have a chance could you rebase the PR on the latest master so the CI has another change to run?

Hi Marc,

I hear your concern, but the workload is not maintained that frequently. I think v2.0 has added ~7 years ago?

Rebased my `speedometer` branch.

marcbonnici (approved):

PR #1285, opened 2025-04-01
Bumping Antutu versions

scojac01:
Upstreaming our working Antutu workload that covers a number of V10 versions.


scojac01:
Superseded by the addition of the Antutu BDP class.

PR #1286, opened 2025-04-08
Adding Antutu BDP class for cli automation

scojac01:
Workload for executing the BDP versions of the Antutu APK.

This will only work with specific APKS provided by Antutu but does
unlock command line automation and the capturing of a result file
as opposed to using UiAuto and Regex.

Known working version: 10.4.3-domesticAndroidFullBdp


wa/workloads/antutu/__init__.py
```diff
@@ -90,12 +114,24 @@ def __init__(self, device, **kwargs):
         super(Antutu, self).__init__(device, **kwargs)
         self.gui.timeout = 1200
 
+    def initialize(self, context):
+        super(Antutu, self).initialize(context)
+        #Install the supporting benchmark
+        supporting_apk = context.get_resource(ApkFile(self, package='com.antutu.benchmark.full'))
+        self.target.install(supporting_apk)
```
marcbonnici, line 121: Do we have `supporting_apks` for previous versions or should this only be if the version is == 10.X?

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +145,98 @@ def extract_scores(self, context, regex_version):
...
+
+    This will only work with specific APKS provided by Antutu but does
+    unlock command line automation and the capturing of a result file
+    as opposed to using UiAuto and Regex.
+
+    Known working version: 10.4.3-domesticAndroidFullBdp
+    '''
+    package_name = 'com.antutu.ABenchMark'
+    test_dir = '/sdcard/Android/data/com.antutu/ABenchMark/files/.antutu/'
+    result_file = '/sdcard/Documents/antutu/last_result.json'
```
marcbonnici, line 201: Do we strictly need these hardcoded paths and will they be the same on all targets?

Could we make use of the `external_storage` and `external_storage_app_dir` target properties or something similar to generate these paths?

wa/workloads/antutu/__init__.py
```diff
@@ -13,8 +13,9 @@
 # limitations under the License.
 #
 import re
-
-from wa import ApkUiautoWorkload, WorkloadError, Parameter
+import os
+import time
+from wa import ApkUiautoWorkload, Workload, WorkloadError, Parameter, ApkFile, File
```
marcbonnici, line 18: nit: Can we keep a single empty line between system and wa imports?

wa/workloads/antutu/__init__.py
```diff
@@ -90,12 +114,24 @@ def __init__(self, device, **kwargs):
         super(Antutu, self).__init__(device, **kwargs)
         self.gui.timeout = 1200
 
+    def initialize(self, context):
+        super(Antutu, self).initialize(context)
+        #Install the supporting benchmark
+        supporting_apk = context.get_resource(ApkFile(self, package='com.antutu.benchmark.full'))
+        self.target.install(supporting_apk)
+        #Ensure the orientation is set to portrait
+        self.target.set_rotation(0)
```
marcbonnici, line 123: I'm assuming it should not, but would setting this change the behaviour for any previous versions of the workload?

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +145,98 @@ def extract_scores(self, context, regex_version):
                             result = float('NaN')
                         entry = regex.pattern.rsplit(None, 1)[0]
                         context.add_metric(entry, result, lower_is_better=False)
+        #Calculate group scores if 'CPU' in entry:
```
marcbonnici, line 148: nit: Fix indentation

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +145,98 @@ def extract_scores(self, context, regex_version):
...
             raise WorkloadError(msg.format(len(regex_version), expected_results))
 
+        context.add_metric('CPU Total Score', cpu_result, lower_is_better=False)
+        context.add_metric('GPU Total Score', gpu_result, lower_is_better=False)
+        context.add_metric('UX Total Score', ux_result, lower_is_better=False)
+        context.add_metric('MEM Total Score', mem_result, lower_is_better=False)
+
+        #Calculate overall scores
+        overall_result = float(cpu_result + gpu_result + ux_result + mem_result)
+        context.add_metric('Overall Score', overall_result, lower_is_better=False)
```
marcbonnici, line 173: Do these metrics still have meaning for previous versions of the workload? 
Should we only be adding them if that are found during result parsing?

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +145,98 @@ def extract_scores(self, context, regex_version):
...
+        self.target.push(settings_xml, self.test_dir)
+        #Ensure the orientation is set to portrait
+        self.target.set_rotation(0)
+        #Remove any pre-existing test results
+        self.target.execute('rm {}'.format(self.result_file))
+
+    def run(self, context):
+        super(AntutuBDP, self).run(context)
+        #Launch the tests
+        activity = 'com.android.module.app.ui.start.ABenchMarkStart --ez isExternal true --es whereTo "test"'
```
marcbonnici, line 222: nit: This could be defined at the class level.

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +145,98 @@ def extract_scores(self, context, regex_version):
...
+        self.target.execute('am start -n {}/{}'.format(self.package_name, activity))
+        #Wait 10 minutres, then begin polling every 30s for the test result to appear
```
marcbonnici, line 224: Question: Is it possible for these tests to run quicker than this or is it expected / required that these tests take at least this length of time to run?

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +145,98 @@ def extract_scores(self, context, regex_version):
...
+        #Launch the apk to initialize the test dir, then kill it
+        self.target.execute('am start {}/com.android.module.app.ui.test.activity.ActivityScoreBench'.format(self.package_name))
+        self.target.execute('am force-stop {}'.format(self.package_name))
+        #Copy the settings.xml to the test dir
+        settings_xml = context.get_resource(File(self, 'settings.xml'))
+        self.target.push(settings_xml, self.test_dir)
+        #Ensure the orientation is set to portrait
+        self.target.set_rotation(0)
+        #Remove any pre-existing test results
+        self.target.execute('rm {}'.format(self.result_file))
```
marcbonnici, line 217: Should this cleanup be performed in the teardown stage?

wa/workloads/antutu/__init__.py
scojac01: The tests are currently running for approximately 18 minutes on the latest flagships. I figured 10 minutes gives enough room for improvement for the foreseeable but I could parameterise this if preferred? (I will fix that typo though in the comment)

wa/workloads/antutu/__init__.py
scojac01: `supporting_apks` have been required since at least version 8 which was released in January 2020.

wa/workloads/antutu/__init__.py
scojac01: No, this is just to ensure the results page is in the correct orientation for the grabbing of results.

wa/workloads/antutu/__init__.py
scojac01: Yeah, back to v8 these will always be present. The subtests that make up the groups will be slightly different but all are captured by the logic just above.

wa/workloads/antutu/__init__.py
scojac01: I've done this but not sure if it's the correct way to do it. To use these properties I can't do it at the class level so I am reusing the same variables in each function that requires it.

wa/workloads/antutu/__init__.py
marcbonnici: Ok, stupid question but in that case how have we been handling the `supporting_apk` before this change? 

Also for backwards compatibility with v7 should we add a >=8 check?

wa/workloads/antutu/__init__.py
marcbonnici: Ok cool, in which case same as above do we need to add a >= 8?

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +146,107 @@ def extract_scores(self, context, regex_version):
...
+    This will only work with specific APKS provided by Antutu but does
+    unlock command line automation and the capturing of a result file
+    as opposed to using UiAuto and Regex.
+
+    Known working version: 10.4.3-domesticAndroidFullBdp
+    '''
+    package_name = 'com.antutu.ABenchMark'
+    activity = 'com.android.module.app.ui.start.ABenchMarkStart --ez isExternal true --es whereTo "test"'
+
+    def initialize(self, context):
```
marcbonnici, line 203: The `initalize` method should only contain actions that are performed once for all instances of a workload.
Things that relate to a specific run should be located in the `setup` stage such as setting the orientation and dealing with existing results otherwise the device may not be in the state we expect.

The intention for each stage can be found: https://workload-automation.readthedocs.io/en/latest/developer_information.html#execution-model

wa/workloads/antutu/__init__.py
marcbonnici: Thanks, I'd probably suggest setting this once either  in `__init__` or early in `initalize` and reusing the attribute in the subsequent locations e.g. similar to what we have in lmbench.

[1] https://github.com/ARM-software/workload-automation/blob/2d14c82f92f3b71089550671773c5718b7e4514a/wa/workloads/lmbench/__init__.py#L100C14-L100C24

wa/workloads/antutu/__init__.py
marcbonnici: Fair enough, that sounds reasonable for now.

wa/workloads/antutu/__init__.py
scojac01: I don't believe the public workload has. I have been running it in a forked repo since v8. 

We could add a check but V7 is so old and almost impossible to get hold of these days. Happy to add if you want though.

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +146,109 @@ def extract_scores(self, context, regex_version):
...
+        super(AntutuBDP, self).initialize(context)
+        #Set the files and directories we need
+        self.test_dir = os.path.join(self.target.external_storage_app_dir, 'com.antutu.ABenchMark/files/.antutu/')
```
marcbonnici, line 206: Could this be updated to use the path join functionality from the target for the full path? 
i.e. `self.taraget.path.join(self.target.external_storage_app_dir, 'com.antutu.ABenchMark', 'files', '.antutu')`

And the same for the below paths.

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +146,109 @@ def extract_scores(self, context, regex_version):
...
+        self.settings_xml = context.get_resource(File(self, 'settings.xml'))
+        self.result_file = os.path.join(self.target.external_storage, 'Documents/antutu/last_result.json')
+        self.output_file = os.path.join(context.output_directory, 'antutu_results.json')
+
+    def setup(self, context):
+        super(AntutuBDP, self).setup(context)
+        #Install the supporting benchmark
+        supporting_apk = context.get_resource(ApkFile(self, package='com.antutu.benchmark.full'))
+        self.logger.info("Installing the supporting APK")
```
marcbonnici, line 215: nit: This could probably be at the debug level.

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +146,109 @@ def extract_scores(self, context, regex_version):
...
+            self.extract_scores(context, self.regex_matches_v10)
         if self.version.startswith('9'):
             self.extract_scores(context, self.regex_matches_v9)
         if self.version.startswith('8'):
             self.extract_scores(context, self.regex_matches_v8)
         if self.version.startswith('7'):
             self.extract_scores(context, self.regex_matches_v7)
+
+
+class AntutuBDP(Workload):
```
marcbonnici, line 188: High level question, what was the reason that we're using the generic `Workload` as opposed to the `ApkWorkload` class which could have handled some of the generic APK management automatically?

wa/workloads/antutu/__init__.py
```diff
@@ -109,16 +146,109 @@ def extract_scores(self, context, regex_version):
...
+        #Set the files and directories we need
+        self.test_dir = os.path.join(self.target.external_storage_app_dir, 'com.antutu.ABenchMark/files/.antutu/')
+        self.settings_xml = context.get_resource(File(self, 'settings.xml'))
+        self.result_file = os.path.join(self.target.external_storage, 'Documents/antutu/last_result.json')
+        self.output_file = os.path.join(context.output_directory, 'antutu_results.json')
+
+    def setup(self, context):
+        super(AntutuBDP, self).setup(context)
+        #Install the supporting benchmark
+        supporting_apk = context.get_resource(ApkFile(self, package='com.antutu.benchmark.full'))
```
marcbonnici, line 214: Could we move the discovery of this APK to `__init__` as well? 
That would allow the workload to fail earlier if the required file can't be found.

wa/workloads/antutu/__init__.py
scojac01: I went with the Workload class as although it's installing an APK it is executing a command line argument. Is that an incorrect approach? Would you prefer this to be an ApkWorkload?

PR #1288, opened 2025-05-07
Add support for Honor of Kings game

metin-arm:
- **framework: utils: Improve logging in revent replay related methods:**

  Also fix typos around the modified lines.

- **workloads: Add support for Honor of Kings game:**

    Introduce a new workload for automating match replays in Honor of Kings.
    This workload leverages ApkReventWorkload to launch the game and play
    back a specified replay file.

    Allow customization through parameters:
    - `activity`: Specifies the activity string of the game.
    - `replay_file`: Designates the replay file to be uploaded.

    The game can be replayed via `wa replay` command as well, but this
    plugin makes data collection (`fps`, `trace-cmd` like augmentations)
    a lot easier.

### DIFF
--- a/wa/workloads/honorofkings/__init__.py
+++ b/wa/workloads/honorofkings/__init__.py
@@ -1,0 +1,61 @@
+#    Copyright 2025 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+
+from wa import ApkReventWorkload, Parameter
+
+
+class HoK(ApkReventWorkload):
+    name = 'honorofkings'
+    uninstall = False
+    clear_data_on_reset = False  # Don't clear assets on exit
+    requires_network = True  # The game requires network connection
+    description = (
+        'Launch a match replay in Honor of Kings.\n\n'
+        'The game must already have a user logged in and the plugins downloaded.'
+    )
+    package_names = [
+        'com.levelinfinite.sgameGlobal',
+        'com.tencent.tmgp.sgame',
+    ]
+
+    parameters = [
+        Parameter(
+            'activity',
+            kind=str,
+            default='.SGameGlobalActivity',
+            description='Activity name of Honor of Kings game.',
+        ),
+        Parameter(
+            'replay_file',
+            kind=str,
+            default='replay.abc',
+            description='Honor of Kings Replay file name.',
+        ),
+    ]
+
+    def setup(self, context):
+        upload_dir = self.target.path.join(
+            self.target.external_storage_app_dir,
+            self.apk.apk_info.package,
+            'files',
+            'Replay'
+        )
+        replay_file = os.path.join(self.dependencies_directory, self.replay_file)
+        self.logger.debug('Uploading "%s" to "%s"...', replay_file, upload_dir)
+        self.target.push(replay_file, upload_dir)
+
+        super().setup(context)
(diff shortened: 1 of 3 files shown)


wa/workloads/honorofkings/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from wa import ApkReventWorkload, Parameter
+
+
+class HoK(ApkReventWorkload):
+    name = 'honorofkings'
+    package_names = ['com.levelinfinite.sgameGlobal']
```
scojac01, line 21: I know there are different packages of this game. I think there is a tencent version. We should add that too. 

Same applies for the activity.

metin-arm: True, but I did not have chance to test other package names.
Maybe we should introduce parameters to customize activity and package name?

scojac01: That would be my approach.

metin-arm: Done. Thanks for the review!

wa/workloads/honorofkings/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+        self.logger.debug(
+            'Uploading "%s/%s" to "%s/Android/data/%s/files/Replay/"...',
+            self.dependencies_directory,
+            self.replay_file,
+            self.target.external_storage,
+            self.package_names[0]
+        )
+        self.target.push(
+            f'{self.dependencies_directory}/{self.replay_file}',
+            f'{self.target.external_storage}/Android/data/{self.package_names[0]}/files/Replay/'
```
marcbonnici, line 80: In the case we provide multiple package names this could end up pushing to the wrong location. Can we use the package name from the resolved APK instead of the package name list?

wa/workloads/honorofkings/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+        Parameter(
+            'activity',
+            kind=str,
+            default='.SGameGlobalActivity',
+            description='Activity name of Honor of Kings game.'
+        ),
+        Parameter(
+            'revent_run_file',
+            kind=str,
+            default='komodo.run.revent',
```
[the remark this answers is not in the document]
metin-arm: No specific reason. I just was not aware of that :)

wa/workloads/honorofkings/__init__.py
metin-arm: My bad. It's actually a single package name. Will fix it.

marcbonnici: Apologies I should elaborate.

The usual way the apk resolution works is the workload provides an array of default package names that are supported for the workload. In this case we could have a single package name, as you had, but IIUC from Scott's comment there could be other package names that we may want to add in the future.

By default a `ApkWorkload` also inherits the `package_name` Parameter [1] so that when WA does the package resolution it looks for either the default package names specified from the workload or as specified by the user [2]. So we shouldn't need to manually add another Parameter here.

Instead I was proposing that we pull the package name package from the resolved apk so that we always ensure that we have the correct name i.e. we use something like `self.apk.apk_info.package` to populate this string. That way we can also avoid having to specifically manipulating the `package_names` parameter in `__init__` as well similar to what we do in the vellamo workload [3].

Does that seem reasonable?

[1] https://github.com/ARM-software/workload-automation/blob/master/wa/framework/workload.py#L190
[2] https://github.com/ARM-software/workload-automation/blob/master/wa/framework/workload.py#L759
[3] https://github.com/ARM-software/workload-automation/blob/master/wa/workloads/vellamo/__init__.py#L103

metin-arm: Hope it's addressed now. Thanks a lot for the explanation.

wa/workloads/honorofkings/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        Parameter(
+            'replay_file',
+            kind=str,
+            default='replay.abc',
+            description='Honor of Kings Replay file name.',
+        ),
+    ]
+
+    def setup(self, context):
+        upload_dir = (
```
marcbonnici, line 49: Could we use `self.target.path.join(...)` to generate the file paths rather than using strings with hard coded `/`'s as a separator here and below?

e.g. https://github.com/ARM-software/workload-automation/blob/2d14c82f92f3b71089550671773c5718b7e4514a/wa/workloads/chrome/__init__.py#L86

metin-arm: Done. Thank you!

marcbonnici (approved):

PR #1289, opened 2025-05-21
Fix: Use `gman` on MacOS for `wa show`

ep1cman:
The version of `man` that ships with MacOS does not support the `-l` argument that `wa show` uses. You can get the GNU man binary via brew that does support this, but they prefix commands with `g` if they overlap with system ones.

This patch checks if the system is MacOS and uses `gman` instead


marcbonnici (commented):
Thanks for fixing this, one small suggestion.

marcbonnici (approved):

PR #1290, opened 2025-05-29
Add cpu_load_poller instrument

ep1cman:
This PR adds a new instrument called `cpu_load_poller`. Based on the existing file poller instrument, it polls `/proc/stats` and creates a CSV of CPU load per core. 

While this might not be the most detailed view of CPU loads compared to other tools like `perf`, its is a simple, lightweight way to get a general idea of what is occurring with CPU cores. Because it is the same format as the file poller it allows easy comparison to other counters accessible via sysfs/debugfs (e.g Frequencies, GPU load, power, memory utilization etc.)


ep1cman:
Ok, I did all that without realizing `proc_stat` existed 😫. These have somewhat similar functionality but this PR has a few advantages:
 - Per core utilization
 - Ability to poll faster (proc_stat, doesn't allow floats for the period, and the timestamps have a resolution of seconds)
 
As well as some disadvantages:
 - The raw data isn't kept. Along with it the break down of the various columns (timestamp,user,nice,system,idle,iowait,irq,softirq,steal,guest,guest_nice)
 - Doesn't include overall system load (but it could if I didn't filter that line out)

ep1cman:
Ok I have have reworked this into a replacement for perf_stat, but in the process lost the ability to align its timestamp format with the `poller` instrument. I will look into adding a flag to pick between them.

The only "breaking" change between this and the old instrument is that the timestamp now includes fractions of seconds. I don't know how big of a deal that is.

